PR #5, opened 2013-06-06
Allow rendering w/ custom template

samatjain:
New argument to create_map(), template, lets you pass a string to use as template for output.

What's the best way to document this?


ocefpaf:
Hi @samatjain,

I was thinking about mobile friendly maps this week.  However, I have very little experience with mobile application.  Did you advance this idea into a template?  Could you share it here?

PR #65, opened 2014-12-29
Fix #64: Remove margins from leaflet-tiles

### LINKED ISSUE #64: Tile alignment in ipython notebook
On my notebooks the first tile is always offset.
It seems there is a top margin for all but the first tile
because of the following css selector in style.css:

```css
.rendered_html * + img
```

maybe adding some css to override margins is an good idea

lennart0901:
This fixes #64

### DIFF
--- a/folium/templates/ipynb_init_css.html
+++ b/folium/templates/ipynb_init_css.html
@@ -4,6 +4,10 @@
         color: black !important;
     }
 
+    .leaflet-container .leaflet-tile {
+       margin: 0;
+    }
+
     .leaflet-control-zoom-in {
         text-decoration: none !important;
     }


lennart0901:
Before patch
before


After patch:
after

ocefpaf:
@lennart0901 Thanks!  Just one last request.  Can you add that notebook to a gist (or something similar) so we can get a nbviewer URL.  That way we can refer to it in the Changelog file.

@birdage BTW we should start a Changelog file!

lennart0901:
https://gist.github.com/lennart0901/b158758d262c58b54c1d

ocefpaf:
@lennart0901 thanks again!

PR #53, opened 2014-10-15
Boundary for zoom and latitude/longitude {Issue #29}

Paradoxeuh:

Issue #29, opened 2014-05-12
Zoom and fit map to bounding box?


birdage:
is there a way at present to zoom and fit the map to a bounding box?

ocefpaf:
This can be achieved with  #67 and #119.

PR #67, opened 2015-01-05
Added max and min zoom keywords.

ocefpaf:
This PR supersedes #53.  It implements adds a keywords for `min_zoom ` to limit the zoom level.

I am not a leaflet expert but I think that to properly solve issue #29 we have to use `map.fitBounds(bounds);`, but it is unclear to me how to proper implement this now.   Maybe another keyqord `bounds=True`? @birdage what do you think?

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -56,8 +56,9 @@ class Map(object):
     '''Create a Map with Folium'''
 
     def __init__(self, location=None, width=960, height=500,
-                 tiles='OpenStreetMap', API_key=None, max_zoom=18,
-                 zoom_start=10, attr=None):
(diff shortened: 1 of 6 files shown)


birdage:
@ocefpaf i think the min and max zoom level was to stop the map zooming all the way out, fit bounds does satisfy this in some respects, but min/max would bound the uses zoom levels 

http://leafletjs.com/reference.html
```
inZoom	Number	null	Minimum zoom level of the map. Overrides any minZoom set on map layers.
maxZoom	Number	null	Maximum zoom level of the map. This overrides any maxZoom set on map layers.
```

ocefpaf:
Thanks @birdage.

birdage:
no worries i wrote a comment then reviewed the code and it looks good!

Issue #49, opened 2014-09-28
Add an option to adjust text popup width and height


ocefpaf:
I was wondering, even though I like the JSON approach (and GeoJSONs are a far more elegant solution than what I am about to propose!), why not add a functionality to modify the width and height of the text popup and let the user add HTML graphics in folium? Both bokeh and mpl3d can save the plots as HTML.

I already toyed with the idea here:
http://nbviewer.ipython.org/github/ocefpaf/secoora/blob/mpld3/notebooks/sandbox/test_mpld3.ipynb

But my knowledge of leaflet is very limited to add the option in a PR for folium. (I guess that it is a css thing, right?)

PR #68, opened 2015-01-09
Sets width to auto.

ocefpaf:
**This PR is not ready to merge!!!**

I just want to start a discussion on how to solve #49.

(Take a look at PR #69 after reading this one.)

In this PR I override the min/max width with the css:
```css
    .leaflet-popup-content {
        width:auto !important;
        }
```

That allows the user to use HTML popup with custom width and height without implementing a "new HTML popup category."  However, that modifications seems to set the simple text width to the [default](http://leafletjs.com/reference.html#popup) `minWidth: 50px`.

The simple texts look awful that way.  User would need to always use HTML and set the width manually.  I am not happy with this solution, but it was the only one I got after reading leaflet docs.

### DIFF
--- a/folium/templates/fol_template.html
+++ b/folium/templates/fol_template.html
@@ -44,7 +44,9 @@
         right:0;
         left:0;
       }
-
+    .leaflet-popup-content {
+        width:auto !important;
+        }
    </style>
 </head>
 


birdage:
@ocefpaf let me know whne its ready to go

ocefpaf:
Closing in favor of #69.

PR #69, opened 2015-01-09
Set width

ocefpaf:
**This PR is not ready to merge!!!**

I just want to start a discussion on how to solve #49.

(Before reading this go to PR #68 to see an alternative implementation.)

In this I added an extra keyword `width` that allows the user to change leaflet's `maxWidth`.  If `width` is omitted it defaults to 300px ([Leaflet default](http://leafletjs.com/reference.html#popup)).

With this modification I do not break simple text behavior like in #68, but I force the user to jump an extra hoop by defining the width twice.  One in the HTML popup text and another in the keyword.

Maybe there is no escape... we have to implement a HTML popup category.  Something like `HTML=True` and activate a `HTML_popup.js` with the width and height options.  But, for now at least, I prefer this solution because it does not break folium's API and adds quick workaround for #49.

@birdage and @wrobstory Can I pick your brains on this issue?  Any ideas on how to do this in a better way?

[notebook example](https://gist.github.com/ocefpaf/ad8c4e60c65eb0d9b835)

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -269,7 +269,7 @@ class Map(object):
     @iter_obj('simple')
     def simple_marker(self, location=None, popup='Pop Text', popup_on=True,
                       marker_color='blue', marker_icon='info-sign',
-                      clustered_marker=False, icon_angle=0):
+                      clustered_marker=False, icon_angle=0, width=300):
         '''Create a simple stock Leaflet marker on the map, with optional
         popup text or Vincent visualization.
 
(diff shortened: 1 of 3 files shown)


ocefpaf:
~~It just came to my that we could use maxWidth=map_width like in the vega template.  Simplifying this a lot...~~

Nope, that would extend all texts to the map width.

birdage:
@ocefpaf let me know whne its ready to go

ocefpaf:
@birdage I am not happy with #68.  I am +0.5 with #69 (because it will not change folium's current behavior.)

What do you think?  Merge #69 and close #68?

birdage:
@ocefpaf yeah that sounds good, close #68 for now

ocefpaf:
@birdage I have been using this PR for a while now.  I am convinced that this method is the easiest way to get custom HTML popups.

Unless you have something under the sleeve, lets get this PR merged!  (And prepare a new release.)

birdage:
@ocefpaf ive got nothing at the mo, so lets merge for now and see if we can come up with anything down the road.

ocefpaf:
Agreed.

PR #73, opened 2015-01-13
Support multiPolyLine

scari:


ocefpaf:
Thanks @scari!  Looks like a good addition.

Could you add a test in `tests/folium_tests.py` to the new template `multiPolyline.js`?  So we do not break it with future modifications.

scari:
Okay. I'll take a look! Thanks!

scari:
I just added test for multiPolyLine. Please take a look. Thanks!

birdage:
@scari Thanks!

ocefpaf:
@scari Awesome!  One last thing, if you allow me to be picky, can you write a docstring for def `multiline()`?

scari:
@ocefpaf No problem at all! I just added it. Sorry that I didn't make my PR neat ;)

ocefpaf:
Thanks @scari  and don't worry about pretty.  That is why we have code reviwes :wink: 

The PR is starting to take shape, but I have two more requests

- Can you write something about how `multiline` differs from `line` in the docstring?
- If I am not mistaken you need to add multilines to `fol_template.html` otherwise the new `var multiline` won't get written.

scari:
@ocefpaf Almost?

folium/folium.py
```diff
@@ -382,6 +382,58 @@ def line(self, locations,
...
+                'opacity': line_opacity}
+
+        varname = 'multiline_{}'.format(count)
+        multiline_rendered = multiline_temp.render({'multiline': varname,
+                                          'locations': locations,
+                                          'options': multiline_opts})
+
+        add_multiline = 'map.addLayer({});'.format(varname)
+
+        self.template_vars.setdefault('multiline', []).append((multiline_rendered,
```
ocefpaf: Looking below I think that this should be `self.template_vars.setdefault('multilines', [])` and not `'multiline'`, right?

folium/templates/fol_template.html
```diff
@@ -139,6 +139,11 @@
       {{ add_line }}
       {% endfor %}
 
+      {% for multiline, add_multiline in multilines %}
```
ocefpaf: @scari you are looping over `multilines` and not `multiline` {% for multiline, add_multiline in multilines %}
If you change it above you'll get it working properly.

ocefpaf:
Almost...  Just check my comments above and, if possible, add a use case example (nbviewer link preferable, but do not worry too much about the example.).

scari:
Thanks for reviewing. I just updated my PR. Here is an use case example. 
http://nbviewer.ipython.org/gist/scari/b030f428bd6d278aec11

ocefpaf:
Great @scari!  Merging...

PR #95, opened 2015-03-13
popups allow unicode

apatil:
Hi @ocefpaf, another small one. This allows unicode titles on popups.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -22,6 +22,8 @@ from pkg_resources import resource_string, resource_filename
 from folium import utilities
 from folium.six import text_type, binary_type, iteritems
 
+import sys
+
 
 ENV = Environment(loader=PackageLoader('folium', 'templates'))
 
@@ -616,10 +618,21 @@ class Map(object):
         if not popup_on:
             return 'var no_pop = null;'
         else:
-            if isinstance(popup, str):
+            if sys.version_info >= (3,0):
+                utype, stype = str, bytes
+            else:
+                utype, stype = unicode, str
+
+            if isinstance(popup, (utype, stype)):
                 popup_temp = self.env.get_template('simple_popup.js')
+                if isinstance(popup, utype):
+                    popup_txt = popup.encode('ascii', 'xmlcharrefreplace')
+                else:
+                    popup_txt = popup
+                if sys.version_info >= (3,0):
+                    popup_txt = popup_txt.decode()
                 return popup_temp.render({'pop_name': mk_name + str(count),
-                                          'pop_txt': json.dumps(popup),
+                                          'pop_txt': json.dumps(str(popup_txt)),
                                           'width': width})
             elif isinstance(popup, tuple):
                 #Update template with JS libs
@@ -653,6 +666,8 @@ class Map(object):
                                           'max_width': max_width,
                                           'json_out': json_out,
                                           'vega_id': vega_id})
+            else:
+                raise TypeError("Unrecognized popup type: {!r}".format(popup))
 
     @iter_obj('geojson')
     def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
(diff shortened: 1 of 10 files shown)


ocefpaf:
Hi @apatil we can do,
```python
isinstance(popup, (str, unicode))
```
to be shorter.  We also need to be smarter about this to ensure py3k compatibility.

Last but not least, we will probably need something like:
```python
'fancy string/unicode'.encode('ascii', 'xmlcharrefreplace')
```

To get a proper HTML.

Are you up to implement this?  If not, can you open an issue and reference this PR?

folium/folium.py
```diff
@@ -616,7 +616,7 @@ def _popup_render(self, popup=None, mk_name=None, count=None,
         if not popup_on:
             return 'var no_pop = null;'
         else:
-            if isinstance(popup, str):
+            if isinstance(popup, str) or isinstance(popup, unicode):
                 popup_temp = self.env.get_template('simple_popup.js')
```
ocefpaf: Use `isinstance(popup, (str, unicode))` instead.

apatil:
Hi @ocefpaf, that's a good test suite. :) I'll give it a shot later today.

apatil:
How does that look, @ocefpaf?

ocefpaf:
Looking great @apatil!  As a folium maintainer and a non-English speaker Thanks for doing this!!

I will review/test it and merge ASAP.

folium/folium.py
```diff
@@ -616,10 +618,21 @@ def _popup_render(self, popup=None, mk_name=None, count=None,
...
+                if isinstance(popup, utype):
+                    popup_txt = popup.encode('ascii', 'xmlcharrefreplace')
+                else:
+                    popup_txt = popup
+                if sys.version_info >= (3,0):
+                    popup_txt = popup_txt.decode()
                 return popup_temp.render({'pop_name': mk_name + str(count),
-                                          'pop_txt': json.dumps(popup),
+                                          'pop_txt': json.dumps(str(popup_txt)),
```
ocefpaf: @apatil I am on the fence here.  I am not sure if the call to `str` is necessary, but it does not hurt...  What is your argument?

folium/folium.py
```diff
@@ -653,6 +666,8 @@ def _popup_render(self, popup=None, mk_name=None, count=None,
                                           'max_width': max_width,
                                           'json_out': json_out,
                                           'vega_id': vega_id})
+            else:
+                raise TypeError("Unrecognized popup type")
```
ocefpaf: @apatil Can you change this to:
```python
raise TypeError("Unrecognized popup type: {!r}".format(popup))
```

folium/folium.py
```diff
@@ -616,10 +618,21 @@ def _popup_render(self, popup=None, mk_name=None, count=None,
...
-            if isinstance(popup, str):
+            if sys.version_info >= (3,0):
+                utype, stype = str, bytes
+            else:
+                utype, stype = unicode, str
+
+            if isinstance(popup, (utype, stype)):
                 popup_temp = self.env.get_template('simple_popup.js')
+                if isinstance(popup, utype):
+                    popup_txt = popup.encode('ascii', 'xmlcharrefreplace')
```
ocefpaf: Loved it!

ocefpaf:
@apatil I just finished review and it is looking really nice.  Can you address the my `raise` request and squash the commits?

Also, can you add an entry in the [Changelog](https://github.com/python-visualization/folium/blob/master/CHANGES.txt).

(If you do not have time for that let me know and I will merge and do it myself :wink: )

folium/folium.py
apatil: I was getting type errors in python3 without it. The encode('ascii', ...) produces a bytes object, which json.dumps doesn't like.

ocefpaf: `str` it is then.

apatil:
@ocefpaf sorry for the delay and the extra commit, but it's mostly squashed. :)

ocefpaf:
Nice!  Thanks for sending another **very** useful PR!

apatil:
My pleasure. :)

PR #106, opened 2015-04-21
Draw GeoJSON first.

### LINKED ISSUE #30: Impossible to click on circle markers when geo_json is called
When both markers and geo_json methods are called, we cannot click in the markers in order to see their popup (at least on firefox).

It seems that it is because the html code to manage the geojson is called after the javascript code which manages the markers.

The issue appears even if the geo_json python method is called in first.

Manually modifing the html code in order to call  L.geoJson before L.circle solves the issue

ocefpaf:
Fixes #30

### DIFF
--- a/folium/templates/geojson_template.html
+++ b/folium/templates/geojson_template.html
@@ -78,6 +78,18 @@
               attribution: '{{ attr }}'
           }).addTo(map);
 
+          {% for style in geo_styles %}
+          {{ style }}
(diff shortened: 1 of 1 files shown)


ocefpaf:
@birdage Travis seems to be following the Brazilian calendar and is taking a holiday today!  If it finishes can you review this one.

birdage:
@ocefpaf haha of course!

PR #109, opened 2015-04-21
Forcing HTTPS when available

### LINKED ISSUE #98: Support for HTTPS/SSL
I spent a bit of time trying to figure out why my map wasn't rendering inline in ipython notebook following these examples: http://nbviewer.ipython.org/gist/bburky/7763555/folium-ipython.ipynb. The problem is that I was running ipython notebook over https, and ssl isn't available for OpenStreetMap's public tiles.

Stamen makes SSL tiles freely available (http://maps.stamen.com/, ctrl-f ssl):

 > If you'd like to display these map tiles on a website that requires HTTPS, use our tile SSL endpoint by replacing http://tile.stamen.com with https://stamen-tiles.a.ssl.fastly.net. Multiple subdomains can be also be used: https://stamen-tiles-{S}.a.ssl.fastly.net
 > 
 > JavaScript can be loaded from https://stamen-maps.a.ssl.fastly.net/js/tile.stamen.js.

Perhaps using a protocol relative URL for stamen tiles, and making them default for SSL (or all the time-- they're gorgeous!), would make folium more easy to pick up, since some beginner ipython noteobok tutorials set up SSL.

A warning message in the documentation about the availability of certain tiles through SSL could be helpful as well.

ocefpaf:
Using `https://` instead of the the scheme agnostic `//` is the recommended pattern now.
(See [the-protocol-relative-url](http://www.paulirish.com/2010/the-protocol-relative-url/).)

With this change users can serve or open the HTML  directly using `file://`.

This PR closes #98.  And hopefully it will stop the "open in browser" question :stuck_out_tongue_winking_eye: 

PS: The following tiles are not available (or I could not test) in HTTPS:

- MapQuest Open
- MapQuest Open Aerial
- Cloudmade (cannot test don't have an AIP_key, but probably works)
- Mapbox (cannot test don't have an AIP_key, but probably works)


ocefpaf:
@birdage This ones deserves some care and thought before merging.  Also, if you decide to merge it we have to update the docs to mention what tiles are not available in HTTPS.

PR #116, opened 2015-04-23
Added non-pixel width and height.

### LINKED ISSUE #22: Non pixel height and width not availible
Non integer values doesn't accessed with `Map` constructor. For example `100%`.

ocefpaf:
@birdage I think I got the `if`/`else` clause, but can you take a look and check for any problems?

[Here](http://nbviewer.ipython.org/urls/gist.githubusercontent.com/ocefpaf/9d7a2877113153195f06/raw/8e68a463054fc25d95bba87015ee10726c47bb65/test_width_height.ipynb) is a notebook with the current implementation.  The browser window or the iframe container can be re-sized and the map will occupy all the available space.  (Closes #22.)

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -84,9 +84,9 @@ class Map(object):
         ----------
         location: tuple or list, default None
             Latitude and Longitude of Map (Northing, Easting).
-        width: int, default 960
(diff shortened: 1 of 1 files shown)


birdage:
looks good!

Issue #123, opened 2015-05-12
Tile Layer Missing options


ozak:
Hi,

I tried using the ``add_tile_layer`` command, but noticed there are many options that one would like to pass which are not available. See [here](http://leafletjs.com/reference.html#tilelayer) for possible options.

birdage:
hi @ozak, this was a basic implementation of a WMS tile layer, feel free to add to it and give back to the community if it does not meet your needs

ozak:
@birdage That was precisely my idea!

birdage:
@ozak love it! :beers:

ozak:
@birdage I just added the missing options.

On another front, is there a way to choose which projection to set the map if one wants to use e.g. OSM? From what I see, it seems to start using spherical mercator, but I'd like to be able to change that. Any pointers?

birdage:
@ozak there is a plugin here https://github.com/kartena/Proj4Leaflet to do just that but i dont know if it does it out the box

ocefpaf:
This is being addressed in #124.  I am closing this to tidy-up the issues list.

PR #124, opened 2015-05-12
added default options to tile_layer

ozak:
This addresses #123


birdage:
@ozak can you share an example image or notebook showing the fix pls. :+1:

ozak:
Will do once I solve come issues with the projection of the data I have and upload the web site I am working on.

birdage:
@ozak great thanks!

ozak:
One question...I have been trying to have the layer be active to start with, which is what the ``active=True`` option should accomplish. Looking at some examples it seemed to be just a mater of having a ``.addTo(map)`` in the ``L.tileLayer()`` command. But that is not the case from some test I just ran. Any ideas?

birdage:
@ozak that should do it in my experience, adding `add to map` should activate the layer auto-magically, can u see the layer in the list?

ozak:
without the change in the ``active`` option that I included, the layer is on the list and can be added by clicking. But if I activate it by adding  the ``.addTo(map)`` like below does not work.

    var tile = L.tileLayer('./data/{z}/{x}/{y}.png',{
                      'minZoom': 0,
                      'maxZoom': 18,
                      'tms': true,
                      'continuousWorld': false,
                      'noWrap': false,
                      'zoomOffset': 0,
                      'zoomReverse': false,
                      'opacity': 1 
                      }).addTo(map);

which is what would be generated with the change I made. Or do you mean another ``folium`` command?

My test basically executed the following code in an IPython notebook, 

    map_osm = folium.Map(location=[0,0], tiles='OpenStreetMap', zoom_start=2)
    map_osm.add_tile_layer(tile_name='Name',tile_url='./data/{z}/{x}/{y}.png', active=True, tms=True)
    map_osm.add_layers_to_map()
    map_osm.create_map(path='./webpage.html')
    map_osm.render_iframe = True
    map_osm

using ``master`` the layer is selectable but not active, and using this pull request it doesn't even work, unless ``active=False``.

birdage:
@ozak interesting, that would work, maybe its the variable declaration? i.e `var tile =` try removing that....not hopeful that will work tho

ozak:
Are you able to have some (non-base) layer active to start with?

birdage:
@ozak not at the min i dont think, the only things active from init are the base layers

ozak:
It looks like one needs to add it to the map's layers:

    var map = L.map('folium_d128601ca2af4b14b77d02ab215b86b9', {
                                 center:[0, 0],
                                 zoom: 2,
                                 maxBounds: bounds,
                                 layers: [base_tile, tilelayer]
                               });

birdage:
@ozak you shouldn't really need to do that, doing `addtomap` should add and activate the layer, also if we go the above way it should be part of a layer control group

ozak:
I completely agree with you. It should add it without having to do much else. Looking at the html output and the code, it seems ``add_layers_to_map()`` does add the control, but it is not clear if it also adds the layer. from [here](http://leafletjs.com/reference.html#map-stuff-methods) I thought it would need another ``addLayer`` command. Am I misunderstanding Leaflet?

ozak:
Looks like one would need to generate some loop to write the elements of ``layer_list`` into the following

    map.addLayer(layer_layer_element);

which would have to come after the ``var map=`` statement.

ocefpaf:
@ozak Can you add a test to your new functionality? Just a simple template test like we have in https://github.com/python-visualization/folium/blob/master/tests/folium_tests.py would do.

ozak:
Ok...I added the test and it worked on my computer. I see it is failing on 3.2, but it is not related to what I have done. It fails due to 

    assert self.map.template_vars['fit_bounds'] == fit_bounds_rendered

ocefpaf:
Python 3.x is crashing randomly and I am unsure why.  Anyway, I re-started the test and Travis-CI is happy now.

Back to you @birdage!  If you think this is ready merge away!

andrewgiessel:
My guess is that the crashes are related to an issue I unearthed in #152:  dictionaries are unordered and thus if you don't use the `sort_keys=True` flag, then the string you get from `json.dumps()` has random order and the string comparison fails.

ocefpaf:
@andrewgiessel Thanks for fixing the tests.

@ozak Sorry for the delay!  Can you squash your commits?

@birdage Are you done with the reviews?

ozak:
@ocefpaf Done!

folium/folium.py
```diff
@@ -221,7 +221,9 @@ def __init__(self, location=None, width='100%', height='100%',
         self.template_vars.setdefault('tile_layers', [])
 
     @iter_obj('simple')
-    def add_tile_layer(self, tile_name=None, tile_url=None, active=False):
+    def add_tile_layer(self, tile_name=None, tile_url=None, active=False, minZoom=0,
+                        maxZoom=18, tms=False, continuousWorld=False, noWrap=False, zoomOffset=0,
+                        zoomReverse=False, opacity=1, attribution='Leaflet'):
```
ocefpaf: `attribution` should be a string about the source of the layer, correct?  If leave the default equal to `None`.

ocefpaf: @ozak can you write some docstring regarding these options?

folium/templates/leaflet-dvf.markers.min.js
```diff
@@ -0,0 +1,6 @@
...
...artMarker.prototype.initialize.call(this,t,e)},options:{weight:1,opacity:1,color:"#000",fill:!0,radius:10,rotation:0,numberOfSides:50,mouseOverExaggeration:1.2,maxDegrees:360,iconSize:new L.Point(50,40)},_highlight:function(t){var e=t.radiusX,i=t.radiusY,n=t.barThickness;return t.oldBarThickness=n,t.oldRadiusX=e,t.oldRadiusY=i,t.radiusX*=t.mouseOverExaggeration,t.radiusY*=t.mouseOverExaggeration,t.barThickness=t.radiusX-e+n,t},_unhighlight:function(t){return t.radiusX=t.oldRadiusX,t.radiusY=t.oldRadiusY,t.barThickness=t.oldBarThickness,t},_loadComponents:function(){var t,e,i,n,a=0,o=0,r=0,s=this.options.maxDegrees||360,f=this.options.rotation,d=this.options,c=this.options.data,l=this.options.chartOptions,u=function(t,e){var i=0;return t[e]&&(i=parseFloat(t[e])),i};for(n in c)t=u(c,n),a+=t;if(a>0)for(n in c)t=parseFloat(c[n]),i=l[n],r=t/a,o=r*s,d.startAngle=f,d.endAngle=f+o,d.fillColor=i.fillColor,d.color=i.color||"#000",d.radiusX=this.options.radiusX||this.options.radius,d.radiusY=this.options.radiusY||this.options.radius,d.rotation=0,d.key=n,d.value=t,d.displayName=i.displayName,d.displayText=i.displayText,e=new L.RadialBarMarker(this._latlng,d),this._bindMouseEvents(e),f=d.endAng
```
ocefpaf: We have to sort this file out.  A recent PR from @BibMartin re-introduced it here in `plugins/leaflet-dvf.markers.min.js` while the template in `.templates/dvf_js_ref.txt` loads is from a [CDN](https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.2/leaflet-dvf.markers.min.js).

@ozak Can you remove it from this PR until we sort this out?

tests/folium_tests.py
```diff
@@ -139,6 +139,38 @@ def test_wms_layer(self):
                                'wms_transparent': 'true'})
         assert map.template_vars['wms_layers'][0] == wms
 
+    def test_add_tile_layer(self):
+        '''Test add_layer URLs'''
+
+        map = folium.Map(location=[44, -73], zoom_start=3)
+        tile_name = "Temperature"
+        tile_url = 'http://gis.srh.noaa.gov/arcgis/services/NDFDTemps/'
+        tile_url += 'MapServer/WMSServer'
```
ocefpaf: Are any of these options specific to a WMS version?  If so we have to document it!

ocefpaf:
Thanks @ozak! I made a few comments in your PR.  Feel free to ping me if you need any clarification.

ozak:
@ocefpaf I'll try to answer to your comments, but won't be able to do so for a couple of days, since I am trying to meet a couple of deadlines. Sorry for the delay!

ocefpaf:
No problem.   The time we demand from contributors is inverse to pay.  That means you get zero pay but infinite time :stuck_out_tongue_winking_eye: 

PS: If you address the comments I might get this merged as is and we can work on it later.

ozak:
The problem is that I have to reacquaint myself with the code and leaflet in general... :smile:

ocefpaf:
> The problem is that I have to reacquaint myself with the code and leaflet in general...

I hear you!  The whole purpose of folium is to avoid that, but as developer we cannot afford to do it :unamused:

ocefpaf:
I am thinking about merging this into the `v0.1.6` instead of master. Do you think you can re-submit your PR there?

ocefpaf:
@ozak with the recent changes on folium I am certain that this belongs in the `v0.1.6` branch.  Let me know if you can point this PR there or if I should try some git-fu to cherry pick this.

ozak:
Sure...I will take a look into it. But as I said it might take a while (at least a month) before I can go back to this.

ocefpaf:
Not a problem.  I am behind the schedule with `v0.1.6`  (and we can always do a `v0.1.7`) :wink:

ocefpaf:
Closing this in favor of #236.

Thanks @ozak!  (I kept your commits in #236)

PR #128, opened 2015-05-13
Fixed test name

ocefpaf:
~~Not ready to merge.  I will add more stuff tomorrow morning.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,20 +8,15 @@ env:
     - CONDA="python=3.4"
 
 before_install:
-    - URL=http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh
-    - wget $URL -O miniconda.sh
+    - wget http://bit.ly/miniconda -O miniconda.sh
     - bash miniconda.sh -b -p $HOME/miniconda
     - export PATH="$HOME/miniconda/bin:$PATH"
-    - hash -r
-    - conda config --set always_yes yes
-    - conda update conda
-    - conda info -a
-    - travis_retry conda create -n test $CONDA pip jinja2 pandas mock six nose
+    - conda update --yes conda
+    - travis_retry conda create --yes -n test $CONDA pip jinja2 pandas mock six nose
     - source activate test
     - travis_retry pip install vincent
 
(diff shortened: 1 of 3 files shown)


ocefpaf:
OK. I finally pushed all the remaining changes I had in my branch.  They are all minor changes, but it fixes a few errors.

ocefpaf:
@BibMartin or @themiurgo can either of you review and merge?

themiurgo:
Two questions @ocefpaf:

- any particular reason why you changed the `{ ... }` with `dict( ... )`? In general, there is no accepted convention for this, but people prefer in general to use the former. While this is debatable for multiline dicts, it seems the first is more practical / legible for stuff that is supposed to be in one line:
```Python
county_df = pd.DataFrame({"FIPS_Code": county_codes}, dtype=str)
county_df = pd.DataFrame(dict(FIPS_Code=county_codes), dtype=str)
```

and even more in the case of empty dicts (`{}` vs `dict()`). The literal syntax is also more efficient (http://doughellmann.com/2012/11/12/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2.html).

- I also noticed you changed `Folium` to `folium` within docs. I would keep the name upperscore (as it is referenced in the the rest of the docs). If we want to change it, we need to do be consistent and change in all the docs through the whole codebase.

ocefpaf:
> any particular reason why you changed the { ... } with dict( ... )?

Nope.  I just like `dict()` better (my brain changes that automatically while I am reading the code).

>  is more practical / legible for stuff that is supposed to be in one line

True

> even more in the case of empty dicts (`{}` vs `dict()`). 

That is a case where I prefer `dict()` for clarity: like `set()`, `tuple()` and `list()`  anyways... I am not teaching code here so this does not apply ;-)

> The literal syntax is also more efficient 

Sold!

> I also noticed you changed Folium to folium within docs. Was this intended?

It was, but you are right.  Lets keep it consistent with the rest.  I do not want anyone to waste time changing this in the rest of the code.

themiurgo:
And it's a merge! :tada: Thanks!

Issue #131, opened 2015-05-18
Redundant popup_on?


themiurgo:
What's the reason for having the parameter `popup_on` on features? In which cases is it useful? To me it seems sensible that if a popup text is passed one wants a popup and otherwise it should be off.

At the very least we should enable a popup if text is passed and disable the popup if no text is passed. Provided that it is useful, the parameter can be maintained to force behaviour (for example showing popup when no popup text is passed and, vice versa, hiding even if text is passed).

ocefpaf:
Yes. Makes sense.

Actually, after reviewing one of your PRs recently, I came to the same
conclusion.  Not sure if @robstory had a reason for that design. We can
address that in a PR and keep discussing there.

(I am away from my laptop for while. If want to implement the PR go ahead,
otherwise I can do this once I get back.)
Em 18/05/2015 19:38, "Antonio Lima" <notifications@github.com> escreveu:

themiurgo:
I have a PR handy. :)

ocefpaf:
Thanks!
Em 19/05/2015 11:13, "Antonio Lima" <notifications@github.com> escreveu:

PR #133, opened 2015-05-19
Remove popup_on

themiurgo:
Solves #131

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -293,7 +293,7 @@ class Map(object):
         self.template_vars.setdefault('data_layers', []).append((data_string))
 
     @iter_obj('simple')
-    def simple_marker(self, location=None, popup='Pop Text', popup_on=True,
+    def simple_marker(self, location=None, popup=None,
                       marker_color='blue', marker_icon='info-sign',
                       clustered_marker=False, icon_angle=0, width=300):
         """Create a simple stock Leaflet marker on the map, with optional
@@ -308,8 +308,6 @@ class Map(object):
             or a tuple of the form (Vincent object, 'vis_path.json')
             It is possible to adjust the width of text/HTML popups
             using the optional keywords `width`.  (Leaflet default is 300px.)
-        popup_on: boolean, default True
-            Pass false for no popup information on the marker
         marker_color
             color of marker you want
         marker_icon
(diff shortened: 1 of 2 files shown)


folium/folium.py
```diff
@@ -676,11 +663,9 @@ def _popup_render(self, popup=None, mk_name=None, count=None,
             Type of marker. Simple, Circle, etc.
         count: int, default None
             Count of marker
-        popup_on: boolean, default True
-            If False, no popup will be rendered
         """
-        if not popup_on:
-            return 'var no_pop = null;'
+        if not popup:
+            return ''
```
ocefpaf: @themiurgo Quick questions before we merge this.  Why you choose an empty string instead of the `'var no_pop = null;'`  Will the marker be clickable with the empty string?  If so, that might be misleading.

themiurgo: In fact if there is no `popup` text, I do not set a javascript variable at all in the HTML template. The marker will not be clickable so no variable is set.

ocefpaf:
Thanks @themiurgo!  The PR looks good to merge.  I just want to test it locally and once I get back in the office I will merge it.

themiurgo:
Great, let me know if any issues pop out! :smile:

folium/folium.py
ocefpaf: > In fact if there is no popup text, I do not set a javascript variable at all in the HTML template.

In a local test the "click" hand appears when hovering a mark making the user think that he can click and see something.  I guess we do need to set `'var no_pop = null;'`.

popuptest

ocefpaf:
> Great, let me know if any issues pop out!

One did not :stuck_out_tongue:

folium/folium.py
themiurgo: In fact, in my tests, that behaviour happens anyways, even if `no_popup` is set. Try the same on the master branch and you'll see the hand popping up anyways, even if popup_on is set to False.

ocefpaf: Ah, so setting it to `null` does not change that.  Let me take a closer look at this later (I am still travelling...)

If that is what leaflet does then lets merge as is.

themiurgo: Sure. Here's an example what comes out from folium (master branch) with `popup_on=False` on a marker:
https://gist.github.com/themiurgo/170dbb742897f9b86da6#file-no_popup-html

themiurgo: See also here:
http://leafletjs.com/examples/quick-start.html

Third map, there are some elements without popup content. Hand still shows. :smile:

ocefpaf: OK, then. It is a Leaflet issue!  Lets merge...

Issue #126, opened 2015-05-12
Automatic location / zoom


themiurgo:
Now that we have `fit_bounds` implemented, we could make the initial `location` parameter optional, and have leaflet automatically figure out an optimal location / zoom to include all the added features (calling fit_bounds on an automatically computed bounding box). This would make plotting some unknown features on iPython notebook quite convenient:

    > fmap = folium.Map()
    > folium.simple_marker(location=...)
       # Here other features get added, and folium keeps track of what's the current bounding box
    > embed(fmap)

Is there any drawback about this behaviour? The only one I can think of, so far, is that it's not so trivial to get the bounding box of a geojson layer.

ocefpaf:
I am :+1: 

I don't see any obvious problems.  @birdage what do think?

birdage:
i like it!

ocefpaf:
And 3 makes a crowd! @themiurgo do you have a PR for that :wink: ?

PS: We should talk about adding you officialy to the project.

themiurgo:
> And 3 makes a crowd! @themiurgo do you have a PR for that :wink: ?

Not yet, but it will come soon. :smile: 

> PS: We should talk about adding you officialy to the project.

Great, hopefully I can be of help!

ocefpaf:
> Great, hopefully I can be of help!

We have only one rule.  Avoid self merges :stuck_out_tongue_winking_eye:

themiurgo:
Makes sense! 2 set of :eyes:  >> 1.

PR #134, opened 2015-05-19
Autobounds

themiurgo:
This attempts to implement #126 .

The PR is far from being ready, but I wanted to put it out so that I can read your suggestions.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -128,11 +128,13 @@ class Map(object):
         self.json_data = {}
         self.plugins = {}
 
-        # Location.
-        if not location:
-            raise ValueError('You must pass a Lat/Lon location to initialize'
-                             ' your map')
+        # No location means we will use automatic bounds and ignore zoom
         self.location = location
+
+        # If location is not passed, we center the map at 0,0
+        if not location:
+            location = [0, 0]
+            zoom_start = min_zoom
 
         # Map Size Parameters.
         try:
@@ -664,6 +666,31 @@ class Map(object):
         self.template_vars.update({'fit_bounds': fit_bounds_str})
 
 
+    def _auto_bounds(self):
+        if 'fit_bounds' in self.template_vars:
+            return
+        # Get count for each feature type
+        ft_names = ["marker", "line", "circle", "polygon", "multiline"]
+        ft_names = [i for i in ft_names if i in self.mark_cnt]
+
+        # Make a comprehensive list of all the features we want to fit
+        feat_str = ["{name}_{count}".format(name=ft_name,
+                                            count=self.mark_cnt[ft_name])
+                    for ft_name in ft_names
+                    for count in range(1, self.mark_cnt[ft_name]+1)
+        ]
+        feat_str = "[" + ', '.join(feat_str) + "]"
+
+        fit_bounds = self.env.get_template('fit_bounds.js')
+        fit_bounds_str = fit_bounds.render({
+            'autobounds': not self.location,
+            'features': feat_str,
+            'fit_bounds_options': json.dumps({'padding' : [30, 30]}),
+        })
+
+        self.template_vars.update({'fit_bounds': fit_bounds_str.strip()})
+
+
     def _popup_render(self, popup=None, mk_name=None, count=None,
                       popup_on=True, width=300):
         """Popup renderer: either text or Vincent/Vega.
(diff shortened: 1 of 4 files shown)


folium/folium.py
```diff
@@ -128,12 +128,14 @@ def __init__(self, location=None, width='100%', height='100%',
...
-            raise ValueError('You must pass a Lat/Lon location to initialize'
-                             ' your map')
+        # No location means we will use automatic bounds and ignore zoom
         self.location = location
 
+        # If location is not passed, we center the map at 0,0
+        if not location:
+            location = [0, 0]
+            zoom_start = min_zoom
+
```
ocefpaf: OK, I like this!  Get me a map no matter what :wink:

folium/folium.py
```diff
@@ -664,6 +666,31 @@ def fit_bounds(self, bounds, padding_top_left=None,
         self.template_vars.update({'fit_bounds': fit_bounds_str})
 
 
+    def _auto_bounds(self):
+        if 'fit_bounds' in self.template_vars:
```
ocefpaf: I am wondering if we should add an option here to fit just an specific feature, like markers, instead of everything.

Here is what got me thinking.  I have an application where I add several markers, lines, etc.  But I want the map to autofit only on the markers (ocean gliders that are moving in time).  That way the user will always get a map centred on them.

On the other hand I like to keep things simple and your current implementation is exactly what most users expect from an autofit.

I am on the fence here. @birdage what do you think?

tests/folium_tests.py
```diff
@@ -58,7 +58,8 @@ def test_init(self):
         assert self.map.location == [45.5236, -122.6750]
         assert self.map.map_size == {'width': 900, 'height': 400}
 
-        nt.assert_raises(ValueError, callableObj=folium.Map)
+        # TODO: assert that does not raise?
+        #nt.assert_raises(ValueError, callableObj=folium.Map)
```
ocefpaf: We do not need to test for this.  Lets just add a twin test `test_manual_bounds`.

folium/folium.py
```diff
@@ -664,6 +666,31 @@ def fit_bounds(self, bounds, padding_top_left=None,
...
+            return
+        # Get count for each feature type
+        ft_names = ["marker", "line", "circle", "polygon", "multiline"]
+        ft_names = [i for i in ft_names if i in self.mark_cnt]
+
+        # Make a comprehensive list of all the features we want to fit
+        feat_str = ["{name}_{count}".format(name=ft_name,
+                                            count=self.mark_cnt[ft_name])
+                    for ft_name in ft_names
+                    for count in xrange(1, self.mark_cnt[ft_name]+1)
```
ocefpaf: It is fine to use range here. I do not believe it will ever be a memory problem in python2 and in python3 range is a generator.

themiurgo: :+1:

tests/folium_tests.py
themiurgo: :+1:

folium/folium.py
themiurgo: I thought of this too and it'd be cool to have. Since folium is basically built on top of Leaflet, I'd say it's good to follow closely what Leaflet does.

Leaflet allows to fit bounds generically and allows to get bounds from any feature or FeatureGroup. In a future implementation where we'll be able to express featuregroups or layers easily in folium, I'd love to include this, but at the moment I'm for leaving things simple. :smile:

ocefpaf: Agreed.  Lets keep it simple then and open an issue with this idea so we do not get lost.

tests/folium_tests.py
themiurgo: We already have a `test_fit_bounds` for which we test the manual bounds. What did you have in mind?

themiurgo:
Ok, now test passes. Seems good to go.

ocefpaf:
@themiurgo Great job!  Squashed away and I'll merge.

themiurgo:
@ocefpaf Squashed!

ocefpaf:
Thanks @themiurgo!  We need another PR updating the Changelog!  (I can do that once I get back.)

themiurgo:
@ocefpaf thanks to you! We should update also the docs at some point. Opening an issue.

Issue #145, opened 2015-07-10
Implementing a good way to handle plugins


BibMartin:
Following the discussion with @themiurgo and @ocefpaf in PR https://github.com/python-visualization/folium/pull/144.

@themiurgo has set the question :
> [This also shows the importance of implementing soon a good way to handle plugins. That way we might have a generic add_plugin(plugin_object) for all the plugins and extra functionalities like this, and no need to make changes to the core object Map and its interface]

BibMartin:
I imagine a plugin object can have 3 or 4 methods :
* `render_css` where one can define plugin-specific styles
* `render_html` where one defines new DOM objects
* `render_js` where one defines how the plugin interacts with the maps
* eventually `render_header` where one can define external javascript imports ; yet this could be done into the HTML part (is it ugly?).

You can see examples of this mindset in https://github.com/BibMartin/folium/commit/5238fb98585c1b5482b6a485ee78049838ff33bd or in https://github.com/BibMartin/folium/commit/1a0469266a22ac421e9e40cb5f68a20fc748f8ae.

Maybe the plugins may inherit from a common object that implments these methods so that you don't have to create the 3-4 methods if the plugin does not require it.

themiurgo:
Thanks a lot! This is great, some other help with to lift the plugin effort. :)

Do you think these 3/4 methods are enough to implement all the plugins shown in #127 . If so, we might have a candidate PR from your concepts commits and then iterate from that.

themiurgo:
Also, I like a lot BibMartin@1a04692 . Did you choose the MarkerCluster example on purpose or randomly? MarkerCluster is already present in the folium base template, along with other functionalities that pollute the template. It would be great to handle it as a plugin instead, and keep the base template for the Leaflet  core functionalities.

BibMartin:
> Do you think these 3/4 methods are enough to implement all the plugins shown in #127 ?

I was having a look at them. For some (like esri, terminator, etc) I think this shall be quite easy provided that the generate code follows the scheme `header-stylesheet-body-script`. I'm not sure for all of them now, but we can start with the easiest ones and adapt the scheme when needed, don't you think so ?

> Did you choose the MarkerCluster example on purpose or randomly?

I did not know that it was already present but it looked like a good example on "how to develop a plugin".

themiurgo:
Ok, I'd say let's start a "plugin" branch here, starting from your commits, and collaborate on that. Main goals:
- perfection the plugin skeleton;
[...]

PR #146, opened 2015-07-11
Plugins ScrollZoomToggler and MarkerCluster

BibMartin:
As discussed in #145 we shall try to put as much code as possible into plugins.

This is two first example on how we could do this.


themiurgo:
Could you please squash the commits in a single one?

BibMartin:
This PR is replaced by #148 where everything is in a single commit.

themiurgo:
Thanks! For future reference, here's how to squash commits:
http://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Issue #105, opened 2015-04-21
Merge fol_template.htm and geojson_template.html


ocefpaf:
@birdage While investigating #30 I found out that several changes to `fol_template.html` are not in `geojson_template.html`, like the custom and clustered markers.

That makes me wonder if we should have only one base template and a sub-template for geojson (not sure how to do that), or only one template that works for both even if we load unnecessary java-stuff.

birdage:
this is correct i just updated the base map fol template when i made the changes, i think you are right we should have one template for these things.

themiurgo:
I agree. I think this is a result of the fact that geojson is considered in folium to be used mostly as an overlay (from the docs: "Folium allows you to plot a GeoJSON or TopoJSON overlay on the map").

This is generally false, and recently it's not uncommon to have find marker features, lines, and polygons encoded as geojson (as shown in [geojson.io](http://geojson.io) and [GitHub](
https://help.github.com/articles/mapping-geojson-files-on-github/), for example).

I think it would be good to slowly change this aspect of folium and handle geojson similarly to any other feature, following closely Leaflet behaviour (http://leafletjs.com/examples/geojson.html).

ocefpaf:
>  it's not uncommon to have find marker features, lines, and polygons encoded as geojson (as shown in geojson.io and GitHub, for example).

@themiurgo that sounds interesting and, after reading #120, I can see the appeal.  The HTML looks much clearer.  However, if I understood it correctly, all GeoJSON would be embedded in the HTML, right?  I guess that some people appreciate the fact that can have a separate file for the GeoJSON.  Still, that does not stop us from implementing something for makers, lines, polygons and etc.

> I think it would be good to slowly change this aspect of folium and handle geojson similarly to any other feature, following closely Leaflet behaviour (http://leafletjs.com/examples/geojson.html).

Agreed.  Is that something you would like to tackle?

themiurgo:
> @themiurgo that sounds interesting and, after reading #120, I can see the appeal. The HTML looks much clearer. However, if I understood it correctly, all GeoJSON would be embedded in the HTML, right? I guess that some people appreciate the fact that can have a separate file for the GeoJSON. Still, that does not stop us from implementing something for makers, lines, polygons and etc.

[...]

PR #147, opened 2015-07-11
Rationalize Templates

themiurgo:
This is a first attempt at rationalising folium templates, leveraging Jinja's awesome features (http://jinja.pocoo.org/docs/dev/templates/#template-inheritance) and avoid code duplication and bugs (such as #105).

@ocefpaf could you take a look at this once you come back from SciPy? :smiley: 

I have tested the simple folium template and it seems to be working as normal. We should also test the geojson and see if it has any problems.

### DIFF
--- a/folium/templates/base.html
+++ b/folium/templates/base.html
@@ -1,0 +1,138 @@
+<!DOCTYPE html>
+<head>
+    {% block head %}
+    <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.3/leaflet.js"></script>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.3/leaflet.css" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js" charset="utf-8"></script>
+    {% endblock head %}
+
+    <style>
+      {% block style %}
+      html, body {
+        width: 100%;
+        height: 100%;
+        margin: 0;
+        padding: 0;
+      }
+      {% endblock style %}
+    </style>
+
+</head>
+
+<body>
+   {% block mapdiv %}
+   <div class="folium-map" id="{{ map_id }}" {{ size }}></div>
+   {% endblock mapdiv %}
+
+   <script>
+      {% block jscode %}
+
+      {{ dvf_js }}
+      {{ d3 }}
+      {{ vega }}
+      {{ jquery }}
+
+      {{ vega_parse }}
+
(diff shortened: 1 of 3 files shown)


themiurgo:
(Btw the PR obviously is not ready yet, just submitted to have a head start ;) )

BibMartin:
This much more elegant than the `render_html` methods etc.

I'm a bit afraid of two things :
* Jinja stuff is quite abstract ; won't it discourage begginners (like I) to get into programming plugins ?
* I cannot figure out how the code would behave if I wanted to combine several plugins. Like for example:

        map = folium.Map()
        map.add_plugin(MarkerCluster(...))
        map.add_plugin(ScrollZoomToggler(...))
        map._repr_html_()

themiurgo:
> Jinja stuff is quite frightening ; won't it discourage begginners (like I) to get into programming plugins ?

Jinja was made for desiners and in general it was made to be simpler than Python coding. I think it should be enough to write that docs page

> I cannot figure out how the code would behave if I wanted to combine several plugins. Like for example:

The code I've pushed does not deal with plugins so I can't answer that. I will answer your question about plugins in the other thread.

ocefpaf:
I like the idea of having a `base` and injecting that into the `geojson` and any other plugins.  However, I am unsure how the current plugin plan fit in here.  @themiurgo can you rebase this?

@BibMartin can you take a second pass at this in light of your plugin branch?

themiurgo:
I'll take a look at it this or the coming weekend (sorry, deadlines and holidays to sort out first) :stuck_out_tongue_closed_eyes:

ocefpaf:
How are we on this?  Is it worth re-basing and merging or does #170 changed the game?

I am sorry for not participating on this as much as I should.  But I got some free time and I want to fix that :wink:

ocefpaf:
ping!

themiurgo:
It's probably good to see what @BibMartin has to say about this, whether this PR still makes sense to be merged and how it relates to the last changes in the codebase.

ocefpaf:
@BibMartin I guess that this does not fit in the latest code, correct?

@themiurgo are you OK closing this?

PS: Before closing let's just check if we can steal any ideas  :wink:

BibMartin:
It does not fit in the lastest code, but the concern of rationalizing templates is still relevant.

In the latest code, most template code is written inside python code (with triple quotes) and most templates are either not used or used in tests only. Though this is not uniformely true. I think a coming question is : do we want to generalize this mindset, or to separate python code and templates ?

ocefpaf:
> I think a coming question is : do we want to generalize this mindset, or to separate python code and templates ?

I like the templates inside the code like you are doing.  Life is much easier that way :wink: 

Unless someone brings a good reason not to do so let's move in that direction.

We need a PR to remove the unused templates and/or updated the tests.

Closing this.

themiurgo:
Yes, I was ok closing this. Perhaps in the future we'll do again something similar.

ocefpaf:
> Perhaps in the future we'll do again something similar.

:+1:

PR #148, opened 2015-07-11
plugin objects definition, with two examples

BibMartin:
As discussed in #145 we shall try to put as much code as possible into plugins.

This is two first example on how we could do this.

This PR replaces #146 : I couldn't find how to stash all the commits in one without creating another PR.


ocefpaf:
Awesome!  Thanks @BibMartin for the PR and @themiurgo for the review!!

PR #151, opened 2015-07-17
Plugins object and a few plugins

BibMartin:
A few devs based on #145. We have :

* A `Plugin` that other plugins can inherit from.
* A `TemplatePlugin` object to develop simple plugins with jinja2.
* A few examples (BoatMarker, Terminator, Layer).
* A notebook of examples for the plugins.


andrewgiessel:
FWIW, Build fails due to lack of urllib2 on py3.3 and py3.4

BibMartin:
I think this is ready for merge. @themiurgo, @ocefpaf do you think of a next step ? can you merge it, please ?

ocefpaf:
Outstanding work @BibMartin!! I am actually happy I procrastinated doing this for so you could come along and do it much better than I envisioned :stuck_out_tongue_winking_eye:

ocefpaf:
@BibMartin I just added you to folium team.  We have only one rule: Avoid self-merges are much as you can.  Welcome and thanks for this awesome PR!

BibMartin:
Thanks to you ;-)

Issue #149, opened 2015-07-14
Display of raster data


andrewgiessel:
I would like to take generate a numpy array and display it as a semi-transparent layer on a leaflet map via folium.  What is the best way to do this?  I recognize that this simple sounding request could actually be quite challenging, but I'm willing to make the effort.

Can anyone point me in the right direction?  Do I really need to generate a full blown tile set using a separate webserver like TileCache?

ocefpaf:
@andrewgiessel I am unsure about the best course to implement something like this.  However, if your image is small enough, you can try `imshow`+`mplleaflet`

ocefpaf:
OK, I tried my own advice and found out it does not work.  `mplleaflet` uses `mplexporter` under the hood, and `mplexporter` does not transform images (yet).

I went ahead and hacked a simple "image overlay" that could be easily added to folium's API:

https://ocefpaf.github.io/python4oceanographers/blog/2015/07/13/interactive_geo/

Do you want to try to implement that?

andrewgiessel:
Wow, this is great!  Sorry I didn't respond earlier.  I am going to try to re-implement this locally and then probably take a crack at a PR.  One thing I was wondering was about transparency and that's possible!  My only question is how this might perform on a large scale (global) since it's basically not tiles.  This is very encouraging, though, thank you for the effort and I'll report back.

Including this for reference:
http://leafletjs.com/reference.html#imageoverlay

ocefpaf:
> My only question is how this might perform on a large scale (global) since it's basically not tiles.

That will depend on the image.  However, if you can set a WMS up you can easily add that layer in a folium instance.

andrewgiessel:
I might be forced to do that, but would like this for rapid visualization of geospatial fields, so if i can get it working with something simple for local purposes (even if it's not performant) that'd be great.

So, I got your example to work!  However, it doesn't seem to work with local files.  As far as I can tell [here](https://github.com/Leaflet/Leaflet/blob/master/src/layer/ImageOverlay.js#L114), Leaflet just makes an <img> in the DOM.  I thought that this should work with a file:// url, but no luck yet.  Ideas?

andrewgiessel:
This is a slightly separate issue, but when I try to open the generated html in another tab, I get a ton of errors because all the external javascript and css references don't start with `http:`.  Shouldn't this be changed?  I can open a PR for that, later.

ocefpaf:
[...]

PR #152, opened 2015-07-17
Add image overlay

andrewgiessel:
This PR addresses #149.

As of submission, it is not complete.  I think I am adding layers correctly and put in the right template form, but it is not adding the layer to the rendered map.

@ocefpaf What am I missing?

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -23,7 +23,7 @@ from folium import utilities
 from folium.six import text_type, binary_type, iteritems
 
 import sys
-
+import base64
 
 ENV = Environment(loader=PackageLoader('folium', 'templates'))
 
@@ -221,6 +221,7 @@ class Map(object):
         self.added_layers = []
         self.template_vars.setdefault('wms_layers', [])
         self.template_vars.setdefault('tile_layers', [])
+        self.template_vars.setdefault('image_layers', [])
 
     @iter_obj('simple')
     def add_tile_layer(self, tile_name=None, tile_url=None, active=False):
@@ -281,18 +282,14 @@ class Map(object):
         data_string = ''
         for i, layer in enumerate(self.added_layers):
             name = list(layer.keys())[0]
-            data_string += '\"'
-            data_string += name
-            data_string += '\"'
-            data_string += ': '
-            data_string += name
             if i < len(self.added_layers)-1:
-                data_string += ",\n"
+                term_string = ",\n"
             else:
-                data_string += "\n"
+                term_string += "\n"
+            data_string += '\"{}\": {}'.format(name, name, term_string)
 
         data_layers = layers_temp.render({'layers': data_string})
-        self.template_vars.setdefault('data_layers', []).append((data_string))
+        self.template_vars.setdefault('data_layers', []).append((data_layers))
 
     @iter_obj('simple')
     def simple_marker(self, location=None, popup=None,
@@ -953,6 +950,92 @@ class Map(object):
         self.template_vars.setdefault('geo_styles', []).append(style)
         self.template_vars.setdefault('gjson_layers', []).append(layer)
 
+    @iter_obj('image_overlay')
+    def image_overlay(self, data, opacity=0.25, min_lat=-90.0, max_lat=90.0,
+                      min_lon=-180.0, max_lon=180.0, image_name=None, filename=None):
+        """Simple image overlay of raster data from a numpy array.  This is a lightweight
+        way to overlay geospatial data on top of a map.  If your data is high res, consider
+        implementing a WMS server and adding a WMS layer.
+
+        This function works by generating a PNG file from a numpy array.  If you do not
+        specifiy a filename, it will embed the image inline.  Otherwise, it saves the file in the
+        current directory, and then adds it as an image overlay layer in leaflet.js.
+        By default, the image is placed and stretched using bounds that cover the 
+        entire globe.
+
+        Parameters
+        ----------
+        data: numpy array OR url string, required.  
+            if numpy array, must be a image format, i.e., NxM (mono), NxMx3 (rgb), or NxMx4 (rgba)
+            if url, must be a valid url to a image (local or external)
+        opacity: float, default 0.25
+            Image layer opacity in range 0 (completely transparent) to 1 (opaque)
+        min_lat: float, default -90.0
+        max_lat: float, default  90.0
+        min_lon: float, default -180.0
+        max_lon: float, default  180.0
+        image_name: string, default None
+            The name of the layer object in leaflet.js
+        filename: string, default None
+            Optional file name of output.png for image overlay.  If None, we use a 
+            inline PNG.
+
+        Output
+        ------
+        Image overlay data layer in obj.template_vars
+
+        Examples
+        -------
+        # assumes a map object `m` has been created
+        >>> import numpy as np
+        >>> data = np.random.random((100,100))
+        
+        # to make a rgba from a specific matplotlib colormap:
+        >>> import matplotlib.cm as cm
+        >>> cmapper = cm.cm.ColorMapper('jet')
+        >>> data2 = cmapper.to_rgba(np.random.random((100,100)))
+
+        # place the data over all of the globe (will be pretty pixelated!)
+        >>> m.image_overlay(data)
+
+        # put it only over a single city (Paris)
+        >>> m.image_overlay(data, min_lat=48.80418, max_lat=48.90970, min_lon=2.25214, max_lon=2.44731)
+        
+        """
+
+        if isinstance(data, str):
+            filename = data
+        else:
+            try:
+                png_str = utilities.write_png(data)
+            except Exception as e:
+                raise e
+
+            if filename is not None:
+                with open(filename, 'wb') as fd:
+                    fd.write(png_str)
+            else:
+                filename = "data:image/png;base64,"+base64.b64encode(png_str).decode('utf-8')
+
+        if image_name not in self.added_layers:
+            if image_name is None:
+                image_name = "Image_Overlay"
+            else:
+                image_name = image_name.replace(" ", "_")
+            image_url = filename
+            image_bounds = [[min_lat, min_lon], [max_lat, max_lon]]
+            image_opacity = opacity
+            
+            image_temp = self.env.get_template('image_layer.js')
+
+            image = image_temp.render({'image_name': image_name,
+                                       'image_url': image_url,
+                                       'image_bounds': image_bounds,
+                                       'image_opacity': image_opacity})
+
+            self.template_vars['image_layers'].append(image)
+            self.added_layers.append(image_name)
+        
     def _build_map(self, html_templ=None, templ_type='string'):
         self._auto_bounds()
         """Build HTML/JS/CSS from Templates given current map type."""
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,13 @@ before_install:
     - wget http://bit.ly/miniconda -O miniconda.sh
     - bash miniconda.sh -b -p $HOME/miniconda
     - export PATH="$HOME/miniconda/bin:$PATH"
+    - hash -r
+    - conda config --set always_yes yes
     - conda update --yes conda
-    - travis_retry conda create --yes -n test $CONDA pip jinja2 pandas mock six nose
+    - conda info -a
+    - travis_retry conda create -n test $CONDA pip jinja2 pandas mock six nose
     - source activate test
     - travis_retry pip install vincent
-
 install:
     - python setup.py install
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,7 @@
 - Added cartodb positron and dark_matter tiles (ocefpaf d4daee7)
 - Forcing HTTPS when available. (ocefpaf c69ac89)
 - Added Stamen Watercolor tiles. (ocefpaf 8c1f837)
+- Added Image Overlay. (andrewgiessel ---)
 
 Bug Fixes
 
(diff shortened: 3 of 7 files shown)


folium/folium.py
```diff
@@ -16,6 +16,8 @@
 import json
 from uuid import uuid4
 
+import imageio
```
ocefpaf: I think its best to import this inside `def image_overlay()` to keep this an "optional" dependency (same way we deal with pandas.)

andrewgiessel: Sounds good - the reason the travis builds fail is because of this dependancy.

folium/folium.py
```diff
@@ -948,6 +950,40 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
         self.template_vars.setdefault('geo_styles', []).append(style)
         self.template_vars.setdefault('gjson_layers', []).append(layer)
 
+    @iter_obj('image_overlay')
+    def image_overlay(self, data, image_name=None, filename='data.png',
+                      lower_left=[-90,-180], upper_right=[90,180], opacity=0.25):
```
ocefpaf: I do not like that leaflet is alphabetical order `(lat, lon)` instead of geometrical order `(lon, lat)`.  Also, by aggregating into `lower_left` and `upper_right` tuples things might get even more confusing.

We have to write a good docstring describing the inputs to solve the confusion. But also consider using: `min_lat=-90, max_lat=90, min_lon=-180, max_lon=180`.  I know it is verbose, but it is unambiguous (and we already use it in the main class).

folium/folium.py
```diff
@@ -948,6 +950,40 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
...
+        """Simple image overlay of raster data from URL or numpy array.
+
+        """
+
+        if isinstance(data, str):
+            pass
+        try:
+            imageio.imwrite(filename, data)
+        except:
+            raise ValueError('Data is not proper image format (NxM, NxMx3 or NxMx4)')
```
ocefpaf: I do not know `imageio` very well to know all the exceptions that might occur.  Maybe a patter that catches `Exception as e` and re-raises it might be better than a `ValueError`.  (Bare in mind that I never did this and I am unsure if it works the way I am imagining :wink:)

folium/folium.py
```diff
@@ -948,6 +950,40 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
...
+
+        if image_name not in self.added_layers:
+            if image_name is None:
+                image_name = data
+            else:
+                image_name = image_name.replace(" ", "_")
```
ocefpaf: Have to do this more that I would like.... maybe it is time for a normalization function...  do not worry about this in this PR.

ocefpaf:
@andrewgiessel nice work!  I made a few inline comments.  As a first pass we need:

1) Docstrings
2) A test
3) Update the `CHANGE.TXT` file

folium/folium.py
ocefpaf: There will be (probably) other failures that are unrelated.  That is something I have on my plate to fix.  Ignore them for now...

BTW add `imageio` to our `.travis.yml`!

andrewgiessel:
Great!  I agree with all of your comments, especially the min_ / max_ arguments- they are much more clear.  I'll make those changes, and the 4 other things:  docs, test, change.txt and travis.xml.

When I run this locally, the generated HTML file doesn't have the have anything in the overlay layer.  I'll look into it, but is there something we need add when the map is rendered to include the overlaid images?

folium/folium.py
andrewgiessel: so so so agree.  I hate the inconsistency in the lat/long order!  I always try to do geometric, too.

andrewgiessel:
also- is there a type [here](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L294-L295)?  Shouldn't we add data_layers not data_string?

andrewgiessel:
Ok, update.  I realized the thing I needed to do was to edit the main html template!  Derp.  Then, I thought I needed the overlay with the other layers (which can work, but isn't needed and was hard to get it to turn on by default).  So, I looked again at what you did in the html and realized it needed to go after the map creation call, and now it works.

TODO still: docstring, test, changes, maybe readme.md (!).

I did a small refactor of the data_layer string/template, but it doesn't seem to do anything and always seems empty.  I can revert that edit if you want, but if you could look at `add_tile_layer()` I think that would be a good thing.

The only major outstanding thing is the bug when you have a bit image and zoom a lot.  I'm sure you'll see what I mean when you try it.  It might not be worth fixing, and instead let upstream (leaflet) fix it.

andrewgiessel:
image

ocefpaf:
This is looking good!  Thanks @andrewgiessel!!

I am having some connection issues this, week so I cannot review this until next Friday.

themiurgo:
Yes, this looks great @andrewgiessel ! :+1:

andrewgiessel:
Thanks @ocefpaf + @themiurgo !  I am really happy with it too.  The one thing I wanted to do was look into going from NxM to NxMx3 using a MPL colormap.  might roll my own simple external lib.

No problem re timing @ocefpaf.  I'll hopefully have everything 'merge-ready' by then.

andrewgiessel:
Ah, just a little research into MPL means that this is already taken care of, and that we don't need to support anything fancy, because it's out of scope.  'image_overlay' will take a png, and that can be an rpg

For reference:

    import matplotlib.cm as cm
    import numpy as np

    converter = cm.ColorMapper('jet')
    imageio.imwriter('test.png', converter.to_rgba(np.random.random((100,100)))

themiurgo:
That's great! It would be great to integrate this into the docs, as it might be out of reach to those not familiar with matplotlib.

Would you mind writing a few lines in the docs? If not that's totally cool and we'll take care of it some other time. Thanks again!

andrewgiessel:
@themiurgo you bet, writing the docstrings now.

This is going to be a nice feature that is intermediate to a full blown WMS server.  We should update README to highlight an example.

andrewgiessel:
Docstring updated.  What version should this go under in CHANGES.txt?

andrewgiessel:
Ok, added a passing test!  I think that the only thing left is to add something to CHANGES.TXT and perhaps readme.rst.  The README looks like it could use a general refresh and so this might be very close indeed.

andrewgiessel:
Oddly, the test fails in travis, but not locally.  Looking into it now.

andrewgiessel:
Ok, the travis build succeeds on the new test.  The error is in fit_bounds - the dictionary is not ordered and thus the rendering happens in a random way.  It passes on 2.7 and 3.4 but not on 3.3.

Output:

    map.fitBounds([[52.193636, -2.221575], [52.636878, -1.139759]],
        {"padding": [3, 3], "maxZoom": 15}
    );
    map.fitBounds([[52.193636, -2.221575], [52.636878, -1.139759]],
        {"padding": [3, 3], "maxZoom": 15}
    );

andrewgiessel:
One thought I had:  does image_overlay to projection to web mercator?

BibMartin:
hi @andrewgiessel ; that looks like great work !

I'm sorry for being late in the discussion.

A few thoughts :

* Maybe this kind of enhancement could be added as plugins (see #151) so that it does not complexify `folium.py` and `fol_template.html`.
* In Leaflet, you can use `imageOverlay` with either a url or a base64 string (like `L.imageOverlay("data:image;base64,iVBORw0K....5ErkJggg==", ...);`). This would be interesting to enable such embedings for those who wants a fully autonomous html (useful for example if you want to embed your folium map inside of an iframe in a notebook.)
* Based on http://stackoverflow.com/questions/902761/saving-a-numpy-array-as-an-image, it may be possible to avoid the dependency to `imageio` (there's an answers defining a function `saveAsPNG` using only the standard library). But I wonder whether it would not be simpler to ask for an image file/object and let the user convert his array to an image outside of folium...?

Again I'm late and sorry for this ; maybe I'll do another PR after your merge to integrate this. But I'd be interested in having your feeling about it.

andrewgiessel:
@BibMartin Thanks!  

* I have to admit I'm not 100% clear re: plugins in folium, although modularity is obviously good. I'd be happy to refactor but it seems like you are on a roll there- maybe after this?

* PNG as a base64 string is a *fantastic* idea as is removing the `imageio` dependency (although I am frankly very impressed with `imageio`).  As stated above, referencing external styles is tricky if there is a mis-match between the current path (in the python kernel) and the current URL path.

andrewgiessel:
As as aside, I mashed around enough last wee to get some tiles rendered to disk using `gdal2tile.py`.  It was not an easy requirement stack to sort out and the tiles really took a long time to generate.  However, the multi-res zoom was obviously awesome and the projections seemed correct.

I think this PR is a good quick and dirty solution, however, I would like feedback about how bad any projection artifacts might be.

Also, any domain experts want to break down the options for tile servers? =)

andrewgiessel:
@ocefpaf @BibMartin @themiurgo 

Aside from the optional feature of doing embedded PNGs, the only thing I think that remains is the CHANGES.txt file.  Let me know what I need to do to move forward!  Thanks!

ocefpaf:
@andrewgiessel:

> One thought I had: does image_overlay to projection to web mercator, or spherical or whatever the basemap is in?

I believe leaflet is always "web mercator".  [Unless...](https://github.com/kartena/Proj4Leaflet)

> The error is in fit_bounds.

Yep.  I ~~have~~ had a fix sitting here, but you beat me to it :stuck_out_tongue_winking_eye:

- Go ahead with the `CHANGES.txt`
- Squash the commits
- And...

I like `imageio` and "clean" HTMLs (no base64 noise) .  But, but, but... **practicality beats purity**.  So one less dependency+stand alone HTMLs is better.

Can you implement @BibMartin suggestions (2 and 3).  If not please open issues for 2 and 3, so we do not lose track.

> As as aside, I mashed around enough last wee to get some tiles rendered to disk using gdal2tile.py

- Awesome!  But lets think about this in another PR.  We have use for simple overlays and tiles.

@BibMartin:

- Regarding your suggestion 1:  I would love to see this implemented as plugins.  But lets do that later and get @andrewgiessel contribution as is.  (Can either one of you open an issue to remind us?)

---
Thanks @themiurgo for the review!
Thanks @BibMartin for the important suggestions!
And thanks @andrewgiessel for the PR!

:tada:  Hurray community!!! :tada:

BibMartin:
@andrewgiessel 
> I'd be happy to refactor but it seems like you are on a roll there- maybe after this?

I agree ; let's finish up on both sides.

@ocefpaf 
> I like imageio and "clean" HTMLs (no base64 noise) . But, but, but... practicality beats purity. So one less dependency+stand alone HTMLs is better.

`imageio` and `base64` are complementary to me. One shall have the choice between using `imageio` or providing a `png` generated elsewhere. And one shall have the choice between linking to a `png` on the disk or embeding it with `base64`.
But again, I imagine this can be done later ; there's already quite a lot of awesome work in this PR.

andrewgiessel:
I wanted to share this gist here.  It takes a numpy array, converts it to a byte array (which is really a string), then both encodes as a PNG.  That buffer can be directly saved to disk or encoded in base64 for direct embedding in HTML

https://gist.github.com/andrewgiessel/d5e1a11604af4aee5962

I'm about to use this to do the changes we talked about above, and then fix the build errors I caused by merging in upstream.  At that point we should be good to merge.

I'll make issues for any outstanding stuff, too.

themiurgo:
Great, thanks a lot! :smile:  Once done with all the outstanding issues, please squash all the commits. If you're unsure about how to do it, please let us know and we'll guide you.

andrewgiessel:
I generally don't squash commits so if you have a URL that'd be handy, otherwise I'll google a bit.

themiurgo:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

tl;dr Identify the HEAD of the first, commit and then run `git rebase -i HASH`. Write `pick` in the first commit line, `squash` in all other commit lines.

ocefpaf:
@andrewgiessel one important thing to note is that you won't be able to `push` after the squash.  You'll have to `push -f` (force push).  That is OK in branches.

andrewgiessel:
DING.  green!  I am going to do some manual tests and then try and squash.

andrewgiessel:
@themiurgo sorry for the ignorance, but how do i find the HEAD of the first commit?  I tried to use the hash of the first commit of the PR, but I think that is one too late (the pick/squash options started after that)

do i need to look in the output of `git log`?

Just don't want to lose work here... and trying to get better at git =)

BibMartin:
You can use `git log`, but you can also wtch there : https://github.com/andrewgiessel/folium/commits/add-image-overlay

andrewgiessel:
I think I did it?

themiurgo:
@andrewgiessel you did it! it would be great if you could change the commit message (shorten it to one line), to something that explains what the commit / PR is about. You can change the commit message by performing again a rebase (git rebase -i) and use the command `reword`.

ocefpaf:
Shoot!!

@andrewgiessel  Now you need to rebase against master (not your fault, we just merged some stuff).  If you want to keep your git-fu road to mastery I can give you some pointer.  Otherwise I can  do that for you in a PR to your branch.

andrewgiessel:
@themiurgo I shortened it using `git commit --amend` and a `git push -f`

@ocefpaf I'll try a rebase on master now...

themiurgo:
I checked out locally your branch, just in case something goes wrong.

andrewgiessel:
... good?  the places i had to change things for the master rebase were in .travis.yml and the tests.

themiurgo:
Everything seems good, it's a merge! :tada: Thanks for the nice contribution @andrewgiessel !

BibMartin:
Congrats @andrewgiessel, @themiurgo ; good contribution (and tricky git moves)

andrewgiessel:
:tada:  :tada:  :tada:  :clap: :clap: :clap: :sunglasses: :sunglasses: :sunglasses:

PR #154, opened 2015-07-19
Enable Travis' container-based infrastructure.

QuLogic:
Since you're using conda, there's pretty much no reason to use `sudo`. This should start and build a little faster.


themiurgo:
Thanks!

PR #166, opened 2015-08-05
Add support for leaflet div markers to folium

shankari:
This allows us to generate polylines that display intermediate points with popups. Updated the tests as well. Tests pass.
    
        bash-3.2$ PYTHONPATH=./tests ipython
        Python 2.7.10 |Anaconda 2.0.1 (x86_64)| (default, May 28 2015, 17:04:42)
        Type "copyright", "credits" or "license" for more information.
    
        IPython 2.1.0 -- An enhanced Interactive Python.
        Anaconda is brought to you by Continuum Analytics.
        Please check out: http://continuum.io/thanks and https://binstar.org
        ?         -> Introduction and overview of IPython's features.
        %quickref -> Quick reference.
        help      -> Python's own help system.
        object?   -> Details about 'object', use 'object??' for extra details.
    
        In [1]: import folium_tests
    
        In [2]: tester = folium_tests.testFolium()
    
        In [3]: tester.setup()
    
        In [4]: tester.test_div_markers()

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -192,6 +192,7 @@ class Map(object):
                               'stamenterrain', 'stamentoner',
                               'stamenwatercolor',
                               'cartodbpositron', 'cartodbdark_matter']
+
         self.tile_types = {}
         for tile in self.default_tiles:
             tile_path = 'tiles/%s' % tile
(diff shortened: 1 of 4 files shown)


BibMartin:
Thanks @shankari, that looks great! Do you think it's possible to implement it a a plugin (see `plugins` branch), to avoid complexifying `fol_template.html` ?

ocefpaf:
:+1: to the new feature.  We have to put the new folium logic out there to avoid asking people to re-do things!!  (Sorry @shankari!)

shankari:
@BibMartin, it is possible, but it might take a while. My current patch meets my needs for visualization, and the visualization is not a core part of what I'm working on - I'm using it to debug something else. I was happy to take the little extra time to clean up my code and contribute back, but I am not sure that I have the time to work on this any further right now.

@ocefpaf, yes, new logic out there would be great. I think that people who are not actively working on folium are going to work off your existing logic/structure (I know that I did).

BibMartin:
@shankari No problem ; I'll go for it. You've already done a great part of the job.

@ocefpaf Shall we merge this PR as is, or create another one that would meet the new machinery ?

ocefpaf:
The PR came with its own test!  How can I say no to that :smile_cat:? Merge away @BibMartin.

ocefpaf:
Thanks for the contribution  @shankari!

themiurgo:
I am against this PR as it is, but it can be changed to avoid code duplication / API complications.

Our #1 rule must be to avoid duplication and to follow closely what Leaflet is doing. So, instead of having yet another object (div icon), we should incorporate this into markers. This is what Leaflet does with divicons, it doesn't use a separate method / code, it just levearges `marker`.

http://jsfiddle.net/erictheise/k6ckm/

ocefpaf:
@themiurgo that is an important point.  But the :ship: has :boat:

themiurgo:
Sorry, it wasn't when I started to write the message.

Well it was a quick :ship:, it only took 8 hours :stuck_out_tongue_winking_eye:

PR #167, opened 2015-08-05
add custom css link

andrewgiessel:
This is related to https://github.com/Leaflet/Leaflet/issues/2282.

Basically, there is a directive in the 0.7.3 css that makes image overlays shift and disappear at high zoom levels.  It has been fixed upstream but is present in the stable release of leaflet.

The fix is simple:  https://github.com/Leaflet/Leaflet/issues/2282#issuecomment-30439260

Then we need to host our own version of the css and change the main folium template.  I've done this at https://rawgit.com/andrewgiessel/leafletstuff/master/leaflet.css.  Maybe we want to host this in the main folium repo, but it's a bit chicken/egg until it's in master.

### DIFF
--- a/folium/templates/fol_template.html
+++ b/folium/templates/fol_template.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <head>
    <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
-   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.3/leaflet.css" />
+   <link rel="stylesheet" href="https://rawgit.com/andrewgiessel/leafletstuff/master/leaflet.css" />
    <script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.3/leaflet.js"></script>
 
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>


BibMartin:
Great @andrewgiessel ! I had noticed the bug also, but didn't make the effot.

I would vote for hosting the css in folium repo. As a matter of officiality/stability, `leaflet repo> folium repo > personal repo`

ocefpaf:
I am more inclined to wait for a new Leaflet release.  This is a bug in a dependency and I see hosting `leaflet.css` here as forking.  (In the old bad meaning of the word forking.)

NB I am write this without the knowledge of leaflet release schedules.  If a new release takes too long we might be forced to host this here...

andrewgiessel:
The leaflet beta was announced on July 15 and I think it is getting close.
I don't like hosting it either, so I tried using the css in leaflet/master.
It breaks current leaflet. Then I tried using the beta release completely
and this broke folium. I might have needed to upgrade other stuff too but
this was the most minor way I found.
I also don't like forking. But, image overlay is broken if we don't change
the css.

ocefpaf:
> Then I tried using the beta release completely and this broke folium. 

Let me take a stab at this first.  If there is any fundamental change in leaflet that demands a big effort on our side we'll go for hosting the "patched" css here  (patch is a better than fork, right :stuck_out_tongue:).

andrewgiessel:
Cool. A branch that uses the beta is probably a good 'test' to see how much
the 1.0 will affect us. I think the leaflet blog has a post w a link to the
changelog

BibMartin:
Maybe we can try to move ImageOverlay to a plugin, and take the occasion to add something like:

    .leaflet-container img.leaflet-image-layer {
        max-width: 15000px !important;
        max-width: none !important;
        }

into the `render_css` method. If I understand well, this would be enough for waitng for Leaflet 1.0

I'll try to work on this ion the coming hours, if you want.

ocefpaf:
@BibMartin that is a far more elegant solution than my approach.  @andrewgiessel are interested into being the first one to write a plugin to folium beside the guy that creates them?

andrewgiessel:
closing this and re-opening against 0.1.6 RC branch

PR #168, opened 2015-08-05
Image overlay tweaks

andrewgiessel:
This adds mercator projection by default to image_overlay.

### DIFF
--- a/folium/utilities.py
+++ b/folium/utilities.py
@@ -345,3 +345,43 @@ def write_png(array):
         png_pack(b'IDAT', zlib.compress(raw_data, 9)),
         png_pack(b'IEND', b'')])
  
+def mercator(geodetic):
+    """This function takes an 2D array in geodetic coordinates (ie: lat x
+    lon unprojected) and converts it to web mercator.  This is needed
+    to correctly overlay an image on a leaflet map, which uses web
+    mercator by default.
+
+    This code works with arrays that match the relative proportions of
+    latitude and longitude of the earth, meaning that they have twice
+    as many points in longitude as latitude (i.e.: (180, 360) for 1
+    degree resolution).
+
+    The code for this is from:
+    http://stackoverflow.com/questions/25058880/convert-to-web-mercator-with-numpy
+
+    Parameters
+    ----------
+    geodetic: 2D numpy array
+        Latitude x Longitude 2D array
+
+    meractor: 2d numpy array
+
+    """
+    
+    def row2lat(row):
+        return 180.0/np.pi*(2.0*np.atan(np.exp(row*np.pi/180.0))-np.pi/2.0)
+    
+    geo = np.repeat(geodetic, 2, axis=0)
+    merc = np.zeros_like(geo)
+    side = geo[0].size
+    for row in range(side):
+        lat = row2lat(180 - ((row * 1.0)/side) * 360)
+        g_row = (abs(90 - lat)/180)*side
+        fraction = g_row-np.floor(g_row)
+
+
+        high_row = geo[np.floor(g_row), :] * (fraction)
+        low_row = geo[np.ceil(g_row), :] * (1-fraction)
+        merc[row, :] = high_row + low_row
+
+    return merc
--- a/folium/utilities.py
+++ b/folium/utilities.py
@@ -344,7 +344,10 @@ def write_png(array):
         png_pack(b'IHDR', struct.pack("!2I5B", width, height, 8, 6, 0, 0, 0)),
         png_pack(b'IDAT', zlib.compress(raw_data, 9)),
         png_pack(b'IEND', b'')])
- 
+
+def _row2lat(row):
+    return 180.0/np.pi*(2.0*np.arctan(np.exp(row*np.pi/180.0))-np.pi/2.0)
+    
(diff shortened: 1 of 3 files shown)


folium/utilities.py
```diff
@@ -345,3 +345,43 @@ def png_pack(png_tag, data):
...
+    ----------
+    geodetic: 2D numpy array
+        Latitude x Longitude 2D array
+
+    meractor: 2d numpy array
+
+    """
+    
+    def row2lat(row):
+        return 180.0/np.pi*(2.0*np.atan(np.exp(row*np.pi/180.0))-np.pi/2.0)
```
ocefpaf: Do you mean `np.arctan` or `np.arctan2`?

andrewgiessel: Bah, arctan. The original code used the math lib instead of bumpy.

andrewgiessel:
Ok, the arctan thing is fixed (I was traveling, sorry for the delay on such a simple change).

Why is the build failing?  The travis logs are pretty confusing to me.

BibMartin:
It seems you have an `AssertionError` in `folium_test.py` at line 559.

    assert image_rendered in self.map.template_vars['image_layers']

andrewgiessel:
ah, i probably know what it is.  thanks, for some reason the travis logs weren't rendering for me.

andrewgiessel:
green!

ocefpaf:
@andrewgiessel `folium_tests.py` needs some refactoring and your `test_image_overlay` is a nice place to start.  Can you create a `test_image_overlay.py` and remove those tests from `folium_tests.py`?

tests/folium_tests.py
```diff
@@ -510,6 +515,7 @@ def test_image_overlay(self):
         self.map.image_overlay(data, filename=image_url)
         assert image_rendered in self.map.template_vars['image_layers']
 
+        # test inline png
```
ocefpaf: I know it is a minor thing, but can you start the sentences with a Capital letter and end with a period**.** Also, remove the white space below.

folium/utilities.py
```diff
@@ -345,3 +345,42 @@ def png_pack(png_tag, data):
...
+    degree resolution).
+
+    The code for this is from:
+    http://stackoverflow.com/questions/25058880/convert-to-web-mercator-with-numpy
+
+    Parameters
+    ----------
+    geodetic: 2D numpy array
+        Latitude x Longitude 2D array
+
```
ocefpaf: Missing a `Returns` here, right?  Although, this function only takes-1 and returns-1, so we could probably remove all this.  It is covered in the first line of the docstring.

folium/utilities.py
```diff
@@ -345,3 +345,42 @@ def png_pack(png_tag, data):
...
+    meractor: 2d numpy array
+
+    """
+    
+    def row2lat(row):
```
ocefpaf: Move this outside and name it `_row2lat`.

andrewgiessel: Thanks for the comments!  Also, I think there is a small bug someplace, as
when I try to use this with actual data I get some bounds issues.  I'm
working on tracking it now now.  When I use a version of the mercator
projection code from elsewhere it works.

Do you want me to leave parameters/returns in `geodetic_to_mercator()` or
not?  I vote to have it there for consistency.  All other comments I have
no problem with at all.

andrewgiessel:
Ah, I know the issue.  `geodetic_to_mercator()` expects a lat x long array, but if you send in a colored array (lat x long x 3 or 4) then it fails.  I will generalize so that it can take either.

folium/utilities.py
ocefpaf: >  I vote to have it there for consistency. 

OK.

andrewgiessel: done

folium/utilities.py
andrewgiessel: done

andrewgiessel:
@ocefpaf I've made the changes you've suggested with the exception of the test refactoring, I think that should be a separate PR.  I can open an issue, or open one and assign it to me, if you want.  The scope can be either to refactor `image_overlay` specific tests or a more general test refactorization.

BibMartin:
Again I feel like the *nasty plugin commercial*, but it may be possible to move imageOverlay into a plugin.

btw, I imagine it would be simpler API if it is splitted into:
* An ImageOverlay plugin that gets a PNG, or PNG string, or PNG filename.
* A utility function `numpy -> png`

I'm sorry to add this amount of work ; I can do the plugin if you want. But as asked @ocefpaf in #167, maybe I shall not be the only one that writes plugins.

ocefpaf:
> I shall not be the only one that writes plugins.

Indeed.  That is the way we want to go so every contributor, myself included, must make the effort to take a stab at it.

@andrewgiessel How do you feel about tackling the plugin approach in this PR?  I am sure @BibMartin will help you all the way :stuck_out_tongue_winking_eye:  (and I will be following closely to learn!)

If not... we can merge it as is.  But we have to put a moratorium on further changes to avoid increasing the effort for the pluginification later.

andrewgiessel:
closing this and re-opening against 0.1.6 RC branch

Issue #169, opened 2015-08-05
Avoid code duplication in div marker


ocefpaf:
As @themiurgo pointed out PR #166 can be incorporate into markers.

> This is what Leaflet does with divicons, it doesn't use a separate method / code, it just leverages marker.  (See http://jsfiddle.net/erictheise/k6ckm/)

themiurgo:
@shankari since you were the author of that PR would you be interested in fixing this, so that folium API stays as clean as possible?

shankari:
I did consider making this a part of marker (a different kind of icon), but I didn't do it for two reasons.
- Leaflet doesn't actually include this as part of marker, it makes it a different kind of icon. But the folium API combines the marker and the icon into same call. Adding support for a different kind of icon in the marker API seemed like it would make it too complex - there are already 3 parameters dealing with icon (`marker-icon`, `clustered_marker` and `icon_angle`) and adding more would turn the API into something out of `matplotlib` :)
- I could have restructured this to more closely mirror the leaflet API - to have a `simple_icon` constructor which would be passed in to a `simple_marker`, but I wasn't sure how that would fit in with your architecture. You don't seem to do that anywhere else.
- It also seemed to me that the main use of the div icon (bunch of little boxes with not much decoration) would be if there are a lot of them in close proximity. So I optimized the API for that. Note that unlike `simple_marker`, it doesn't not take a single location, but instead, a list of locations and popups.

themiurgo:
> there are already 3 parameters dealing with icon (marker-icon, clustered_marker and icon_angle) and adding more would turn the API into something out of matplotlib :)

Agreed, but adding more data types/methods also has a cost: makes it harder to maintain them. :smile: In the end what you want to plot on this map is two things: a line, and a list of styled markers along that line. Also, I agree that clustered marker should be refactored somehow, again, to follow as much as possible Leaflet.

> I could have restructured this to more closely mirror the leaflet API - to have a simple_icon constructor which would be passed in to a simple_marker, but I wasn't sure how that would fit in with your architecture. You don't seem to do that anywhere else.

True, it's not done anywhere else. At the moment is under heavy restructuring, it's trying to find its soul. Since it's entirely based on Leaflet, refactoring and following as much as possible the API will make it easier to develop new features. We can't change (easily) what's been made in the past but we can think better for future features.

[...]

Issue #155, opened 2015-07-26
Handling external and internal dependencies


BibMartin:
Folium generates a web page that depends on several objects (css,js,json, data...). These objects may be :
* provided by an external web page or web service
* stored on the disk
* embeded into the HTML file

I wonder how folium shall handle that dependencies. Depending on what you need, on how you serve your web page, etc, you may be willing to have everything embeded into a single html, or to have things loaded on the fly.
It becomes tricky when trying to implement plugins that depends on libraries that are only served on HTTP (or only HTTPS).

One could:
* Have every dependency downloaded when installing Folium
* Have everything downloaded when generating the map
* Have everything linked into the html ; so that they are downloaded by the client directly
* Provide a choice between the former strategies

themiurgo:
I'm not sure I understand this. I see two potential problems:
- plugins/resources served only on HTTP/HTTPS.
- resources available only locally (like a json file);

On the other hand I don't understand what you mean by "have every dependency downloaded when installing Folium"? If Library X is not served online, how would you install it? Please provide an example.

BibMartin:
If library X is served only on HTTP and I serve my map over HTTPS, then I can :
* Download it when I install folium (for example have `urllib` downloading it into `setup.py`, or have it in the project's requirements), though it's probably not the best solution to me.
* Donwload it when generating the map.

And once I have downloaded it ; I have the choice between
* Serve it and link to it
* Embed it into the main HTML.

Antoher potential problem (I have face recently with Leaflet-terminator) is :
* plugins/ressources whose url changes.

ocefpaf:
> plugins/resources whose url changes.

Yep, that sucks... We have to face that we just cannot keep track of all URL resources.

We still have some objects that might be either stored on the disk or embedded into the HTML file.  I am one the fence here.  On one hand "embedding" would return a stand alone HTML file, while storing on disk would be "cleaner" (people keeping their HTML under version control like that).

But I digress... I think your concern are the only the plugins, right?

In that case I would avoid downloading and installing non-served dependencies.  We should keep it simple and rely on served (hopefully stable and preferably over HTTPS scheme) libraries.

themiurgo:
Ok, I now see what you mean.

I would make a few simplifying assumptions to avoid covering all scenarios:
- the plugin resources are available over HTTPS (for example, for Leaflet Terminator you can use https://rawgit.com/joergdietrich/Leaflet.Terminator/master/L.Terminator.js)
- resource URLs do not change (often).

[...]

Issue #138, opened 2015-05-22
Request: Additional file input


ColeTownsend:
It would be awesome if you could insert a CSS file when building this to style the page a bit. Or add additional elements to the webpage itself.

This is my workaround for now:

```python
# inset css file post map creation and title it.
            soup = BeautifulSoup(open("map/{}/index.html".format(year)))
            meta = soup.find('meta')
            css = soup.new_tag('link')
            css['rel'] = "stylesheet"
            css['href'] = "../map.css"
            meta.insert_after(css)

            body = soup.find('body')
            h1 = soup.new_tag('h1')
            h1.string = "TDF {} Map".format(year)
            body.insert_after(h1)
```

ocefpaf:
:+1:  for an option for a "full" CSS override in case you do not want to use the default.

:-1:  For supporting additional elements.  That would bring a complexity that is beyond the scope of folium.  And you can always edit the HTML later :wink: 

Is this something you might want to take a stab at?

themiurgo:
At some point we will probably need to define some template areas where we can inject some elements for the plugin feature #18 , for example "headers". Maybe we can make those methods public and allow people to import their own `<script>` or `<style>`.

ColeTownsend:
@ocefpaf Is it beyond the scope? It already creates a map element that is its own HTML page. It wouldn't be a horrible option and would be super useful for titles. As I'm programmatically building multiple maps, I don't want to hand write any of the files. 

One could presumably import the map file into a more robust page. I'm serving them as their own pages individually.

Ex: http://coletownsend.github.io/ben-cole-project/

ocefpaf:
@ColeTownsend I can be convinced otherwise, but right now I am biased towards keeping folium simple and solving some more pressing issues we have.

Still, if you send a PR and some nice arguments, we can keep talking and evolving this idea.  It seems that @themiurgo even has a model for this in his head already :wink: 

So lets keep this issue opened.

ocefpaf:
pinging @BibMartin here.  I would like your opinion on this issue.

BibMartin:
In the plugins machinery, we define 4 places in the template where a plugin can inject code :
* header (for injecting javascript import, or css import)
* css (for injecting just one or two css styles)
* html (for creating you own div, etc)
* javascript (for injecting code into the leaflet script)

So it is quite easy to create a plugin that injects a CSS into folium's template.

[...]

Issue #34, opened 2014-06-03
Support GeoPandas


barryrowlingson:
Its possible to use Folium to map using data and geometry from a GeoPandas object, but it requires writing the GeoPanda geometry to geojson (using the .to_json method);

something like, for gp being a geopandas obejct:

```
open("geom.json","w").write(gp.to_json())
```
and then you can map a column 'q' in the GeoPandas object:

```
map.geo_json("geom.json", data=gp, columns=["name", "q"], key_on='feature.properties.name', ...etc)
```

It should be possible to remove the steps of writing the json and the keying for a GeoPandas frame. Something like:

```
map_osm.geo_json(gp, column="q",....etc...)
```

ocefpaf:
@barryrowlingson the main problem that is the name "`geo_json`", that implies a `GeoJSON` and not a `GeoPandas` object.  We could either introduce a new method that is more "duck type friendly."  This method could take a `GeoPandas`, `GeoJSON`, `Shapely`, `GPX` and other similar objects and files.

We could then convert everything to `GeoJSON` and add the layer or we could look into the Leaflet plugins that do that for us and, in same cases, preserve the data format.

Suggestions?

BibMartin:
We can easily imagine a plugin doing that. It would look like:

    map_osm.add_plugin(GeoPandas(gp))

But maybe we shall think about having a python object for storing geospatial data into folium. Something that would have methods like `from_geojson`, `from_gpx`, ... and methods like `to_geojson`, `to_leaflet`, etc. Not reinventing the wheel ; it may inherit from `shapely` or `geopandas`. What do you think ?

ocefpaf:
I am inclined towards `from_geojson`, `from_gpx`, and `to_geojson`, `to_leaflet`... But we already have `geo_json`.  Maybe we should break the API in a future version...

wrobstory:
Ugh, I hate that I ever wrote the `geo_json` method. Terrible API design. 

Food for thought: If I Were To Rewrite This Today:
* That method would be called `data_overlay`, to follow the `image_overlay` convention.
* `data_overlay`s could be GeoJSON, TopoJSON, GeoPandas object, whatever. 
* When `overlay` is called, a `FoliumDataSource` is created. This class would be the translation layer between the various file formats and GeoJSON, which would be used to build the map. No matter what data source is backing your `FoliumDataSource`, and API to access GeoJSON from it will be the same. This is the API that `data_overlay` will interact with. 
[...]

PR #170, opened 2015-08-06
Move Map and Plugin object to a tree structure

BibMartin:
3 simple commits, and an important one (https://github.com/python-visualization/folium/commit/561f060f3b17c3039ef708f100be8596bf7ae44f).

The goal of the latter is to move progressively the Map and Plugin objects into a tree structure.

What I have in mind is a tree like :

    +- figure
        +- header
        +- css
        +- body
        +- script
        +- maps
            +- Markers
            +- Layers
            +- Plugins
            ...
        +- Others

Each *node* of the tree would have a `render` method that can have an effect on his parents.

This would have several positive effects:
* One can embed several maps into a single page, and eventually add other html stuff, as asked in #138.
* Each Leaflet object would have a python *mirror* so that one can compute interactions between objects. For example, `marker`, `popup`, `divIcon` and `awesomeMarker` need that kind of interactions (see #169). In my mind, this corresponds to the idea of a `FoliumDataSource`, as @wrobstory mentioned in #34.
* Some plugins will be easier to implement : provided that a Jinja temple has a `render` method, it can be a leaf in that tree. This may help today's plugins and @themiurgo's #147 to converge.
* Each node implements how it's children are rendered (and in which order), which provides more flexibility.
* Issue #155 would find an elegant solution : an external link would be an object that you can render several ways (http, https, embeded). (see https://github.com/python-visualization/folium/commit/5e90c1bf3f2e5e51c062c041393b95c3386710ba)

Well, discussion is open... ;-)

### DIFF
--- a/folium/plugins/marker.py
+++ b/folium/plugins/marker.py
@@ -5,9 +5,31 @@ Marker plugin
 
 """
 from jinja2 import Template
-#import json
+import json
 
 from .plugin import Plugin
+
+class Popup(Plugin):
+    def __init__(self, html, max_width=300):
+        super(Popup, self).__init__()
+        self.plugin_name = 'Popup'
+        self.html = json.dumps(html)
+        self.max_width = max_width
+        self.map = None
+
+    def render(self, **kwargs):
+        assert self.map is not None, "Cannot render a Popup that have no 'map' attribute."
+        #{{ pop_name }}.bindPopup({{ pop_txt }});
+        #{{ pop_name }}._popup.options.maxWidth = {{ width }};
+        self.map.figure.script['Popup_'+self.object_name] = Template("""
+        var Popup_{id} = L.popup({{
+            maxWidth: '{maxWidth}'
+            }}).setContent({html});""".format(
+                html=self.html,
+                maxWidth=self.max_width,
+                id=self.object_name,
+                ))
+        return 'Popup_'+self.object_name
 
 class Icon(Plugin):
     def __init__(self, color='blue', icon='info-sign', angle=0):
@@ -17,8 +39,8 @@ class Icon(Plugin):
         self.icon = icon
         self.angle = angle
         self.map = None
-        
-    def render(self):
+
+    def render(self, **kwargs):
         assert self.map is not None, "Cannot render an Icon that have no 'map' attribute."
         self.map.figure.script['Icon_'+self.object_name] = Template("""
         var Icon_{id} = L.AwesomeMarkers.icon({{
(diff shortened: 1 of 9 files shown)


BibMartin:
I forgot ; for the moment, the tree structure is:

    +- Map
        +- figure
            +- header
            +- css
            +- body
            +- script
        +- template_vars
        +- plugins
        ...

The idea is to move progressively `template_vars` into a structured tree (without changing the API), then to have `figure` embed several maps.

themiurgo:
I feel like this PR is actually 2PRs, one part is the plugin render idea, the other is the tree structure.

Some questions:
- Why is the root name called figure? Shouldn't it be something like "page" or "root"?
- Why is plugin a child of maps? Surely each plugin affects the whole page and needs to inject their own header. If I embed two maps with the same sets of headers it will be difficult to track whether the js/css has been already injected beforehand.

BibMartin:
> Why is the root name called figure? Shouldn't it be something like "page" or "root"?

I chose `figure` because I have something like matplotlib in mind : a figure in which you can put several maps/axes. I thought we could get closer to their function names. I would be okay for "page" or "root".

> Why is plugin a child of maps?

A plugin generally adds something on the map. But you're true, it can require to add somthing to the overall figure. This is why I have designed the thing so that each node knows who is his parent, and can have an effect on it. For example, if you have something like 

    Figure
        map1
            boatmarker1
            boatmarker2
        map2
            boatmarker3
            boatmarker4

Then each boatmarker will do something like

    boatmarker.map.figure.header['boatmarker_js'] = Link('http://...')

But as it is the same Figure object, they will simply overwrite the former header.
BTW, this is why we need to use `OrderedDict` instead of lists : order is important, but we also need to get objects by their key/name.

BibMartin:
Let's say that `VegaPopup` is working (though it's pure prototype). @wrobstory I would be happy if you have time to wartch https://github.com/python-visualization/folium/commit/8ce53ea009654e2f67e1220436b81650d327791c as it is going far from `folium.Map._popup_render` and I'm not really sure of the javascript generated.

BibMartin:
:green_heart: At last !

folium/plugins/marker.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def render(self, **kwargs):
+        """TODO : docstring here"""
+        assert self.map is not None, "Cannot render a Popup that have no 'map' attribute."
+        self.map.figure.header['d3'    ] = JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js")
+        self.map.figure.header['vega'  ] = JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/vega/1.4.3/vega.min.js")
+        self.map.figure.header['jquery'] = JavascriptLink("https://code.jquery.com/jquery-2.1.0.min.js")
+        self.map.figure.script['vega_parse'] = Template("""function vega_parse(spec, div) {
+            vg.parse.spec(spec, function(chart) { chart({el:div}).update(); });}""")
+
```
ocefpaf: With this approach it should be relatively easy to create mpld3 (pr even bokeh) popups.

BibMartin:
This is outdated. Replaced by #203

PR #180, opened 2015-08-14
Image overlay tweaks

andrewgiessel:
I think… this is it.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -952,16 +952,23 @@ class Map(object):
 
     @iter_obj('image_overlay')
     def image_overlay(self, data, opacity=0.25, min_lat=-90.0, max_lat=90.0,
-                      min_lon=-180.0, max_lon=180.0, image_name=None, filename=None):
-        """Simple image overlay of raster data from a numpy array.  This is a lightweight
-        way to overlay geospatial data on top of a map.  If your data is high res, consider
-        implementing a WMS server and adding a WMS layer.
+                      min_lon=-180.0, max_lon=180.0, image_name=None, filename=None,
+                      data_projection='mercator'):
+        """Simple image overlay of raster data from a numpy array.  This is a
+        lightweight way to overlay geospatial data on top of a map.
+        If your data is high res, consider implementing a WMS server
+        and adding a WMS layer.
 
-        This function works by generating a PNG file from a numpy array.  If you do not
-        specifiy a filename, it will embed the image inline.  Otherwise, it saves the file in the
-        current directory, and then adds it as an image overlay layer in leaflet.js.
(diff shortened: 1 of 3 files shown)


andrewgiessel:
ah, almost.  I have stuff in folium.py.  one sec!

andrewgiessel:
Just for reference, this is the old PR

https://github.com/python-visualization/folium/pull/168/files

The only files I changed in that huge list of changes that got merged into the PR are:  folium.py, folium_tests.py and utilities.py.  I think this matches those changes exactly.

ocefpaf:
@andrewgiessel I just merged this and I am playing with it before the release.  First thing I note is that you chose `opacity` for transparency.  How about changing it to `alpha`, to be more "matplotlib-ish"?

ocefpaf:
And the default of 0.25 might be too transparent, maybe 0.65?  Or even 1 and let the user change it?

ocefpaf:
We are almost there!  Specifying a local file or URL works, but the numpy array seems to be buggy:

http://nbviewer.ipython.org/gist/ocefpaf/7c9a0a5ec354ab9d3238

@andrewgiessel Can you take a look at it?

andrewgiessel:
@ocefpaf How about 0.5?  I set it that low because I use data overlays and I want to see the map underneath.  If it's photographs or landsat, then it's probably something different.

re: numpy array:  what is the shape of `data = plt.imread('itaguare_cropped.jpg')`?

andrewgiessel:
also, keep in mind that if a file or URL string, there is no re-projection done, and to have it match the underlying map projection it needs to be in web mercator.

for numpy arrays, by default, we assume geodetic arrays and project them.  I think the doc string could be clearer...

ocefpaf:
> @ocefpaf How about 0.5? I set it that low because I use data overlays and I want to see the map underneath. If it's photographs or landsat, then it's probably something different.

I am leaning towards `1` and letting the user adjust the opacity.  Pinging @Bibmartin and @themiurgo  for more votes here.

> I think the doc string could be clearer...

Indeed!  If you want to patch it up just make sure to work on the updated `v0.1.6` branch.

> re: numpy array: what is the shape of data = plt.imread('itaguare_cropped.jpg')?

`data.shape = (484, 686, 3)`.  The bug seems to be in the `write_png()` function.  (See http://nbviewer.ipython.org/gist/ocefpaf/227104bc764fb78951cf)

BibMartin:
I would agree that default opacity is `1` : this is the only *natural* default value, and it's the default value for most systems (Leaflet, Matplotlib, ...).
As for the option name, I have a slight preference for `alpha` because python users are used to it.

Anywa

Issue #181, opened 2015-08-14
 Regression with 0.1.5 using vincent line chart


ianozsvald:
I've just upgraded from folium 0.1.4 to 0.1.5 (https://github.com/python-visualization/folium/issues/172), I think I have a regression. I've attached two pictures, the first is a folium map with a vincent line chart (random data) attached to London, in 0.1.4 the line chart's popup box encompasses the chart, with 0.1.5 it is clipped.

In both cases folium is using vincent 0.4.4. I've switched folium versions and the bug is consistent.

I'm using Linux Mint 17.1 (Ubuntu 14.04) with Python 3.4. The bug manifests in both Firefox and Chrome.

Is there a quick fix I can make to the demo code below? Or is this a bug that needs fixing?

folium 0.1.4 with the line chart:
![vincent_linechart_using_folium014](https://cloud.githubusercontent.com/assets/273210/9277453/105a4a66-42a2-11e5-8fca-9550e1b0dabc.png)

folium 0.1.5 with exactly the same code:
![vincent_linechart_using_folium015](https://cloud.githubusercontent.com/assets/273210/9277449/0c5c495a-42a2-11e5-9a29-acafc4fdf6d2.png)

Sample self-contained code to generate this:
```
import pandas as pd
import numpy as np
import folium
import vincent

# using folium 1.4
# python -m http.server  # must serve up as json, can't use file-open

sales = folium.Map(location=[51.5, 0], zoom_start=7, tiles='stamentoner')


def make_hex_color(mn, mx, value):
    """Make a black->red hex colour"""
    assert mn <= value
    assert value <= mx
    percent = (1/(mx-mn)) * (value-mn)
    red_in_hex_range = int(255 * percent)
    # make a HTML color code e.g. #ff0000
    color = "#{:02x}0000".format(red_in_hex_range)
    return color


# make some random sales data
prices = np.random.uniform(low=100000, high=10000000, size=12).astype(np.int_)
df = pd.DataFrame({'price': prices})

# make a vincent Line from the dataframe
bar = vincent.Line(df['price'], width=200, height=150)
bar.axis_titles(x='Month', y='')
bar_json = bar.to_json()
# export the data to a file as per vincent's requirement
with open('bar.json', 'w') as f:
    f.write(bar_json)

# make a marker that'll host the line chart
sales.circle_marker(location=[51.521282, -0.115279], radius=5000,
                    line_color='#ff0000',
                    fill_color=make_hex_color(0, 1000000, 1000000),
[...]

PR #182, opened 2015-08-14
Fix some typos and pep8

ocefpaf:
I was fixing #181 and started to pep8 and fix typos... So I decided to send this as a PR first and then the actual fix.  @BibMartin or @themiurgo can either of you review and merge this one?

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -359,7 +359,8 @@ class Map(object):
         self.template_vars.setdefault(name, []).append(append)
 
     @iter_obj('div_mark')
-    def div_markers(self, locations=None, popups=None, marker_size=10, popup_width=300):
+    def div_markers(self, locations=None, popups=None,
+                    marker_size=10, popup_width=300):
         """Create a simple div marker on the map, with optional
         popup text or Vincent visualization. Useful for marking points along a
         line.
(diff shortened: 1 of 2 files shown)


ocefpaf:
Re-based and good to go.

folium/folium.py
```diff
@@ -404,9 +413,12 @@ def div_markers(self, locations=None, popups=None, marker_size=10, popup_width=3
                                        'icon': "{'icon':"+icon_name+"}"
                                        })
 
-            popup_out = self._popup_render(popup=popup, mk_name='div_marker_{0}_'.format(call_cnt),
```
BibMartin: Why do you fix this line like this instead of something like

    mk_out = 'div_marker_{0}_'.format(call_cnt)
    popup_out = self._popup_render(popup=popup, mk_name=mk_out)

?

I don't know pep8 very well, but it seems to me that it would be more understandable.

folium/utilities.py
```diff
@@ -1,18 +1,21 @@
...
 Utility module for Folium helper functions.
 
-'''
+"""
 
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import division
+
+import time
```
BibMartin: Why do we need to import time ?

BibMartin:
2 comments in the code ; as for the rest, it seems good to me.

folium/utilities.py
ocefpaf: It is used in [this line](https://github.com/ocefpaf/folium/blob/pep8/folium/utilities.py#L260).  I am unsure of the consequences of its absence until today, but I am slowly revising `utilities.py`.

folium/folium.py
ocefpaf: Yep.  Much better.  I am just going to call it `mk_name` instead of `mk_out` to be more consist.

BibMartin:
Ok, let's merge :tada:

ocefpaf:
Thanks.  That kind of cleanup was in my TODO list for folium since day 1 (well.. It is python so day 0)!

PR #183, opened 2015-08-14
Fix vincent max popup width.

ocefpaf:
I wonder if we really need this.  Maybe using the popup `width` is enough even if the user specifies a width greater than the map itself.  It will look silly, but maybe that is what the user wants.

Closes #183


ocefpaf:
@BibMartin or @themiurgo can you please review and merge?  Once this gets on master I will port it to `0.1.6`.

(This fixes #183)

PR #187, opened 2015-08-16
write_png fix!

andrewgiessel:


andrewgiessel:
@ocefpaf your example notebook works locally now!  This slipped through because the code oddly worked before in the case of using square arrays (width and height are equal).  We also needed to flip the array up down to make it make sense.

andrewgiessel:
I can tag on some other doc string edits, too, if this works for you

ocefpaf:
Merged!  Thankss @andrewgiessel!

ocefpaf:
> I can tag on some other doc string edits, too, if this works for you

It is working with that example and I am creating more examples here.  (Like plots generated from Basemap and cartopy as overlays.)

Lets fix the docstrings in another PR.  I also want to discuss:

- The use of `None` as the default projection.  You can see that something is weird in the image I used;
- The default `opacity = 1`;
- Full example where the projections makes a difference;
- Less wordy docstrings.

After that we should release `v0.1.6`.  (And then start porting this as a plugin :stuck_out_tongue_winking_eye: )

andrewgiessel:
Cool, I'll open an sane default / docstring PR.

How did your other examples go?

ocefpaf:
> How did your other examples go?

Not ready yet...  I lost my weekend window.  Now that will have to until... The day job is calling!

Issue #173, opened 2015-08-12
RE-factor the tests


ocefpaf:
As folium changes its internal architecture we need more robust tests than the monolithic `test_folium.py`

wrobstory:
I would also strongly recommend a move from nose to py.test

ocefpaf:
> move from nose to py.test
:+1:

PR #189, opened 2015-08-17
Using py.test instead of nose

ocefpaf:
First step to close https://github.com/python-visualization/folium/issues/173

@BibMartin or @themiurgo  please... :smile_cat:


BibMartin:
Wow ! This looks good (and big amount of work).
Globally I'm okay. But I'm much too ignorant about travis to be confident for merge.

ocefpaf:
So lets break it into pieces to make it easier to discuss.

- We had 30 tests and the are all [passing](https://travis-ci.org/python-visualization/folium/jobs/75993480#L359) in the new framework :tada: 
- I removed the `Jinja2` version constraint `==2.7.2` so we can test with the latest version and avoid surprises.
- We no longer support `distutils`.  `setuptools` is a requirement now.
- Inside `travis.yml` and `setup.py` I am reading the requirements from `requirements.txt`.  In the future I want to have `requirements-extras.txt` (with: pandas, vincent, geopandas, etc), and `requirements-dev.txt` (the testing stuff: mock and pytest).  That way we will have only ~~one~~ 3 places to look for the dependencies instead of writing them on `.travis.yml`, `setup.py`, and the req files. 
- I cleanup `setup.py` a little bit and added the test class (so we can run `python setup.py test`)
- `nose` is gone!  I tweaked the tests to have the same prefix (`test_something.py`) and changed the classes names so `pytest` can discover them.  The data path is now absolute, so we can run the tests from any directory.

In the future I want to break the current tests into their own files (`teat_line.py`, `test_marker.py`, etc).  I know it sounds like an overkill, but I want to beef up the tests a lot to avoid regressions,  and with the monolithic file we have it will be a nightmare.

BibMartin:
@ocefpaf Thanks for explainations.

I was a bit frightened by my ignorance on travis ; but the tests pass and we don't ask more in this PR.

Issue #176, opened 2015-08-13
Cannot inline a map that binds features to data


papajohn:
Below, I've written a minimal program that ideally would run inside a Jupyter Notebook and generate the colored map from http://bl.ocks.org/wrobstory/5609830 . Unfortunately, the map does not render because `_build_map` creates HTML that depends on a `data.json` file (generated in `create_map`, which I never call). 

Instead of requiring File IO in order to create a map that binds features to data (by hard-coding `data.json` as the data path), would you consider adding the option of passing data to `_build_map` and inlining it into the HTML? That way, rendering the result in a Jupyter notebook would be straightforward.

The data files in this example are taken from http://bl.ocks.org/wrobstory/5609830
Example:

```python
import json
import pandas
import folium
from IPython.core.display import HTML

data = pandas.read_csv('tests/us-unemployment.csv')
map = folium.Map(location=[48, -102], zoom_start=3)
geo_path = 'tests/us-states.json'
geo = json.load(open(geo_path, 'r'))
map.geo_json(
    geo_path=json.dumps(geo), 
    data=data, 
    columns=['State', 'Unemployment'], 
    fill_color='YlGn', 
    key_on='feature.id',
)

def inline(m, width, height):
    """Returns an embedded iframe of a folium.map."""
    m._build_map()
    src = m.HTML.replace('"', '&quot;')
    style = "width: {}px; height: {}px".format(width, height)
    return '<iframe srcdoc="{}" style="{}"; border: none"></iframe>'.format(src, style)

HTML(inline(map, 900, 500))
```

Result:

![screen shot 2015-08-13 at 11 04 33 am](https://cloud.githubusercontent.com/assets/320145/9257571/2675d204-41ab-11e5-8722-b8ead9cbeecc.png)

papajohn:
By the way, this hack makes the map display correctly in the notebook by inlining the JSON data.

```
html = inline(map, 900, 500)
for name, data in map.json_data.items():
    html = html.replace('d3.json, ' + repr(name), 'function(callback){callback(null, JSON.parse('+ repr(json.dumps(data).replace('"', '&quot;')) + '))}')
HTML(html)
```

ocefpaf:
We are working on a better integration with IPython/Jupyter notebooks.  Good to know that the JSON in-lining works.  I am pretty sure a lot of people will use that until we actually fix this. Thanks @papajohn!

ocefpaf:
[...]

PR #190, opened 2015-08-17
Simpler HTML repr

ocefpaf:
I removed all the IPython notebook stuff and simplified the `_repr_html_` to inline the JSONs.  This should allow for just leaving the `map`object as the last line in the notebook and get a rich HTML display as expected.  I made a bunch of examples here:

http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/tree/master/

I appreciate as much input as I can get here before merging.  I am targeting the `v0.1.6` branch and, if this works out there in the wild, we can go for @BibMartin far more elegant [solution](https://github.com/python-visualization/folium/issues/176#issuecomment-131693563).

PS 1: Note that the `geo_path` option does not work.  I wonder if we should treat all GeoJSONs as `geo_str`. If someone finds the time please take a look at the example above and try to find problems.

PS 2: I ran all those notebooks in the latest `Jupyter` and `IPython 2.3.1` with success.


wrobstory:
uhhhhhhhhhhh these notebooks are amazing. You all are doing such great work.

BibMartin:
This looks great ! I would be okay to merge.
I'll wait 12 more hours to enable others' input

ocefpaf:
> I'll wait 12 more hours to enable others' input

@BibMartin your are much wiser than me :wink:

BibMartin:
I just try to learn from our former mistake :stuck_out_tongue_winking_eye:

BibMartin:
:clock1: It's time to merge. Cheers !

ocefpaf:
Thanks!  `v0.1.6` is almost ready...

Conengmo:
Hi @sertalpbilal, two things, could you open a new issue for these kind of bug reports? And can you try upgrading to the latest version of folium? Should be 0.5 or even the Git master.

sertalpbilal:
@Conengmo Sure, will do!

PR #191, opened 2015-08-18
docstring and defaults

ocefpaf:
Addressing some of the points we raised in https://github.com/python-visualization/folium/pull/187.

@andrewgiessel since this is your baby can you please take a look?

Here is a working example:
http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/blob/master/test_image_overlay_gulf_stream.ipynb

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -951,44 +951,37 @@ class Map(object):
         self.template_vars.setdefault('gjson_layers', []).append(layer)
 
     @iter_obj('image_overlay')
-    def image_overlay(self, data, opacity=0.75, min_lat=-90.0, max_lat=90.0,
+    def image_overlay(self, data, image_opacity=1, min_lat=-90.0, max_lat=90.0,
                       min_lon=-180.0, max_lon=180.0, image_name=None,
-                      filename=None, data_projection='mercator'):
-        """Simple image overlay of raster data from a numpy array.  This is a
-        lightweight way to overlay geospatial data on top of a map.
-        If your data is high res, consider implementing a WMS server
-        and adding a WMS layer.
+                      filename=None):
+        """
+        Simple image overlay of raster `data` from an image or a NumPy array.
+        This is a lightweight way to overlay geospatial data on top of a map.
+        The default lon, lat corners are the whole globe.  Adjust it to your
+        image.
 
-        This function works by generating a PNG file from a numpy
-        array.  If you do not specify a filename, it will embed the
-        image inline.  Otherwise, it saves the file in the current
-        directory, and then adds it as an image overlay layer in
-        leaflet.js.  By default, the image is placed and stretched
-        using bounds that cover the entire globe.  By default, we
-        assume that your data is in geodetic projection and thus
-        project it to web mercator for display purposes.  If you are
-        overlaying a non-georeferenced image, set data_projection to
-        None.
+        If `data` is is a NumPy array a PNG file will be created for you.
+        If you do not specify a filename the image will be embedded inline.
+
+        NOTE: The image must be projected as Web Mercator
+        (https://en.wikipedia.org/wiki/Web_Mercator).  See the examples below
+        for re-projection.
 
         Parameters
         ----------
-        data: numpy array OR url string, required.
-            if numpy array, must be a image format,
-            i.e., NxM (mono), NxMx3 (RGB), or NxMx4 (RGBA)
-            if url, must be a valid url to a image (local or external)
-        opacity: float, default 0.75
-            Image layer opacity in range 0 (transparent) to 1 (opaque)
-        min_lat: float, default -90.0
-        max_lat: float, default  90.0
-        min_lon: float, default -180.0
-        max_lon: float, default  180.0
+        data: NumPy array OR image.
+            The NumPy array can NxM (mono), NxMx3 (RGB), or NxMx4 (RGBA)
+        image_opacity: float, default 1
+            Image layer image_opacity in range 0 (transparent) to 1 (opaque)
+        min_lat: float (default: -90)
+        max_lat: float (default:  90)
+        min_lon: float (default: -180)
+        max_lon: float (default"  180)
         image_name: string, default None
             The name of the layer object in leaflet.js
         filename: string or None, default None
             Optional file name of output.png for image overlay.
             If None, we use a inline PNG.
-        data_projection: string or None, default 'mercator'
-            Used to specify projection of image.  If None, do no projection
 
         Output
         ------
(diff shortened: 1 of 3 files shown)


andrewgiessel:
This looks great!  Thanks for tackling it.

A comment:  While I've never really been satisfied with `data_projection` as an argument name, I do like the explicitness of using 'mercator' and None.  That said, I don't think we should incorporate arbitrary projections (like using proj4 or whatever, blech)  If we use True/False, then maybe 'project_to_mercator' might be a clearer argument name.  There is just a lot of ambiguity.

Also, projection only works if you use it with an array, not with a local or external file.

ocefpaf:
I am not happy with `data_projection` either.  To be honest I am not happy with that option
and the current implementation.  The image distortion seen [here](http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/blob/master/test_image_overlay.ipynb) worries me...

Maybe should remove that for now and try to implement that using [rasterio](https://github.com/mapbox/rasterio) or Cartopy.

PS: Any idea about what is going on with travis?

andrewgiessel:
Ah, I think the issue is that the algorithm is for global data, not local.
So, for my global data overlays it works well.  For smaller single images,
it won't.  So, generally speaking, I think we 2 have different use cases
(which are both very important):  user generated data (my primary use), and
landsat data (yours?).  The former will tend to be global data, and this
algo will work.  The later doesn't work well.

rasterio looks pretty cool.  I don't know enough about cartopy to know how
to do this, but I should learn about both- do you want me to do that and
then look into a more general solution?

ocefpaf:
I am re-thinking this issue and  I believe it is better to avoid any re-projection inside folium.  Lets just say it takes the image in `Web Mercator` and leave the transformation to the user.  My notebook shows one example of re-projecting Geodetic data to PlateCarree (which seems to be OK in the map.)

We can easily create example with your `geodetic_to_mercator()` and `rasterio` too.  But I am leaning towards not hard-wiring this into folium.

andrewgiessel:
Ya, projection is obviously a difficult problem and generally outside the
scope of folium, I agree.

On the other hand, the utility of not having to use cartopy, make a figure,
and save a temporary file for simple overlays of numpy arrays is pretty
huge for me (and others, I believe).  The likelihood is high that anything
generated in numpy will be geodetic, so some projection needs to be done.

If `geodetic_to_mercator` is left in folium.utilities (and perhaps
referenced in the docstrings or docs?) I think this is an OK compromise.  A
globally dataset is a common use case for generated data, and we account
for that.  Anything else can be handled by the user.

ag

ocefpaf:
> If `geodetic_to_mercator` is left in folium.utilities (and perhaps
> referenced in the docstrings or docs?) I think this is an OK compromise.

That makes sense. As soon as get back in the office I will change this PR
to include the transformation back into utilities.  We just need an example
with globaly generated data for the docstring and I think I got one.

ocefpaf:
We can try to use this in the global exemple:

http://scitools.org.uk/cartopy/docs/latest/examples/waves.html
Em 18/08/2015 11:42, "Filipe Pires Alvarenga Fernandes" <ocefpaf@gmail.com>
escreveu:

> > If `geodetic_to_mercator` is left in folium.utilities (and perhaps
> > referenced in the docstrings or docs?) I think this is an OK compromise.
>
> That makes sense. As soon as get back in the office I will change this PR
> to include the transformation back into utilities.  We just need an example
> with globaly generated data for the docstring and I think I got one.
>

andrewgiessel:
Ya, good find.  That example works well:

image

image


The latter is correct.

andrewgiessel:
You can wrap `data` in a MPL colormap for fancy colors.

ocefpaf:
OK.  I am done with this PR.  [Here](http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/blob/master/test_image_overlay_geodetic.ipynb) is an example that needs the projection step.  The same example is illustrated in the docstrings.

Summary:
- Renamed the `opacity` option to `image_opacity` to be consistent with the rest of the API;
- The `image_opacity` is no transparency (`1`);
- No automatic projection is performed.  The method assumes the image/data is in Web Mercator.
- There is an example and a utility function to project the data as Web Mercator (`folium.utilities.geodetic_to_mercator`)

@andrewgiessel can you take another look and give the :+1:  or :-1: 

After that @BibMartin or @themiurgo can take another look and merge.

folium/folium.py
```diff
@@ -951,44 +951,37 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
...
+        """
+        Simple image overlay of raster `data` from an image or a NumPy array.
+        This is a lightweight way to overlay geospatial data on top of a map.
+        The default lon, lat corners are the whole globe.  Adjust it to your
+        image.
+
+        If `data` is is a NumPy array a PNG file will be created for you.
+        If you do not specify a filename the image will be embedded inline.
+
+        NOTE: The image must be project as Web Mercator
```
andrewgiessel: s/project/projected

folium/folium.py
```diff
@@ -951,44 +951,37 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
...
-            if numpy array, must be a image format,
-            i.e., NxM (mono), NxMx3 (RGB), or NxMx4 (RGBA)
-            if url, must be a valid url to a image (local or external)
-        opacity: float, default 0.75
-            Image layer opacity in range 0 (transparent) to 1 (opaque)
-        min_lat: float, default -90.0
-        max_lat: float, default  90.0
-        min_lon: float, default -180.0
-        max_lon: float, default  180.0
+        data: NumPy array OR image path.
```
andrewgiessel: or external URL

andrewgiessel:
Some minor comments in the docstring, but :+1: from me!  Great work and thanks for polishing this up so that it's more consistent with the whole project.

folium/folium.py
ocefpaf: Virtually all the external data file (GeoJSONs, images) can be local or a URL.  To keep it consistent with the other docstrings I will leave just `image` ~~path~~.

folium/folium.py
ocefpaf: Thanks!

ocefpaf:
One thing we need to come back when we pluginify this is the inline vs path reference.  I think this should be a global option (standalone HTML vs separate files) and not inside the method.

ocefpaf:
@BibMartin we have @andrewgiessel :+1: and green `Travis`.  Can you take a quick look and merge?

BibMartin:
It seems good. Congrats and thanks to both of you!

PR #188, opened 2015-08-16
Add negative values to legend

ocefpaf:
**This PR is not ready to merge!!**  (This Fixes #184)

If the user provides the tick-marks it will work, but the auto-scale is not.  Here is an example using specifying the legend tickmarks (`threshold_scale`).

```python
import json
import numpy as np
import pandas as pd

state_unemployment = 'data/US_Unemployment_Oct2012.csv'
df = pd.read_csv(state_unemployment)

with open('data/us-states.json') as f:
    states = json.load(f)

geometry, geo_id = [], []
for state in states['features']:
    geometry.append(state['geometry']['coordinates'])
    geo_id.append(state['id'])

states = pd.DataFrame(np.c_[geo_id, geometry], columns=['state', 'geometry'])

threshold_scale = np.linspace(df['Unemployment'].min(),
                              df['Unemployment'].max(),
                              6, dtype=int)
threshold_scale = threshold_scale.tolist()

mapa = folium.Map(location=[48, -102], zoom_start=3)
mapa.geo_json(geo_path='data/us-states.json', data=df,
              columns=['State', 'Unemployment'],
              key_on='feature.id',
              fill_color='YlGn', fill_opacity=0.7, line_opacity=0.2,
              legend_name='Unemployment Rate (%)',
              threshold_scale=threshold_scale)


mapa.create_map("map_manual.html")
```

PR #193, opened 2015-08-19
introduce free scale for geo_json

koldunovn:
This PR introduce `freescale` option to the `geo_json` function. It's allow to plot values that do not have 0 minimum and are, for example, float. It do not change default functionality (AFAIU from @ocefpaf a lot of people rely on it for creating choropleth maps), but gives opportunity to people like me to use real values for the plots :)

small map

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -986,7 +986,6 @@ class Map(object):
             if threshold_scale and len(threshold_scale) > 6:
                 raise ValueError
             domain = threshold_scale or utilities.split_six(series=series, freescale=freescale)
-            print(domain)
             if len(domain) > 253:
                 raise ValueError('The threshold scale must be length <= 253')
             if not utilities.color_brewer(fill_color):
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -828,7 +828,7 @@ class Map(object):
                  data=None, columns=None, key_on=None, threshold_scale=None,
                  fill_color='blue', fill_opacity=0.6, line_color='black',
                  line_weight=1, line_opacity=1, legend_name=None,
-                 topojson=None, reset=False):
+                 topojson=None, reset=False, freescale=False):
         """Apply a GeoJSON overlay to the map.
 
         Plot a GeoJSON overlay on the base map. There is no requirement
@@ -985,7 +985,8 @@ class Map(object):
             series = data[columns[1]]
             if threshold_scale and len(threshold_scale) > 6:
                 raise ValueError
-            domain = threshold_scale or utilities.split_six(series=series)
+            domain = threshold_scale or utilities.split_six(series=series, freescale=freescale)
+            print(domain)
             if len(domain) > 253:
                 raise ValueError('The threshold scale must be length <= 253')
             if not utilities.color_brewer(fill_color):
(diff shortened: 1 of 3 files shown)


ocefpaf:
Awesome @koldunovn!  It is nice to see you here.

One thing that is bothering me is the confusing API.  We have right now:

- `threshold_scale=None`
- the old "auto-starts-at-0-quantile-based"
- the new `free-scale=True/False`.

I have an uncommitted amend to #188 to remove the auto stuff and recommend people, that uses this for choropleths, to always use threshold_scale='auto' (old behavior) or `threshold_scale=my_scale`.  So `free-scale=True/False` can go away and actually became the default behavior.

Calling @BibMartin, @themiurgo, and @wrobstory here since this is a significant change.

PS: Maybe we can close #188 and do that here in your PR.

koldunovn:
I was initially inspired by your #188 , but then decide to dig a bit deeper :) 
Changing default behaviour will break people's code, so they might not appreciate it :) The free-scale flag allows to have a separate playground for functionality that is not needed for people who make choropleths, like having split based not on quantiles, maybe more categories or even continuous color scales...

ocefpaf:
We are breaking stuff for `v0.2.0` (currently brewing on `master`).  I would take that PR as is in the `v0.1.6` branch.

wrobstory:
A few thoughts: 

* This method is so overloaded right now. It really feels like the legend stuff should actually be a separate call on the Map object. 

* For any future legend method, I can foresee a `quants` argument where you pass in your own quantiles (since they are being redefined when you use `freescale` here, which is kind of hidden behavior. Not like the original quantiles weren't). 

* I can't remember why I started every legend at 0, but it feels like kind of a bad decision. It's possible that I did it because it *does* make a lot of legends look nicer. I think perhaps we should default to `lin_min`.

BibMartin:
Beware the nasty *plugin guy* is there.

I would love to have this implemented into plugins so that we can keep it on `master`.
I imagine one plugin for choropleth, and another one for the legend.

I'm sorry: I have few time this week. But I can help on this from next week, if needed.

koldunovn:
@ocefpaf If you decide to do so, let me know and I will edit doc strings.

@wrobstory It would be nice to have more control and options both for the legend and for splitting the values in to categories, and being able to use more than 6 categories. I am not familiar at all with d3, but this code seems to imply that this is possible:
```python
          if len(domain) > 253:
                raise ValueError('The threshold scale must be length <= 253')
```
Having separate legeng might make it easier to create legend also for `image_overlay`, that @ocefpaf is working on now AFAIU.  

The thing with `quants` is indeed strange, and if for the time being you decide to go for this PR, I can add them as an argument (making `geo_json` even more busy).

In current situation:
```python
'lin_max': int(domain[-1]*1.1)
```
 defaulting to `lin_min` for non zero minimum will put the minimum (if it's not negative) out of scale due to  `1.1` multiplication. Not sure why it's there...

ocefpaf:
> Beware the nasty plugin guy is there.

Where is the pointer to the docs?

Oh wait! Here it is:

http://nbviewer.ipython.org/github/python-visualization/folium/blob/master/docs/developer/plugins_introduction.ipynb

@koldunovn for now lets take the slow road.  Can you point this to `v0.1.6`.  Later we should see what it takes to implement plugins for:

- `raster_overlay`
- `vector_overlay`
- `add_legend` with free scale as default, a `quants=None/'auto'` keyword where `'auto'` is the current behavior, and no `threshold_scale`.

(or maybe `overlay_{raster/vector}` so both can be found easily by tab completion?)

koldunovn:
@ocefpaf Let me know if I can do anything else...

koldunovn:
Turns out that my dream of having more colors for the colorscale were only two lines away. Now you can make something like this (this is India BTW) :
Big map

ocefpaf:
This looks good.  However, this belongs in the `v0.1.x` series.  Folium `v0.2.x` will break some stuff and we want to implement everything as "plugins."

I am closing.  in favor of #194.

PR #196, opened 2015-08-20
introduce free scale for geo_json

koldunovn:
Same as #193 , but for v0.1.6. Hope I did not screw this up this time :)

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -737,7 +737,7 @@ class Map(object):
                  data=None, columns=None, key_on=None, threshold_scale=None,
                  fill_color='blue', fill_opacity=0.6, line_color='black',
                  line_weight=1, line_opacity=1, legend_name=None,
-                 topojson=None, reset=False):
+                 topojson=None, reset=False, freescale=False):
         """Apply a GeoJSON overlay to the map.
 
         Plot a GeoJSON overlay on the base map. There is no requirement
@@ -804,6 +804,9 @@ class Map(object):
             keyword argument will enable conversion to GeoJSON.
         reset: boolean, default False
             Remove all current geoJSON layers, start with new layer
+        freescale: if True use free format for the scale, where min and max values
+            are taken from the data. It also allow to plot allow to plot values < 0 
+            and float legend labels.
 
         Output
         ------
@@ -892,8 +895,9 @@ class Map(object):
 
             # D3 Color scale.
             series = data[columns[1]]
-            if threshold_scale and len(threshold_scale) > 6:
-                raise ValueError
+            if freescale == False:
+                if threshold_scale and len(threshold_scale) > 6:
+                    raise ValueError
             domain = threshold_scale or utilities.split_six(series=series)
             if len(domain) > 253:
                 raise ValueError('The threshold scale must be length <= 253')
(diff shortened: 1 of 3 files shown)


folium/utilities.py
```diff
@@ -305,7 +305,11 @@ def base(x):
     quants = [0, 50, 75, 85, 90]
```
ocefpaf: @koldunovn Can you get this line down (in between 308-309), so we can easily see that the two options are doing.

folium/folium.py
```diff
@@ -913,9 +917,21 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
...
+            if freescale==False:
+                legend = leg_templ.render({'lin_min': 0,
+                           'lin_max': int(domain[-1]*1.1),
+                           'tick_labels': tick_labels,
+                           'caption': name})
+
+
+            else:
+                legend = leg_templ.render({'lin_min':  domain[0],
+                                           'lin_max':  domain[-1],
```
ocefpaf: @koldunovn you are not padding 0.1 here.  How does that look?

folium/folium.py
```diff
@@ -892,8 +895,9 @@ def json_style(style_cnt, line_color, line_weight, line_opacity,
 
             # D3 Color scale.
             series = data[columns[1]]
-            if threshold_scale and len(threshold_scale) > 6:
-                raise ValueError
+            if freescale == False:
```
ocefpaf: Change this to `if not freesacale` (pep8)

folium/folium.py
koldunovn: It look fine - examples you can see in the original PR. As I mention there padding will put positive non zero lin_min out of scale.

folium/folium.py
```diff
@@ -804,6 +804,9 @@ def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
             keyword argument will enable conversion to GeoJSON.
         reset: boolean, default False
             Remove all current geoJSON layers, start with new layer
+        freescale: if True use free format for the scale, where min and max values
+            are taken from the data. It also allow to plot values that do not have 0 
+            minimum and are, for example, float.
```
ocefpaf: How about: `... allow to plot values < 0 and float legend labels.`

folium/utilities.py
koldunovn: I am not sure if this is hat you want...

folium/folium.py
ocefpaf: OK.  `'lin_min': int(domain[0]*1.1)` was probably a copy-and-paste mistake from my original PR).

The idea was to extend the legend a little bit.  We could be smart about it and use a sign function, expand, and re-added the sign.   But the legends look good without this, so lets leave it be.

folium/utilities.py
ocefpaf: Nope.  Just get the line `quants = [0, 50, 75, 85, 90]` on top of `return [base(np.percentile(arr, x)) for x in quants]`.

koldunovn: Oh, yes, make sence.

folium/folium.py
koldunovn: Without padding it looks more "normal" to me, but it's a mater of taste.

ocefpaf:
@koldunovn can you squash the commits?

@BibMartin and @themiurgo I am OK with this PR as it fixes the zero based legend and expand the legend use for other cases without breaking people codes. I think is a nice addition until we have `v0.2.0`

I will be grateful if any of you two can take a second look at this PR. (I am planing to merge it tomorrow morning.)

PR #199, opened 2015-08-22
Fix Stamen Toner URL to use .png as stated at maps.stamen.com. Update…

tahmed42:
… different attribution required specifically for Toner as also stated there.

### DIFF
--- a/folium/templates/tiles/stamentoner/attr.txt
+++ b/folium/templates/tiles/stamentoner/attr.txt
@@ -1,1 +1,1 @@
-Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>.
+Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://www.openstreetmap.org/copyright">ODbL</a>.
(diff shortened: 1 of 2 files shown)


ocefpaf:
At Stamen Toner I see exactly what we have now:

tiles

However, the OpenStreetMaps [FAQ](http://wiki.openstreetmap.org/wiki/Legal_FAQ) states:

> Our requested attribution is "© OpenStreetMap contributors".
You must also make it clear that the data is available under the Open Database Licence. This can be achieved by providing a "License" or "Terms" link which links to www.openstreetmap.org/copyright or www.opendatacommons.org/licenses/odbl.

So I guess we have to change all maps `attr.txt` that makes reference to OpenStreetMaps!

@tahmed42 Are you willing to do that in this PR?

PS: Are there any advantages of using PNG instead of JPEG?  I find that JPEGs tiles are usually easier on the bandwidth than PNGs.

ocefpaf:
Thanks @tahmed42!

tahmed42:
No problem!

Weird, at maps.stamen.com, they say:
For Toner: Map tiles by Stamen Design, under CC BY 3.0. Data by OpenStreetMap, under ODbL.
For everything else: Map tiles by Stamen Design, under CC BY 3.0. Data by OpenStreetMap, under CC BY SA.

But maybe *they're* out of date on that and OSM requires ODbl for everything now.
Also re: png, the only reason I made the change to that is because at maps.stamen.com towards the bottom in the "Elsewhere" section, it stated:
Many applications and libraries understand the notion of map URL templates. These are ours:
http://tile.stamen.com/toner/{z}/{x}/{y}.png
http://tile.stamen.com/terrain/{z}/{x}/{y}.jpg
http://tile.stamen.com/watercolor/{z}/{x}/{y}.jpg

So it went out of its way to use png for toner and I read that as jpg not being available for it (maybe it was available in the past?). Moreover though, I noticed toner wasn't loading right in folium but with the change to png, it worked. Tested that out using the custom tileset abiilty in folium, which came in super handy.

ocefpaf:
> But maybe they're out of date on that and OSM requires ODbl for everything now.

Yep.  I opened an issue for it #202.

tahmed42:
Ah whoops, didn't catch your comment edit before replying about pngs, haha.

ocefpaf:
No problem.  Thanks again for the PR!

PR #200, opened 2015-08-24
Fix WMS layer control

ocefpaf:
Commit #152 introduced a bug in the `add_layers_to_map` (The top control to select the layers).

In this PR:
- I removed the `image_overlay` from the `added_layers`.  I could not make it work due to a bug I was not able to find.
- I reverted `add_layers_to_map` to its original code and things are working now.

TODO: Re-add `image_overlay` to the `added_layers` without breaking `add_layers_to_map` for other layers.

@andrewgiessel Let me know your thoughts on this since it was your original PR.

@BibMartin Can you review and merge?  One it is done I will port it to master.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -261,19 +261,17 @@ class Map(object):
     def add_layers_to_map(self):
         """
         Required function to actually add the layers to the HTML packet.
+
         """
-        layers_temp = self.env.get_template('add_layers.js')
-
-        data_string = ''
-        for i, layer in enumerate(self.added_layers):
+        data_layers = ''
+        for k, layer in enumerate(self.added_layers):
             name = list(layer.keys())[0]
-            if i < len(self.added_layers)-1:
-                term_string = ",\n"
+            data_layers += '\"{}\": {}'.format(name, name)
+            if k < len(self.added_layers)-1:
+                data_layers += ",\n"
             else:
-                term_string += "\n"
-            data_string += '\"{}\": {}'.format(name, name, term_string)
+                data_layers += "\n"
 
-        data_layers = layers_temp.render({'layers': data_string})
         self.template_vars.setdefault('data_layers', []).append((data_layers))
 
     @iter_obj('simple')
@@ -1042,7 +1040,9 @@ class Map(object):
                                         'image_opacity': image_opacity})
 
             self.template_vars['image_layers'].append(image)
-            self.added_layers.append(image_name)
+            # FIXME: I am excluding the image_overlay from the layer
+            # control for now.  There is a bug and I cannot find it.
+            # self.added_layers.append({image_name: image_url})
 
     def _build_map(self, html_templ=None, templ_type='string'):
         self._auto_bounds()


andrewgiessel:
I'll take a look later this week

ag

andrewgiessel:
Thanks for the ping @ocefpaf.  Just getting back online after some travels.

I remember asking about this during while I was building my first PR (https://github.com/python-visualization/folium/pull/152#issuecomment-122463625) but I think a review fell through the cracks and our tests didn't catch it. My apologizes for introducing a bug- but maybe it's just exposing something that needs to be fixed.  

**Code Reference**

[original code](https://github.com/andrewgiessel/folium/blob/e2c03109f8160c0eb59f4b1a2e165631c0fd3819/folium/folium.py#L274-L295)
[my refactor (in master)](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L276-L293)
[this pr](https://github.com/ocefpaf/folium/blob/fix_wms_layer/folium/folium.py#L260-L275)

The original (and the code my refactored version mimicked) builds a string and uses [this template](https://github.com/python-visualization/folium/blob/master/folium/templates/add_layers.js) to render it, which is simply `var layer_list = { {{ layers }} };`

You can see that the original code didn't actually use the rendered string and instead used the constructed string.  I refactored the string construction for clarity (which I see you kept) and assumed that we needed to use the result from the template.  IMO this is the key issue- do we need the template?  Do some layers assume a rendering and some don't?  It seems as though this could be changed to do all the string stuff in the template, instead of this two stage rendering.

I will look more carefully at what the different layers expect - I probably am doing something non-standard in `image_layers`, despite my intentions.

@ocefpaf 
@BibMartin

ocefpaf:
> I remember asking about this during while I was building my first PR (#152 (comment)) but I think a review fell through the cracks and our tests didn't catch it. My apologizes for introducing a bug- but maybe it's just exposing something that needs to be fixed.

The only thing "more lacking" than our tests are our reviewer free time :stuck_out_tongue_winking_eye: 

Do not worry!  As you said we ended up exposing something that needs some re-thinking.  The code that  created the list of layers was hackish.  We need to think how are we going to add this functionality for `v0.2.0`.

Right now this PR return the *status quo* of broken-but-working.

BibMartin:
Sorry, I've been long to answer ; it seems good to me.

@ocefpaf : I have in mind something to fix this in #203's mindset. I'll try to do it in the coming days.

ocefpaf:
>  I have in mind something to fix this in #203's mindset. I'll try to do it in the coming days.

Great I do not want this hack to leave `v0.1.6` :wink:

andrewgiessel:
What's the current status/who's assigned?  are we waiting on #203 before
tackling this?

ocefpaf:
> What's the current status/who's assigned?

Can I "volunteer" you? :wink: 

> Waiting on #203 before tackling this?


Yes.  I really want all new code to be in the light of #203.

andrewgiessel:
Sure thing

PR #201, opened 2015-08-24
Fix stamentoner URL and attr.

ocefpaf:
Porting #199 to `v0.1.6`


ocefpaf:
Self merging this one since it is already in master.

PR #203, opened 2015-08-26
Features

BibMartin:
This is another version of #170 .
This is not ready for merge. But I would like to share it and have your point of view.

At the cost of scrambling things in version 2, I think this is a more robust implementation of #170's ideas.

The main new thing is a `Feature` object that has `_parent` and `_children` attributes, and a `render` method. Everything inherits from this (Map, Figure, Layer, ...).

* The `_parent` is either another `Feature`, or `None` (if this is the root of the tree).
* The `_children` is an `OrdredDict` of `Feature`s.
* The `render` method calls the `render` method of each children, it can have a bodre effect on the parent, and it returns a string.


ocefpaf:
@BibMartin this is looking awesome.  I really like the clarity of things.  I will take a more careful review once I get back home (I am travelling now).

themiurgo:
:+1: I am not so sure of including Figure as a Feature (I would follow more closely what Leaflet is doing, and keep features being only markers and the likes) but the general direction is great.

BibMartin:
Hi all ! I'm almost finished with what I had in mind.
If you want to have a look at the result : http://nbviewer.ipython.org/github/bibmartin/folium/blob/features/examples/Features.ipynb
Of course, this is only a skeleton. I guess we can create straightforward functions to do things in one line -- and to ensure compatibility with previous version.

BibMartin:
@themiurgo : Sorry, I realize that a former post I did here has not been recorded.
You're true, `Feature` is not a good name for the object I've written, because everything is a `Feature` here while *feature* has another meaning in Leaflet.
Can we rename `Feature` into something else ? Maybe `Node` or `Nodus`, provided that they are the *nodes* of the folium tree.

ocefpaf:
> Maybe Node or Nodus, provided that they are the nodes of the folium tree.

:smile_cat: I like the analogy, but I am not sure if `Node` is a a catchy name, but since I cannot come up with a better one I am OK with it.  @andrewgiessel Wanna to join the discussing?  Any name suggestion?

themiurgo:
How about Element? A tad generic, like in xml and the likes.

ocefpaf:
I like `Element` more than `Node`, but no strong feelings.

BibMartin:
I am okay for `Element` and will change the names soon, if no one cries.
Anyway, this is not a strategic choice: only contributors will face it.

andrewgiessel:
Thanks for the ping @ocefpaf.  In general it seems like this is a very elegant restructuring of the underlying representation of the map; great job @BibMartin !

I like `element` in analogy with XML.  I'm with @themiurgo with the oddness of the `Figure` being a feature.  Why do some of these examples have a `Figure` as the top level element, but the first example uses `map`?  It seems to me figure is an unnecessary extra level of depth, but I could be missing something.

Now for some API nitpicking :grin:  I was a bit surprised at `child.add_to(parent)`.  To me, `parent.add_children()` makes more sense, but also seems a little verbose.  Maybe just `add()` or `append()`?  Furthermore, it seems like `add_children` should at the least take an arbitrary number of parameters via *args, which could be `elements` or iterators of `elements`.  I'm thinking something like this would be nice:
 
    from features import Map, RegularPolygonMarker
    m = Map([0,0], zoom_start=1)
    mk = RegularPolygonMarker([0,0]) 
    mk2 = RegularPolygonMarker([0,45])
    m.add(mk1, mk2)
    # or
    m.add([mk1, mk2])
    # or some combo of the two

I understand that in theory you can go both ways, but for API clarity/simplicity/common usage it seems to make the most sense to only allow adding children to the parent.

BibMartin:
Thanks @andrewgiessel for your feedback.

> I like element in analogy with XML.

The name `Element` is definitely adopted.

> Why do some of these examples have a Figure as the top level element, but the first example uses map?

In my mind, a `Figure` is a container in which you can put several elements.
A `Map` is more specifically a *Leaflet* object into that container.
Doing the distinction between the two enables a clearer code, and extra features. For example:
* you would be able to put several maps in the same figure (I plan to compute this soon, based on something like `fig.add_subplot` in matplotlib).
* you can create a figure hosting a `Vega` graph, rather than a `Map`
* you can inherit from `Figure` with another template if you want you map to be displayed into a specific webpage.
* ...

In the first example, I've just created a shortcut for the user : when you call `my_map._repr_html_()`, then we check that the map's parent is a `Figure` (eventually create one), and call `my_map._parent._repr_html_()` so that the map in anyway represented into a figure.

> Now for some API nitpicking :grin: I was a bit surprised at child.add_to(parent). ...

You tip is really good. There's something to improve here. Today, `add_children` adds only one child in a row, but accept arguments `name` (for specifying the key in the children `OrderedDict`) and `index` (for specifying the position in the children `OrderedDict`). These arguments are important for the core machinery.
A few (disorganized) suggestions :

    m.add(A=childA, B=childB)                    # Handle name, but not the position
    m.add(childA, name='A').add(childB, name=B)  # Handle name and position, but no iterator
    m.iter_add([childA,childB])                  # Specific for iterations

ocefpaf:
> In my mind, a `Figure` is a container in which you can put several elements.

I get the concept and I like the matplotlib analogy, but I am not OK with the name as well.  We are so distant from matplotlib in everything else anyways.

How about `Body`, `Page`, or `Canvas`?

PS: Generating a map and another object along side it will be awesome!

ocefpaf:
@themiurgo Can you be responsible for the merging this one once @BibMartin is finished?

BibMartin:
> We are so distant from matplotlib in everything else anyways.

I agree ; but maybe this would be a step for getting closer to it. Most users know mpl's API ; it would help them to have something close (even if not equal).

> How about `Body`, `Page`, or `Canvas`?

I would avoid `Body` because `body` is just a part of an html page. `Page` or `Canvas` would be okay.

BibMartin:
There's still a lot of work to do here:

* Plug what's missing (geojson, choropleth, imageoverlay, ...)
* Organize code into files/subfiles ...
* Compute tests and adapt them to `test.py`

We can :
* Wait a while that I finish the biggest part of it, and merge then,
* Merge as it is, to let further contributions be in this mindset,
* Merge in another branch,

any point of view ?

ocefpaf:
> Wait a while that I finish the biggest part of it, and merge then,

Waiting to merge or not is up to you.
As soon as you think this is ok to go to master, even with missing pieces,
we'll do it.

I will avoid merging anything but bug fixes until then.  (And even those
will be in a bugfix release branch.)

BibMartin:
I think I'm done (for the moment) and ready to merge.

As told before there's still a lot to do for recomputing plugins (see https://github.com/BibMartin/folium/blob/features/examples/Features.ipynb for the list) and for binding today's API with this structure.

For the moment, I've splitted the code into 3 files only :
* `element.py` stores all the *root* objects (`Element`, `Figure`, `Link`, `MacroElement`, ...)
* `map.py` stores the basics of leaflet maps (`Map`, `Marker`, `Layer`, ...)
* `features.py` stores future *plugins*

@themiurgo @ocefpaf ; if you have time to review... :coffee:

ocefpaf:
I am teaching a workshop this week, but I will try to take a look at it tomorrow.

@themiurgo If you think this is OK do not wait for me.

ocefpaf:
@BibMartin I just gave this a more through review and I am :+1: 

However, I really want @themiurgo's option and,if possible, some of @wrobstory's too.

This is a big change we and all need to be on the same page before merging.

BibMartin:
Well, in fact this PR does not break anything (if you do `import folium`, you still get the former version of the code.)

I will create another PR aiming at plugging the API back ; so that when you `import folium` you get the new objects, with backward compatibility and deprecation warnings.

ocefpaf:
I am planing to merge this next Friday...  *Speak now or forever hold your peace!*

ocefpaf:
And done!

andrewgiessel:
👏

BibMartin:
Thanks @ocefpaf ; sorry I forgot to squash. Is there a way to repair ?

ocefpaf:
No problem.  There are several commits but, for most of them, it is actually better to be separated.

PR #207, opened 2015-09-07
Plug elements in `0.1.*` API

BibMartin:
*(the 36 first commits are the same as #203)*

As written in #203, this breaks the `0.1.*` code in order to plug `elements.py`.

A few tests are passing and a lot of them are failing.

The goal is to adapt/pass the 30 existing test and ensure :

* that you can still use the former API, with deprecation warnings
* that the new API si used everywhere in the background.


ocefpaf:
@BibMartin you are accumulating a lot of work here and it is getting hard to review as time passes.  Yous plans are to merge #203 first, right?

BibMartin:
@ocefpaf Yes, I plan to have #203 merged and then rebase this one to hide the commits they have in common.
btw : this PR is indeed *half-finished* (I opened it to give visibility on where I'm going) ; I expect (hope) it will be easier to review when finished.

ocefpaf:
>  I expect (hope) it will be easier to review when finished.

Thanks!  The problem to review is on my side. (Lack of time :unamused:)

BibMartin:
Can you please have a look at former commit (https://github.com/BibMartin/folium/commit/6832e24444eed695d9fa2215a67fec827cb12b2a). I wonder whether these two tests (`test_geo_json_bad_color` and `test_geo_json_bad_threshold_scale`) are still useful :question:

ocefpaf:
Both `test_geo_json_bad_color` and `test_geo_json_bad_threshold_scale` are probably \* no longer useful.  They might be in a `choropleth` method that may or may not be derived from the `geo_json` method.  I still need to think a little bit more on how to do this properly.

\* I got a GitHub `Sorry, this diff is temporarily unavailable due to heavy server load.` :exclamation:

BibMartin:
@ocefpaf 
> They might be in a choropleth method that may or may not be derived from the geo_json method. I still need to think a little bit more on how to do this properly.

For the moment you can do a choropleth in doing

    m = Map(...)
    g = GeoJson(...)                   # Create the geo_json shapes
    g.add_children(GeoJsonStyle(...)   # Define the colouring
    m.add_children(g)
    m.add_children(ColorScale(...))    # Set the legend

Of course, we will need a shortcut method. Maybe simply

    m = Map(...)
    m.add_children(GeoJson(..., style=GeoJsonStyle(...), legend=ColorScale(...)))

Or as before, `m.geo_json(...)` is not crazy to me.

ocefpaf:
> Or as before, m.geo_json(...) is not crazy to me.

The thing is.  We either let the user be responsible for `ColorScale` or we create a convenience choropleth method to adjust the ColorScale automatically\*.  I would like `geo_json` to just add the GeoJSON assuming very little about the underlying data.

\* We could even use some of PySAL [methods](https://ocefpaf.github.io/python4oceanographers/blog/2015/08/24/choropleth/) to determine the ColorScales.

BibMartin:
:white_check_mark: Green at last ! And ready for review...
@ocefpaf @themiurgo if you have time...

ocefpaf:
Awesome!  I will take a look before the weekend.

Thanks a lot for the hard work @BibMartin!

folium/features.py
```diff
@@ -197,8 +199,46 @@ def __init__(self, data):
...
+            {% endmacro %}
+            """)
+    def render(self,**kwargs):
+        super(TopoJson,self).render(**kwargs)
+
+        figure = self.get_root()
+        assert isinstance(figure,Figure), ("You cannot render this Element "
+            "if it's not in a Figure.")
+
+        figure.header.add_children(\
```
ocefpaf: I guess you do not need that slash.  right?

folium/features.py
```diff
@@ -197,8 +199,46 @@ def __init__(self, data):
...
+            JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/topojson/1.6.9/topojson.min.js"),
+                                   name='topojson')
+
 class GeoJsonStyle(MacroElement):
-    def __init__(self, color_domain, color_code, color_data=None, key_on='feature.properties.color'):
+    def __init__(self, color_domain, color_code, color_data=None,
+                 key_on='feature.properties.color',
+                 weight=1, opacity=1, color='black',
+                 fill_opacity=0.6, dash_array=0):
```
ocefpaf: I am OK with these defaults, but they are different from what we are currently using. Unless you have strong feelings for changing I think we should keep the old defaults.

folium/features.py
```diff
@@ -248,6 +303,33 @@ def render(self,**kwargs):
...
+        self._template = self._env.get_template('color_scale.js')
+
+    def render(self,**kwargs):
+        super(ColorScale,self).render(**kwargs)
+
+        figure = self.get_root()
+        assert isinstance(figure,Figure), ("You cannot render this Element "
+            "if it's not in a Figure.")
+
+        figure.header.add_children(\
```
ocefpaf: The slash again.

folium/folium.py
```diff
@@ -8,247 +8,28 @@
...
-                height_type = 'px'
-                assert height > 0
-            else:
-                height_type = '%'
-                height = int(height.strip('%'))
-                assert 0 <= height <= 100
-        except:
-            msg = "Cannot parse height {!r} as {!r}".format
-            raise ValueError(msg(height, height_type))
-        self.height = height
```
ocefpaf: I am glad to see this go away :wink:

folium/folium.py
```diff
@@ -440,6 +158,10 @@ def line(self, locations,
             or a tuple of the form (Vincent object, 'vis_path.json')
             It is possible to adjust the width of text/HTML popups
             using the optional keywords `popup_width` (default is 300px).
+        latlon: bool, default True
+            Whether locations are given in the form [[lat,lon]] or not ([[lon,lat]] if False).
+            Note that the default GeoJson format is latlon=False,
+            while Leaflet polyline's default is latlon=True.
```
ocefpaf: Since we are changing so much already maybe we should break a little bit more and drop location and use named args for lon and lat too.

tests/test_plugins.py
```diff
@@ -28,89 +28,36 @@ def test_marker_cluster(self):
...
                                            wind_speed=45,
                                            color="#8f8"))
-        mapa.add_plugin(plugins.BoatMarker((46, -30),
+        mapa.add_children(plugins.BoatMarker((46, -30),
                                            heading=-20,
                                            wind_heading=46,
                                            wind_speed=25,
                                            color="#88f"))
-        mapa._build_map()
+        mapa._repr_html_()
```
ocefpaf: Lovely!

ocefpaf:
I am :+1: 

However, I am a bit overwhelmed with all the changes and I'd really appreciate more eyes on this PR before we get it merged.

@themiurgo Do you think you can take a quick look at this?  If not, @BibMartin since  it has been so long you wrote this I guess you are now an independent observer of your own code :stuck_out_tongue: so take a second look and  give me the green to merge.

folium/features.py
BibMartin: I guess you're right. It's a (bad?) habbit I took not to rely on brackets for multi-line.
Tell me and I'll put them away.

ocefpaf: I'd rather not have them there.  But that is just cosmetics...  We still need a better review than mine to get this merged as soon as possible!

folium/features.py
BibMartin: I think we have to keep `weight, opacity, color, fill_opacity, dash_array` because they are the Leaflet's names.

For the others, I'm open to ideas.

Anyway, I think this will change soon, because today's code enables to do choropleth (have a different `fillColor` per feature), but you can imagine changing another parameter (`color` for example) per feature.

If you're okay, I would like to keep it like this and open another PR to tackle it.

folium/folium.py
BibMartin: My goal in `folium/folium.py` was not to break the API to ensure backward compatibility as much a s possible. A new API would be designed progressively in `folium/map.py`

BibMartin:
Thanks @ocefpaf for first review.

>  If not, @BibMartin since it has been so long you wrote this I guess you are now an independent observer of your own code :stuck_out_tongue: so take a second look and give me the green to merge.

I will try the exercise, but I would be more confident if @themiurgo had time to comment : it's not only technical changes but also a different mindset ; I can challenge my code, but barely my own *philosophy*

folium/features.py
ocefpaf: > If you're okay, I would like to keep it like this and open another PR to tackle it.

Yes please!  I am all for bite sized PRs :wink:

folium/folium.py
ocefpaf: Yeah.  That was more a rant than a comment/review.  Just ignore me when I am talking about lon,lat ordering.

ocefpaf:
Last for reviews  :wink: 

Merging this Tomorrow!

ocefpaf:
Thanks @BibMartin  :tada:

PR #218, opened 2015-10-29
Fix python3 issue in Map._repr_html_()

BibMartin:
A tiny PR to fix a bug I had not seen when working with python 3 :

When you do

    foo = b'foo'
    "{}".format(foo)

you get `"b'foo'"` instead of `"foo"`. So you have to decode explicitely from bytes to str

### DIFF
--- a/folium/element.py
+++ b/folium/element.py
@@ -267,7 +267,7 @@ class Figure(Element):
 
         iframe = '<iframe src="{html}" width="{width}px" height="{height}px"></iframe>'\
             .format(\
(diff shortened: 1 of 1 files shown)


ocefpaf:
Looks good.  Finally a simple PR for me to merge :wink:

BibMartin:
@ocefpaf 
 > Folium right now is low in manpower.

But we have really good/reactive mergers

ocefpaf:
:smile:

Issue #219, opened 2015-10-29
Tweak the default width


ocefpaf:
I experiment master with jupyter yesterday and the default width is adding a left-right scroll bar.  We might need to tweak this a little bit.

BibMartin:
You're true, I get the same thing. But it depends on the browser, resolution etc.

There's the same issue in mplleaflet and you can have a look at https://github.com/jwass/mplleaflet/pull/22 for ideas.

We can :
* Keep a fixed size and decrease it so that it will fit in most browsers
* Set the width to "100%" and use a fixed value for the height (600px seems reasonable to me).
* Use a trick inspired from http://fettblog.eu/blog/2013/06/16/preserving-aspect-ratio-for-embedded-iframes/ to set width to 100% and force a given aspect ratio (66% in this example).  

        iframe = '<iframe src="{html}" style="position:absolute;width: 100%;height: 100%;left: 0; top: 0;"></iframe>'\
            .format(\
                    html = "data:text/html;base64,"+base64.b64encode(html.encode('utf8')).decode('utf8'),
                   )            
        aspect_ratio_div = """
            <div style="position: relative; width: 100%;height: 0;padding-bottom: 66%;">{}</div>
            """.format(iframe)
        return aspect_ratio_div

ocefpaf:
> Set the width to "100%" and use a fixed value for the height (600px seems reasonable to me).

Sounds good to me.

andrewgiessel:
I kinda like the fixed proportion trick, but 100% x 600px is fine too.

ps: hi!

ocefpaf:
Fixed in #221.

Thanks @BibMartin for the PR and @andrewgiessel for highlighting a better option than mine :wink:

PR #221, opened 2015-10-30
Fix issue #219

BibMartin:
Finally I used the "trick" ;-)

### DIFF
--- a/folium/element.py
+++ b/folium/element.py
@@ -175,7 +175,23 @@ _default_css = [
     ]
 
 class Figure(Element):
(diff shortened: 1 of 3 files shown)


ocefpaf:
@BibMartin good job.  It looks awesome in Firefox and Chrome.  I re-sized the window in many different ways and the map always looked good!

I only had a problem in Chromium.  There is an intermittent line in the tiles (happen with all tile sources):

tiles

Since this comes and goes I guess we can just assume it is a Chrome thing and leave it be.

Merging this.

Issue #206, opened 2015-09-07
Retina Maps


francisbautista:
Will there be plans for supporting Retina OSM maps for Folium? I'm having meter-pixel issues with screens that are Retina and non-retina.

ocefpaf:
Never thought about that but it shouldn't be hard to implement it for the tiles.  I will leave [this](http://leafletjs.com/reference.html#tilelayer-detectretina) for reference.

BibMartin:
See #222

themiurgo:
Can this be closed?

BibMartin:
Sure ; it's solved by #222

PR #222, opened 2015-10-30
Add Retina Display Support (issue #206)

BibMartin:
I have no retina display ; so I have no possibility to see how Leaflet reacts.
Maybe @francisbautista can confirm that this works.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -21,7 +21,7 @@ class Map(MacroElement):
                  left="0%", top="0%", position='relative',
                  tiles='OpenStreetMap', API_key=None, max_zoom=18, min_zoom=1,
                  zoom_start=10, attr=None, min_lat=-90, max_lat=90,
-                 min_lon=-180, max_lon=180):
+                 min_lon=-180, max_lon=180, detect_retina=False):
         """Create a Map with Folium and Leaflet.js
 
         Generate a base map of given width and height with either default
@@ -58,6 +58,9 @@ class Map(MacroElement):
             Initial zoom level for the map.
         attr: string, default None
             Map tile attribution; only required if passing custom tile URL.
+        detect_retina: bool, default False
+            If true and user is on a retina display, it will request four tiles of half the specified
+            size and a bigger zoom level in place of one to utilize the high resolution.
 
         Returns
         -------
@@ -102,7 +105,7 @@ class Map(MacroElement):
         self.max_lon = max_lon
 
         self.add_tile_layer(tiles=tiles, min_zoom=min_zoom, max_zoom=max_zoom,
-                            attr=attr, API_key=API_key)
+                            attr=attr, API_key=API_key, detect_retina=detect_retina)
 
         self._template = Template(u"""
         {% macro header(this, kwargs) %}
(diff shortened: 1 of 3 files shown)


andrewgiessel:
I can try it too

ocefpaf:
@andrewgiessel can you review and merge this one too?

andrewgiessel:
Ya let me try tonight!

andrewgiessel:
Ok, this looks good to go!  I verified this 2 different ways:

1) on my laptop and on an external screen, I loaded a test notebook and refreshed the page on both screens.  I could see the size of the tiles were smaller on my laptop than on the external screen.

2) This is kinda cool- you can use chrome's developer tools to go into 'device' mode, where you can simulate changing the resolution, size and network speed (!) the device that loads the page.  Here are two screen shots of doing this with high and low density screen (and low speed to aid in catching the tiles loading)

screen shot 2015-11-03 at 1 00 25 pm

screen shot 2015-11-03 at 12 59 16 pm

So, I think this is good.  Can I merge?  It's one commit- do I need to clean it up at all?

ocefpaf:
> So, I think this is good. Can I merge?

Merge away!

Thanks @BibMartin for the implementation and @andrewgiessel for the review.

themiurgo:
Do you guys know why Leaflet defaults `detect_retina` to false? It shouldn't have any drawbacks so I would be tempted to default it to True on folium.

andrewgiessel:
My guess is bandwidth. Downloading 2x the tiles you can see

themiurgo:
Mapbox-powered maps (FourSquare & co.) and GMaps defaults to True, despite bandwidth. :grinning:

ocefpaf:
I guess that there are other reasons:

https://github.com/Leaflet/Leaflet/issues/1985

themiurgo:
After all, it seems it's beacause of this: "detectRetina is optional (false by default) to prevent map features (like city titles) from being too small to read on retinal tiles" (source: https://github.com/Leaflet/Leaflet/commit/443181af07391cd624f6328ffd9d0e30351642bf). I don't think we care about that, should we turn it on by default?

ocefpaf:
How about the performance issue I mentioned above?

Ironically it happens only on the IPad using Safari:
> I test it too on Android mobile with detectRetina: true, and no problem. Is only for Safari on iOs.

themiurgo:
Sorry, I thought it was solved because I saw "closed". I'll test it and see if still applies, that's an issue from 2013 so it might have been solved from iOS side.

ocefpaf:
I guess it is worth asking that question in the leaflet issue page.

https://github.com/Leaflet/Leaflet/issues/4014

themiurgo:
On my iPhone there were no performance issues. But yes, the text labels are incredibly small and unreadable, I think that is the reason behind that.

ocefpaf:
We have to bare in mind that setting `false` as the default we ensure the results will be the same in any platform, retina or no. By forcing the user to set it to `true` the user acknowledges that he is targeting platforms where that is desired.

ocefpaf:
We have an official answer:

https://github.com/Leaflet/Leaflet/issues/4014#issuecomment-156441089

themiurgo:
Cool, let's leave it as it is. :+1:

ocefpaf:
We should improve the docs though.  Highlighting the option and say why `false` is the default.

Issue #213, opened 2015-10-15
Heatmaps on Folium


francisbautista:
Is there a way to use heatmaps on folium w/o going back to Leaflet.js and its heatmap plugin set?

ocefpaf:
@francisbautista that would be a nice addition.  Folium right now is low in manpower.  PRs are welcome :wink:

ocefpaf:
Hi @francisbautista,

@BibMartin did an awesome job implementing a plugin system for folium.  He just used that system to add the `leaflet-heat.js`.  Feedback is appreciated!  (And if you produce a cool map consider sending it to https://github.com/ocefpaf/folium_notebooks.  I am putting a gallery together.)

(See #224)

PR #224, opened 2015-10-30
Create HeatMap as suggested in #213

BibMartin:
Using https://github.com/Leaflet/Leaflet.heat

### DIFF
--- a/folium/plugins/heat_map.py
+++ b/folium/plugins/heat_map.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""
+Heat map
+--------
+
+Create a HeatMap layer
+"""
+import json
+from jinja2 import Template
+
+from folium.element import JavascriptLink, Figure
+from folium.map import TileLayer
+
+class HeatMap(TileLayer):
+    def __init__(self, data, name=None,
+                 min_opacity = .5, max_zoom=18, max_val=1.,
+                 radius=25, blur=15, gradient=None,
+                 overlay = True):
+        """Create a Heatmap layer
+
+        Parameters
+        ----------
+        data : list of points of the form [lat, lng] or [lat, lng, weight]
+            The points you want to plot.
+            You can also provide a numpy.array of shape (n,2) or (n,3).
+        name : str
+            The name of the layer that will be created.
+        min_opacity  : default 1.
+            The minimum opacity the heat will start at.
+        max_zoom : default 18
+            Zoom level where the points reach maximum intensity (as intensity scales with zoom),
+            equals maxZoom of the map by default
+        max_val : float, default 1.
+            Maximum point intensity
+        radius : int, default 25
+            Radius of each "point" of the heatmap
+        blur : int, default 15
+            Amount of blur
+        gradient : dict, default None
+            Color gradient config. e.g. {0.4: 'blue', 0.65: 'lime', 1: 'red'}
+        """
+        super(TileLayer, self).__init__()
+        self._name = 'HeatMap'
+        self.tile_name = name if name is not None else self.get_name()
+
+        self.data = [[x for x in line] for line in data]
+        self.min_opacity = min_opacity
+        self.max_zoom = max_zoom
+        self.max_val = max_val
+        self.radius = radius
+        self.blur = blur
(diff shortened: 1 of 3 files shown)


ocefpaf:
You are on :fire:  this week!

BibMartin:
:musical_note: Hurray ! It's holy-holyday ! :musical_note:

andrewgiessel:
Oh yahhhh! :+1: :tada:

ocefpaf:
heatmap

folium/plugins/heat_map.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, data, name=None,
+                 min_opacity = .5, max_zoom=18, max_val=1.,
+                 radius=25, blur=15, gradient=None,
+                 overlay = True):
+        """Create a Heatmap layer
+
+        Parameters
+        ----------
+        data : list of points of the form [lat, lng, weight]
+            The point you want to plot.
```
ocefpaf: Let's support 3D numpy as well.  We can listify them internally.

Question:  In your test example you pass a 2D array and things work.  I guess that we should raise an error instead, no?

tests/test_plugins.py
```diff
@@ -134,3 +134,9 @@ def test_timestamped_geo_json(self):
         mape = folium.Map([47, 3], zoom_start=1)
         mape.add_children(plugins.TimestampedGeoJson(data))
         mape._repr_html_()
+
+    def test_heat_map(self):
+        data = (np.random.normal(size=(100,2))*np.array([[1,1]])+np.array([[48,5]])).tolist()
+        mapa = folium.Map([48., 5.], tiles='stamentoner', zoom_start=6)
+        mapa.add_children(plugins.HeatMap(data))
+        mapa._repr_html_()
```
ocefpaf: This test only if the HTML is created, correct? I guess that all the plugins are tested like that...  Not sure if we can do better.

folium/plugins/heat_map.py
BibMartin: the js library accepts arrays of shape (n,2) and (n,3) ; so we shall do the same thing. A new commit just do that.

tests/test_plugins.py
BibMartin: Correct.

We can do a bit better in doing the same thing as `test_folium.py` : compare the output of `.render()` with code generated from a validation template.
But this is a heavy procedure and I'm not sure this is needed for all plugins (am I lazy ?).

folium/plugins/heat_map.py
ocefpaf: > the js library accepts arrays of shape (n,2) and (n,3) ; so we shall do the same thing.

OK.  If the JS support let's do that then... But what is the reason?  I confess I just glimpsed over the docs, but I could not find a rational for that.

tests/test_plugins.py
ocefpaf: > But this is a heavy procedure and I'm not sure this is needed for all plugins (am I lazy ?).

Heavy?  Yes.
Are you lazy?  No!
Should someone else do that? (Me :unamused:)

ocefpaf:
It is all good to me.  Just a minor typo and then good to go!

Thanks for the great work @BibMartin!

BibMartin:
typo fixed ; commits squashed ...

ocefpaf:
And merged!

jackboot7:
Amazing work @BibMartin, just used your HeatMap plugin to generate a pretty map of geolocalized data.

Thank you!

ocefpaf:
Hi @jackboot7 consider sending your notebook to our ["gallery."](https://github.com/python-visualization/folium/tree/master/examples)

jackboot7:
@ocefpaf: Sure, I'll do it as soon as I'm done with my program.

ocefpaf:
Thanks :+1:

jackboot7:
@ocefpaf How would you prefer that I share my usage of Folium with you, guys? I made a blog post about the my use case (http://www.jackboot7.com/visualizing-tweets.html )

ocefpaf:
The blog reference is fine.  We don't have a list of use cases right now but I want to start one.

However, if you find the time it would be nice to get a PR with that notebook in our examples folder:

https://github.com/python-visualization/folium/tree/master/examples

Issue #226, opened 2015-10-31
Create `Map.create_map`


BibMartin:
As discussed with @ocefpaf in https://github.com/BibMartin/folium/commit/5d665c553adf6dc61cea729f382923108aaf9860,
we need to build a new `create_map()` method that enables to write a map into a file.
(the former method has been destoyed by 5d665c553adf6dc61cea729f382923108aaf9860)

BibMartin:
See #227

PR #227, opened 2015-10-31
Add save and create_map methods

BibMartin:
As mentionned in #226.

Finally, I did a `save` method for all elements, and a `create_map` method for `Map` only, with a deprecation warning.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -30,6 +30,23 @@ class Map(_Map):
     """This class inherits from the map.Map object in order to provide bindings to
     former folium API.
     """
+    def create_map(self, path='map.html', plugin_data_out=True, template=None):
+        """Write Map output to HTML.
+
+        Parameters:
+        -----------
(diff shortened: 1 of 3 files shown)


folium/folium.py
```diff
@@ -30,6 +30,23 @@ class Map(_Map):
...
+        Parameters:
+        -----------
+        path: string, default 'map.html'
+            Path for HTML output for map
+        plugin_data_out: boolean, default True
+            Deprecated, not used anymore
+        template: string, default None
+            Deprecated, not used anymore
+        """
+        warnings.warn("%s is deprecated. Use %s instead" % ("Map.create_map",
```
ocefpaf: Awesome!  I was going to suggest that.  I like `save` or `save_map` better than `create_map`.

ocefpaf:
Merging this one.  Thanks again @BibMartin!

Issue #211, opened 2015-10-06
DOI using GitHub+Zenodo


ocefpaf:
Every now and then I get e-mails from academics asking how to cite `folium`.  I guess it is time to create a DOI number for it.

@andrewgiessel is this something you are interested in doing?

andrewgiessel:
Sure!

Looks like we can follow this guide [here](https://guides.github.com/activities/citable-code/).  It seems we need to have an official release (in :octocat: terms) to make it work.  We can use the most recent release, or wait till we do the impending one.  Either way is fine by me.

andrewgiessel:
Ok, I did everything we need.  When we make a new release, it'll trigger the generation of a DOI.

I'm assigning myself to finish the process after we put out our next release.

ocefpaf:
Thanks @andrewgiessel!  Zenodo issues a badge, right?  Can you put that in the README?

andrewgiessel:
We won't get a DOI (and then a badge) until we make a new release.  It uses
a GH webhook to generate everything, which we haven't triggered.

ocefpaf:
Ah OK.  But how about minting one for `v0.1.5`?

PS: Are you interested in taking care of `v0.1.6` :wink:

andrewgiessel:
Hey guys!

https://zenodo.org/record/33534#.VkIUpK6rSL4

=)

Opening a PR to add the badge shortly

ocefpaf:
:+1:

PR #240, opened 2015-11-10
add DOI badge

andrewgiessel:


andrewgiessel:
Finishes #211 !  With every release, it should update the DOI.  Not sure if we'll have to change the badge code or not (probably not?)

ocefpaf:
> Not sure if we'll have to change the badge code or not (probably not?)

I guess that with `latestdoi` we don't need to update the badge.

PR #250, opened 2015-11-15
Div icon

ocefpaf:
@BibMartin this is my first attempt to dive in the new code.  So I have a few questions:

- The original `DivIcon` class was inheriting from `MacroElement`.  I changed it to `Icon`, although I don't think it matters in this case.
- You carefully tried to preserve the method `div_markers`, but I am starting to think that should go away.  It is not really what the original `leaflet` `divIcon` is all about.
- If we agree that `div_markers` should ho we can safely remove the `static_div_icon.js` and the tests for that method.
- I believe there more `<templates>.js` that we can remove.

Note that in this PR I already removed the template.  To see the failure keeping that go to:

https://travis-ci.org/ocefpaf/folium/jobs/91228684

:fr: I am praying for the victims of this horrible attack.  I Hope everybody close to you are alright. :fr:

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -19,9 +19,9 @@ class WmsTileLayer(TileLayer):
     def __init__(self, url, name=None,
                  format=None, layers=None, transparent=True,
                  attribution=None):
-        """TODO docstring here
-        Parameters
-        ----------
+        """
+        TODO docstring here
+
         """
         super(TileLayer, self).__init__()
         self._name = 'WmsTileLayer'
@@ -56,7 +56,10 @@ class RegularPolygonMarker(MacroElement):
                  color='black', opacity=1, weight=2,
                  fill_color='blue', fill_opacity=1,
                  number_of_sides=4, rotation=0, radius=15):
-        """TODO : docstring here"""
+        """
+        TODO docstring here
+
+        """
         super(RegularPolygonMarker, self).__init__()
         self._name = 'RegularPolygonMarker'
         self.location = location
@@ -106,7 +109,10 @@ class RegularPolygonMarker(MacroElement):
 class Vega(Element):
     def __init__(self, data, width='100%', height='100%',
                  left="0%", top="0%", position='relative'):
-        """TODO : docstring here"""
+        """
+        TODO docstring here
+
+        """
         super(Vega, self).__init__()
         self._name = 'Vega'
         self.data = data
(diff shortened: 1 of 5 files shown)


ocefpaf:
This PR is the first one of a long term plan to get us closer to matpltolib via mplexporter and mplleaflet.

See: http://nbviewer.ipython.org/gist/ocefpaf/ab173ae7a7ee40430cdf

BibMartin:
> :fr: I am praying for the victims of this horrible attack. I Hope everybody close to you are alright. :fr:

Thank you for your prayers ; everyone I know is fine, but I have a deep thought for past and future victims.

BibMartin:
> This PR is the first one of a long term plan to get us closer to matpltolib via mplexporter and mplleaflet.

That's good idea ; I guess we can end-up with a `MatplotlibFigure` object that does mplleaflet's job.
I wonder whether we shall do this in calling mplleaflet, or get inspiration from it and do it in folium's style.

themiurgo:
Great PR @ocefpaf . I love the notebook with arrow DivIcons, you just gave me a great idea to visualise GPS trajectories. :+1: 

@BibMartin There's nothing else to add to what has been already said in these days. Let's all hope in a better future. :fr:

ocefpaf:
> you just gave me a great idea to visualise GPS trajectories.

Glad that was helpful!  My real intent is to show ocean currents though :wink:  (Soon in a notebook example.)

ocefpaf:
> That's good idea ; I guess we can end-up with a MatplotlibFigure object that does mplleaflet's job.
I wonder whether we shall do this in calling mplleaflet, or get inspiration from it and do it in folium's style.

I like the idea of a `MatplotlibFigure`.  I think that the best course of action is to factor out the `fig2json`.
The fig2json is pretty simple and the only dependency is `mplexporter` and of course `matplolitb`.

PS: I removed the `div_markers` method and tests.  I added some tests for the new `DivIcon`.
Looking at this again I really don't think we should inherent from icon.  Most of we inherent is not applicable to the `DivIcon` (rotation, color, etc).  What do you think?  Should I just inherit from `MacroElement`?

ocefpaf:
> - The original `DivIcon` class was inheriting from `MacroElement`.  I changed it to `Icon`, although I don't think it matters in this case.

@BibMartin I changed it back to `MacroElement` because, even though divIcon is an Icon, it does not make sense to expose the `Icon` methods here.

Let me know how bad this is and if we can salvage some ideas for a better PR :grimacing:

folium/features.py
```diff
@@ -363,39 +364,73 @@ def render(self, **kwargs):
...
+                    {% if this.icon_size %}iconSize: [{{this.icon_size[0]}},{{this.icon_size[1]}}],{% endif %}
+                    {% if this.icon_anchor %}iconAnchor: [{{this.icon_anchor[0]}},{{this.icon_anchor[1]}}],{% endif %}
+                    {% if this.popup_anchor %}popupAnchor: [{{this.popup_anchor[0]}},{{this.popup_anchor[1]}}],{% endif %}
+                    {% if this.className %}className: '{{this.className}}',{% endif %}
+                    {% if this.html %}html: '{{this.html}}',{% endif %}
                     });
                 {{this._parent.get_name()}}.setIcon({{this.get_name()}});
             {% endmacro %}
-            """)
+            """)  # noqa
```
BibMartin: What does `# noqa` mean ?

ocefpaf: No questions asked.  It serves to tell humans and automatic tools to not mess with that line.

BibMartin:
@ocefpaf 
It looks fine to me. I just don't understand why you've dropped out the `class` structure in `test_features.py` and unstacked everything in different functions. But I guess this is not the *active* part of the PR.

> Let me know how bad this is and if we can salvage some ideas for a better PR :grimacing:

It's good to me. Just maybe I should write proper code so that you don't need to pep8 and docstring it.

ocefpaf:
> I just don't understand why you've dropped out the `class` structure in `test_features.py` and unstacked everything in different functions.

Something about `pytest` was not working properly.  I can't remember what.  However, I believe that `pytest` is designed to work better that way.

I will wait for #261 and #262.  (Then I will go back to this one, rebase, polish, and ping you again :smile:)

ocefpaf:
Rebased and good for another round if review.

BibMartin:
Merged : thanks @ocefpaf

Issue #257, opened 2015-11-18
Marker and other features are not accessible.


themiurgo:
Since 9abdd8f4d8ff5d106ff3f681c9150a039b9d7872 Marker, Popup and other features are not accessible from `folium.features`.

I believe what happened is that pep8 marked the imports as unused (as those classes are not used in `folium.features`), and tests did not catch that because they still involve the legacy usage (e.g. `Map.simple_marker` instead of `Map.add_to(Marker)`).

Two things we need to do to avoid this from happening again:
- We should decide where to access features from. I propose the root namespace).
- We should add / modify tests to use also the new feature addition.

ocefpaf:
> I believe what happened is that pep8 marked the imports as unused

Not really.  I removed those on purpose as I thought they should not be on features.  But I am re-thinking my decision, since it is easier to access everything in there.

> We should decide where to access features from. I propose the root namespace

I like that idea.  We need to discuss this with @BibMartin and @andrewgiessel.

> We should add / modify tests to use also the new feature addition

Again... Our tests are our Achilles heels.  Let's leave this issue open until we implement something in that regard.

ocefpaf:
> Again... Our tests are our Achilles heels. Let's leave this issue open until we implement something in that regard.

I am closing this and opening another issue.

PR #258, opened 2015-11-18
Make features accessible from root

themiurgo:
This temporarily makes every feature accessible from the root namespace to solve #257 . We can later remove features that are less used if we prefer, or keep it like this.

### DIFF
--- a/folium/__init__.py
+++ b/folium/__init__.py
@@ -3,4 +3,10 @@ from __future__ import absolute_import
 
 __version__ = '0.2.0.dev'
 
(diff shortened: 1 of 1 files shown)


ocefpaf:
Are you finished with this one?  Ready for merge?  (I am just asking because I had to refresh the changes.  Probably a rebase.)

themiurgo:
Yes, merge away if you are ok with the changes!

ocefpaf:
I am up for experiment at least :smile_cat:

PR #262, opened 2015-11-27
Fix #223

### LINKED ISSUE #223: key_on='features.properties.key' seems to be broken
When passing the `key_on=` option the docs mentioned that `key_on='features.properties.key'` is valid syntax, but that is not working.

ocefpaf:
Should be `feature.properties.key` and not `features.properties.key`.


ocefpaf:
Self merging this one.  It won't matter soon anyways. I hope to see `geo_json` dead soon :wink:

Issue #259, opened 2015-11-27
Unable to assign string directly to popup using master branch


jelmelk:
Doing some small testing and I did the follwing:

    some_marker = folium.Marker(location=[45.5215, -122.6261], popup='test')

I would get this error: `AttributeError: 'str' object has no attribute 'get_name'`. If I change that to `popup=folium.Popup('test')` it works fine. Is that intended?

Main reason I'm asking is because it seems somewhat tedious having to call that to assign a string to the `popup`.

BibMartin:
@jelmelk 
>  Is that intended?
> Main reason I'm asking is because it seems somewhat tedious having to call that to assign a string to the popup.

In the background, we cannot do without the `Popup` object, because one may be willing to create more complicated popups (with a vega chart for instance).

But I agree it's tedious, and in the foreground we can shortcut things in the `Marker` code to catch cases when the arg `popup` is a `str` (or `bytes`), and to replace it by `Popup(popup)`.

It'd be great if you did a PR for that (and thus join the <a href="https://github.com/python-visualization/folium/network/members">happy community of contributors</a>). Otherwise, I'll do it on Monday.

jelmelk:
I'm looking at that right now. Would it be better to change each marker individually since the error rises at `map.py` on line 389 for `Marker` and `element.py` on line 414 for `CircleMarker` etc...

They all have `element.py` line 46 in common which is where the exception is actually being called from. What would be the preference in this case?

BibMartin:
I imagine you can simply fix it in the `Marker.__init__` is testing when `isinstance(popup,text_type) or isinstance(popup,str_type)` and then overwrite the popup attribute.

themiurgo:
We should strive to generalise this, while we're at it. You might want to have a popup also for Polylines, for example.

BibMartin:
Fixed by #283. Closing this.

PR #265, opened 2015-11-27
Popup can accept strings explicitly

jelmelk:
This is to address https://github.com/python-visualization/folium/issues/259. Assuming I did the PR properly.


ocefpaf:
Awesome @jelmelk!  Thanks for the PR!

Can you build a test for this behavior?

jelmelk:
Am I correct to assume you're talking about test_folium.py?

ocefpaf:
You are adding this behavior for `Marker`, ` CircleMarker`, and `RegularPolygonMarker`.  Correct?

I guess we should also add this for lines...

ocefpaf:
> Am I correct to assume you're talking about test_folium.py?

Yep.

jelmelk:
> You are adding this behavior for Marker, CircleMarker, and RegularPolygonMarker. Correct?

Yes, those are the only ones that needed it. So would it be a single function for each one?

ocefpaf:
> Yes, those are the only ones that needed it. So would it be a single function for each one?

Maybe we could have been more clever and try to avoid that repetition, but your implementation is OK. I will take a better look tomorrow (going to bed now).

@BibMartin any observations?

jelmelk:
I agree, but like I said in https://github.com/python-visualization/folium/issues/259#issuecomment-160141767 it could be done individually or in one spot that catches all three. Let me know tomorrow, I can change it if needed.

BibMartin:
Hi @jelmelk ; thanks for the PR !
Provided the `Marker` class is very simple, I think you can implement it only in `Marker` and let the other classes inherit from it. 
BTW it would make sens to have all marker stuff inherit from the same root class.
Anyway, we're only sparing 6 lines of code this way :arrow_right:  it's not strategic and I would accept the PR like this.

jelmelk:
Due to the branch mix up on my part, I'm closing this PR.

Issue #198, opened 2015-08-21
Read GeoJSON style from file as default


ocefpaf:
We have a template ([geojson_style.js](https://github.com/python-visualization/folium/blob/master/folium/templates/geojson_style.js)) for styling the GeoJSON object.

```jinja
function {{ style }}(feature) {
    return {
        {%- if quantize_fill %}
        fillColor: {{ quantize_fill }},
        {%- else %}
        fillColor: '{{ fill_color }}',
        {%- endif %}
        weight: {{ line_weight }},
        opacity: {{ line_opacity }},
        color: '{{ line_color }}',
        fillOpacity: {{ fill_opacity }}
    };
}
```

The template is filled [here](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L1014-L1018) and that usually happens with the default values.  But if the GeoJSON file has `properties` defined a simple `function style_1(feature) {return feature.properties;}` would suffice and, in same cases, that is the expected behavior.

I would like to discuss here the possibility of making "reading from properties" the default behavior instead of the default values we have and leave the `kwargs` as an override option.

ocefpaf:
It won't be default, but a convince function for the simple-spec will be available.

See #266.

ocefpaf:
We should also implement a way to add `onEachFeature`.

(See https://github.com/python-visualization/folium/issues/6#issuecomment-161994518)

baobobs:
It would be great to have popup capability added to geojson / choropleth layers. Eager to see `onEachFeature` added. Thanks guys!

ocefpaf:
Closing this as we never found a champion to write the simple style.

jgoad:
@ocefpaf Sorry I just followed the rabbit hole of issues to get to this point. Does this mean there is no implementation for popups on geojson layers?

ocefpaf:
> @ocefpaf Sorry I just followed the rabbit hole of issues to get to this point. Does this mean there is no implementation for popups on geojson layers?

I plan to work on this soon, unless someone beats me to it.

jzf2101:
Any news on this issue?

Issue #120, opened 2015-05-10
Support GeoJSON simplestyle


themiurgo:
It would be great to support GeoJSON simplestyle (https://github.com/mapbox/simplestyle-spec), as it already happens in geojson.io and GitHub.

ocefpaf:
@BibMartin Now that the re-factoring dust is settling we should start thinking about this.

BibMartin:
It can be done base on #266 discussion. See @ocefpaf 's idea.

themiurgo:
When this is ready I'd make it default, as I can see only advantages and no drawbacks. What do you think?

ocefpaf:
We never found a champion for this implementation and I am biased to leave `folium` as-is. It is relatively simple for someone to create `style_function` that parses the `simplestyle`.

PR #266, opened 2015-11-28
replace GeoJsonStyle by kwarg style_function in GeoJson

BibMartin:
To simplify choropleth generation.

You have to provide a function `feature -> style` which is to me the most generic API.

I need to update `Map.geo_json` and tests before merge.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -166,7 +166,7 @@ class Vega(Element):
 
 
 class GeoJson(MacroElement):
-    def __init__(self, data):
+    def __init__(self, data, style_function=None):
         """
         Creates a GeoJson plugin to append into a map with
         Map.add_plugin.
@@ -174,12 +174,14 @@ class GeoJson(MacroElement):
         Parameters
         ----------
         data: file, dict or str.
-        The GeoJSON data you want to plot.
-        * If file, then data will be read in the file and fully
-          embedded in Leaflet's JavaScript.
-        * If dict, then data will be converted to JSON and embedded
-          in the JavaScript.
-        * If str, then data will be passed to the JavaScript as-is.
+            The GeoJSON data you want to plot.
+            * If file, then data will be read in the file and fully
+              embedded in Leaflet's JavaScript.
+            * If dict, then data will be converted to JSON and embedded
+              in the JavaScript.
+            * If str, then data will be passed to the JavaScript as-is.
+        style_function: function, default None
+            A function mapping a GeoJson Feature to a style dict.
 
         Examples
         --------
@@ -190,22 +192,43 @@ class GeoJson(MacroElement):
         >>> # Providing string.
         >>> GeoJson(open('foo.json').read())
 
+        >>> # Providing a style_function that put all states in green, but Alabama in blue.
+        >>> style_function=lambda x: {'fillColor': '#0000ff' if x['properties']['name']=='Alabama' else '#00ff00'}
+        >>> GeoJson(geojson, style_function=style_function)
         """
         super(GeoJson, self).__init__()
         self._name = 'GeoJson'
         if 'read' in dir(data):
-            self.data = data.read()
+            self.data = json.load(data)
         elif type(data) is dict:
-            self.data = json.dumps(data)
-        else:
             self.data = data
+        else:
+            self.data = json.loads(data)
+
+        if 'features' not in self.data.keys():
+            # Catch case when GeoJSON is just a single Feature or a geometry.
+            if not (isinstance(self.data, dict) and 'geometry' in self.data.keys()):
+                # Catch case when GeoJSON is just a geometry.
+                self.data = {'type' : 'Feature', 'geometry' : self.data}
+            self.data = {'type' : 'FeatureCollection', 'features' : [self.data]}
+
+        if style_function is None:
+            style_function = lambda x: {}
+        self.style_function = style_function
 
         self._template = Template(u"""
             {% macro script(this, kwargs) %}
-                var {{this.get_name()}} = L.geoJson({{this.data}}).addTo({{this._parent.get_name()}});
+                var {{this.get_name()}} = L.geoJson({{this.style_data()}})
+                    .addTo({{this._parent.get_name()}})
+                    .setStyle(function(feature) {return feature.properties.style;});
             {% endmacro %}
             """)  # noqa
 
+    def style_data(self):
+        for feature in self.data['features']:
+            feature.setdefault('properties',{}).setdefault('style',{}).update(
+                self.style_function(feature))
+        return json.dumps(self.data)
 
 class TopoJson(MacroElement):
     def __init__(self, data, object_path):
@@ -244,81 +267,6 @@ class TopoJson(MacroElement):
         figure.header.add_children(
             JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/topojson/1.6.9/topojson.min.js"),  # noqa
             name='topojson')
-
-
-class GeoJsonStyle(MacroElement):
-    def __init__(self, color_domain, color_code, color_data=None,
-                 key_on='feature.properties.color',
-                 weight=1, opacity=1, color='black',
-                 fill_opacity=0.6, dash_array=0):
-        """
-        TODO docstring here
-
-        """
-        super(GeoJsonStyle, self).__init__()
-        self._name = 'GeoJsonStyle'
-
-        self.color_domain = color_domain
-        self.color_range = color_brewer(color_code, n=len(color_domain))
-        self.color_data = json.dumps(color_data)
-        self.key_on = key_on
-
-        self.weight = weight
-        self.opacity = opacity
-        self.color = color
-        self.fill_color = color_code
-        self.fill_opacity = fill_opacity
-        self.dash_array = dash_array
-
-        self._template = Template(u"""
-            {% macro script(this, kwargs) %}
-                {% if not this.color_range %}
-                    var {{this.get_name()}} = {
-                        color_function : function(feature) {
-                            return '{{this.fill_color}}';
-                            },
-                        };
-                {%else%}
-                    var {{this.get_name()}} = {
-                        color_scale : d3.scale.threshold()
-                              .domain({{this.color_domain}})
-                              .range({{this.color_range}}),
-                        color_data : {{this.color_data}},
-                        color_function : function(feature) {
-                            {% if this.color_data=='null' %}
-                                return this.color_scale({{this.key_on}});
-                            {% else %}
-                                return
-                                  this.color_scale(this.color_data[{{this.key_on}}]);
-                            {% endif %}
-                            },
-                        };
-                {%endif%}
-
-                {{this._parent.get_name()}}.setStyle(function(feature) {
-                    return {
-                        fillColor: {{this.get_name()}}.color_function(feature),
-                        weight: {{this.weight}},
-                        opacity: {{this.opacity}},
-                        color: '{{this.color}}',
-                        fillOpacity: {{this.fill_opacity}},
-                        dashArray: '{{this.dash_array}}'
-                        };
-                    });
-            {% endmacro %}
-            """)
-
-    def render(self, **kwargs):
-        super(GeoJsonStyle, self).render(**kwargs)
-
-        figure = self.get_root()
-        assert isinstance(figure, Figure), ("You cannot render this Element "
-                                            "if it's not in a Figure.")
-
-        figure.header.add_children(
-            JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js"),  # noqa
-            name='d3')
-
 
 class ColorScale(MacroElement):
     def __init__(self, color_domain, color_code, caption=""):
@@ -476,7 +424,6 @@ class CircleMarker(MacroElement):
                 .addTo({{this._parent.get_name()}});
             {% endmacro %}
             """)
-
 
 class LatLngPopup(MacroElement):
     def __init__(self):
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -16,7 +16,7 @@ from folium.six import text_type, binary_type
 from .map import Map as _Map
 from .map import Icon, Marker, Popup, FitBounds
 from .features import (WmsTileLayer, RegularPolygonMarker, Vega, GeoJson,
-                       GeoJsonStyle, CircleMarker, LatLngPopup,
+                       CircleMarker, LatLngPopup,
                        ClickForMarker, ColorScale, TopoJson, PolyLine,
                        MultiPolyLine, ImageOverlay)
 from .utilities import color_brewer, write_png
@@ -442,10 +442,17 @@ class Map(_Map):
 #
 #        self.template_vars.update({'fit_bounds': fit_bounds_str.strip()})
 
-    def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
+    def geo_json(self, *args, **kwargs):
+        """This method is deprecated and will be removed in v0.2.1. See
+        `Map.choropleth` instead.
+        """
+        warnings.warn('This method is deprecated. Please use Map.choropleth instead.')
+        return self.choropleth(*args, **kwargs)
+
+    def choropleth(self, geo_path=None, geo_str=None, data_out='data.json',
                  data=None, columns=None, key_on=None, threshold_scale=None,
                  fill_color='blue', fill_opacity=0.6, line_color='black',
-                 line_weight=1, line_opacity=1, legend_name=None,
+                 line_weight=1, line_opacity=1, legend_name="",
                  topojson=None, reset=False):
         """Apply a GeoJSON overlay to the map.
 
@@ -506,8 +513,8 @@ class Map(_Map):
             GeoJSON geopath line weight.
         line_opacity: float, default 1
             GeoJSON geopath line opacity, range 0-1.
-        legend_name: string, default None
-            Title for data legend. If not passed, defaults to columns[1].
+        legend_name: string, default empty string
+            Title for data legend.
         topojson: string, default None
             If using a TopoJSON, passing "objects.yourfeature" to the topojson
             keyword argument will enable conversion to GeoJSON.
@@ -520,17 +527,14 @@ class Map(_Map):
 
         Example
         -------
-        >>> m.geo_json(geo_path='us-states.json', line_color='blue',
+        >>> m.choropleth(geo_path='us-states.json', line_color='blue',
                       line_weight=3)
-        >>> m.geo_json(geo_path='geo.json', data=df,
+        >>> m.choropleth(geo_path='geo.json', data=df,
                       columns=['Data 1', 'Data 2'],
                       key_on='feature.properties.myvalue', fill_color='PuBu',
                       threshold_scale=[0, 20, 30, 40, 50, 60])
-        >>> m.geo_json(geo_path='countries.json', topojson='objects.countries')
+        >>> m.choropleth(geo_path='countries.json', topojson='objects.countries')
         """
-        warnings.warn("%s is deprecated. Use %s instead" %
-                      ("geo_json", "add_children(GeoJson)"),
-                      FutureWarning, stacklevel=2)
 
         if threshold_scale and len(threshold_scale) > 6:
             raise ValueError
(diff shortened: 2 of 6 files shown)


folium/features.py
```diff
@@ -184,22 +184,56 @@ def __init__(self, data):
 
                     # providing string
                     GeoJson(open('foo.json').read())
+            style_function: function, default None
+                A function mapping a GeoJson Feature to a style dict.
+                Ex: If you GeoJson has the form
```
ocefpaf: `If your`

folium/features.py
```diff
@@ -184,22 +184,56 @@ def __init__(self, data):
 
                     # providing string
                     GeoJson(open('foo.json').read())
+            style_function: function, default None
```
ocefpaf: I think we should have some defaults and [simplestyle-spec](https://github.com/mapbox/simplestyle-spec/tree/master/1.1.0) is a good candidate.

ocefpaf:
> You have to provide a function `feature -> style` which is to me the most generic API.

I like that as a low level interface: powerful and simple.  We just need to provide some high level methods wrapping that and, as you mention, the choropleth is the first obvious one.

I also think we should comply with the [simple-spec](https://github.com/mapbox/simplestyle-spec/tree/master/1.1.0)  by default.  That would solve #198 and #120.

@BibMartin if I understood your implementation it would be a matter if writing a `style_function` for the simple-spec, correct? We can do that later once your implementation is complete. We just need to be clear in the docs that the `style_function` will expect leaflet defaults, like `fillOpacity` instead of `fill-opacity`.

PS: Another killer PR!  Thanks!!

themiurgo:
I'm against this as it is now, unless you manage to convince me. :grinning:  

In folium the name `GeoJson` has always been confused with `Choroplets`. I think this is a major error and it needs to be fixed soon. Naming this object GeoJsonStyle goes in the opposite direction. Perhaps ChoroplethStyle instead? A hint that this is not just GeoJson is that it has properties like "color domain", "range" and such. GeoJson is not about mapping values to colors.

In the end we want:

- To be able to quickly import / export features as GeoJson.
- That GeoJson follows simple-spec, as @ocefpaf . Let's look at geojson.io for reference. Folium should be the Python version of that, augmented with all the nice plugins it has already and Python c
- To be able to generate choroplet, but that's a different

Let me know what you think about this. :smile:

ocefpaf:
> In folium the name `GeoJson` has always been confused with `Choroplets`. I think this is a major error and it needs to be fixed soon.

Agreed.  And we are working on that here.

> Naming this object GeoJsonStyle goes in the opposite direction. Perhaps ChoroplethStyle instead? 

This PR removes the `GeoJsonStyle` and adds a `kwarg` to style the GeoJSON for any purpose and not only choropleth, so I do not think any renaming is necessary.

> A hint that this is not just GeoJson is that it has properties like "color domain", "range" and such. GeoJson is not about mapping values to colors.

But that **is** styling that goes in the ~~Geo~~JSON when plotting.

> - To be able to quickly import / export features as GeoJson.
> - That GeoJson follows simple-spec, as @ocefpaf . Let's look at http://geojson.io for reference. Folium should be the Python version of that, augmented with all the nice plugins it has already.

And I believe that this implementation, with a default set to comply with the simple-spec, achieve this goal.  What do you think @BibMartin?

> - To be able to generate choroplet, but that should be handled by different methods than the GeoJson import / export.

Why not just another styling function that we can, and should, add another method that call this class and pass a styling function for easy choropleth generation.

Note that we do not need to force anything beyond the simple-spec as the default, but if people have wonky JSONs styles then have the tool to pass them here.

BibMartin:
With this PR, you can:
* Put a GeoJson on the map, with `GeoJson(my_geojson)`.
* If `my_geojson` features have an attribute `properties.style` expressed in **Leaflet style syntax**, it will be taken into account on the map. It can be used for choropleth, but also for various reasons.
* You can overwrite (`update`) the attributes `properties.style` in providing a styling function : `GeoJson(my_geojson, my_style_function)`.
* You can use the old `Map.geo_json` method to draw a choropleth. I'm okay with both of you to rename it `choropleth` ; but I would take the occasion to change it's API.

@ocefpaf 
>  like that as a low level interface: powerful and simple. We just need to provide some high level methods wrapping that and, as you mention, the choropleth is the first obvious one.

True. I imagine we shall have the `GeoJson` object for low-level, and a `Map.choropleth` method for high-level choropleth. I've just kept the `Map.geo_json` name for backward-compatibility, but I'm okay to move it asap to `Map.choropleth`.

> I also think we should comply with the simple-spec by default. That would solve #198 and #120.
> @BibMartin if I understood your implementation it would be a matter if writing a style_function for the simple-spec, correct?

I did not thought about creating a `simple-style` style_function ; that'd be a great simple way of doing it. But honestly, I'm not comfortable with simple-style because
* there are things belonging to mapbox
* it does more than just styling (it creates Icons, popups, etc)
* it creates another syntax, while Leaflet has it's own

If you're okay, I would at least let the simple-style part for another PR.

ocefpaf:
> I did not thought about creating a `simple-style` style_function ; that'd be a great simple way of doing it. But honestly, I'm not comfortable with simple-style because
* there are things belonging to mapbox

Yes, but I believe that the simple-spec is more wide spread than Leaflet's defaults.  I see it a lot in many different mapping clients.

> * it does more than just styling (it creates Icons, popups, etc)

That is desirable sometimes.  Like the choropleth stuff: `range`, `color domain`, etc :wink: 

> * it creates another syntax, while Leaflet has it's own

Maybe that is the real question.  What should be folium's default?  simple-spec or Leaflet's standards?  I am inclined to simple-spec. However, since PR makes it easy to provide styling functions, maybe we can provide the simple-spec function as a built-in option.  Not  sure :expressionless: 

> If you're okay, I would at least let the simple-style part for another PR.

Yes.  I am OK with that.  You already have a rebase on your plate :stuck_out_tongue_winking_eye:

BibMartin:
> > * it does more than just styling (it creates Icons, popups, etc)
> 
> That is desirable sometimes. Like the choropleth stuff: range, color domain, etc :wink: 

I'm okay, but it means that we cannot support the full spec with this object. We need to have a style object that can contain a Icon, Popup and so on. I'll think about how to do it simply for another PR.

As for this PR, do I break the `geo_json` method API immediatly in favor of a `choropleth` ?

ocefpaf:
> I'm okay, but it means that we cannot support the full spec with this object. We need to have a style object that can contain a Icon, Popup and so on.

That is fine.  I though something along the lines of mapping `fillOpacity` &rarr;  `fill-opacity`, etc, while just ignoring `Icons`/`Popups` and other stuff that do not map directly.

Maybe, under that light, the simplespec should not be the default.

We can still implement a simple-spec styling function and provide it as an option.  That function would be the first step to adding a simple-spec GeoJSON.  And we can think about implementing this mapping style function + some Icon/Pop loop that would make a "simple-spec method."  To me that is another :+1: to the robustness of this PR.

BibMartin:
It is (painfully) rebased ; and ready to merge.
Unless you want to take the occasion to replace

    def geo_json(...):
        """

by

    def geo_json(**kwargs):
        warnings.warn('This method is deprecated. Please use Map.choropleth instead.')
        return self.choropleth(**kwargs)

    def choropleth(...):
        """

I can do it before squashing.

folium/features.py
```diff
@@ -169,7 +169,7 @@ def render(self, **kwargs):
 
 
 class GeoJson(MacroElement):
-    def __init__(self, data):
+    def __init__(self, data, style_function=None):
```
ocefpaf: We still need to add a slim version of the simple-spec option.  I am OK doing that in another PR.  Let's just open an issue to keep track.  (Or a new comment on the current issues we have open about the simple-spec.)

folium/features.py
```diff
@@ -193,22 +195,41 @@ def __init__(self, data):
...
+            self.data = json.load(data)
         elif type(data) is dict:
-            self.data = json.dumps(data)
+            self.data = json.loads(json.dumps(data)) # a lazy way of doing deep copy.
         else:
-            self.data = data
+            self.data = json.loads(data)
+
+        if 'features' not in self.data.keys():
+            # Catch case when GEOJson is just a single Feature or a geometry.
```
ocefpaf: GEOJson &rarr; GeoJSON

folium/features.py
```diff
@@ -193,22 +195,41 @@ def __init__(self, data):
...
+        >>> GeoJson(geojson, style_function=style_function)
         """
         super(GeoJson, self).__init__()
         self._name = 'GeoJson'
         if 'read' in dir(data):
-            self.data = data.read()
+            self.data = json.load(data)
         elif type(data) is dict:
-            self.data = json.dumps(data)
+            self.data = json.loads(json.dumps(data)) # a lazy way of doing deep copy.
```
ocefpaf: I like to start phrases with a capital letter and end with a period.  Seems silly, but helps readability later on.

folium/features.py
```diff
@@ -193,22 +195,41 @@ def __init__(self, data):
...
         else:
-            self.data = data
+            self.data = json.loads(data)
+
+        if 'features' not in self.data.keys():
+            # Catch case when GEOJson is just a single Feature or a geometry.
+            if not (isinstance(self.data, dict) and 'geometry' in self.data.keys()):
+                # Catch case when GEOJson is just a geometry.
```
ocefpaf: Same as above.

folium/features.py
```diff
@@ -248,81 +269,6 @@ def render(self, **kwargs):
             JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/topojson/1.6.9/topojson.min.js"),  # noqa
             name='topojson')
 
-
-class GeoJsonStyle(MacroElement):
```
ocefpaf: Glad to see this go away!

folium/folium.py
```diff
@@ -583,20 +578,44 @@ def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
             color_domain = [data_min+i*(data_max-data_min)*1./nb_class
                             for i in range(1+nb_class)]
         else:
-            color_domain = [-1, 1]
+            color_domain = None
```
ocefpaf: :+1:

folium/folium.py
```diff
@@ -583,20 +578,44 @@ def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
...
+        if color_domain and key_on:
+            key_on = key_on[8:] if key_on.startswith('feature.') else key_on
+            color_range = color_brewer(fill_color, n=len(color_domain))
+            get_by_key = lambda obj,key : (obj.get(key,None) if len(key.split('.'))<=1
+                                           else get_by_key(obj.get(key.split('.')[0],None),
+                                                           '.'.join(key.split('.')[1:])))
+            color_scale_fun = lambda x: color_range[len([u for u in color_domain
+                                                         if u<=color_data[get_by_key(x,key_on)]])]
+        else:
+            color_scale_fun = lambda x: fill_color
```
ocefpaf: If my understanding is correct the choropleth would be virtually this function, but renamed and with a few modifications.  Correct?

Since we are killing the `geo_json()` anyways, maybe we can just replace by the new `choropleth()` in the next PR.

folium/folium.py
```diff
@@ -583,20 +578,44 @@ def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
...
+            geo_json = GeoJson(geo_data, style_function=style_function)
 
-        geo_json.add_children(geo_json_style)
+        # Create GeoJsonStyle.
+        #geo_json_style = GeoJsonStyle(
+        #    color_domain, fill_color, color_data=color_data, key_on=key_on,
+        #    weight=line_weight, opacity=line_opacity, color=line_color,
+        #    fill_opacity=fill_opacity)
+
+        #geo_json.add_children(geo_json_style)
```
ocefpaf: Do we need this commented out code?

folium/folium.py
```diff
@@ -583,20 +578,44 @@ def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
...
         self.add_children(geo_json)
-        self.add_children(color_scale)
+        # Create ColorScale
```
ocefpaf: End with a `.`

ocefpaf:
> It is (painfully) rebased

Sorry.  I won't wait so long to review typos and pep8 style.  (And I should not do that when we have many PRs pending.)

> Unless you want to take the occasion to replace

I think we are better off doing that in another PR.

I made a quick review.  Nothing important.

 Let me know if you want to address those here or if you want me to do those later.

@themiurgo are you in peace with this PR or do you still have some doubts?  Your comments are valuable to us!

I am :+1: to merge this PR.

folium/features.py
themiurgo: This should be changed to `data.copy()` to improve readability.

themiurgo:
I'm mainly ok. I'm not 100% ok with the name, but let's include it and see how it work out in the future. :+1:

themiurgo:
> I did not thought about creating a `simple-style` style_function ; that'd be a great simple way of doing it. But honestly, I'm not comfortable with simple-style because
* there are things belonging to mapbox
* it does more than just styling (it creates Icons, popups, etc)
* it creates another syntax, while Leaflet has it's own

I hear your concerns, but at the moment `simplespec-style` is the only standard, the naming used by Leaflet is not a standard, it is tied to their javascript implementation and nobody uses it outside Leaflet. Also, siplespec-style is increasingly used in many places, including GitHub (a non-exaustive list:
https://github.com/mapbox/simplestyle-spec/wiki/Implementations). I think folium needs to be easily extendable and customizable, but have sane defaults. simplespec-style is within the sane default in my opinion.

BibMartin:
Thanks @themiurgo and @ocefpaf for the review.

@ocefpaf 
> Let me know if you want to address those here or if you want me to do those later.

I'll try to apply changes tonight (CET) and squash in the same time.

ocefpaf:
> I'll try to apply changes tonight (CET) and squash in the same time.

No rush.  Just ping me once you are done.

folium/features.py
BibMartin: I'm careful (maybe too much) with `.copy`, provided it's not a deep copy.

For example in this code :

    x = {'a' : {'b' : {'c' : {'d' : 12}}}}
    y = x
    z = x.copy()
    x['a']['b']['c']['d'] = 13

In the end, `x,y,z` have all the same value.

themiurgo: If you have reasons to believe you have nested dicts and you want to do deep copy then just use `copy.deepcopy()`. More Pythonic, no need for comments. :)

By the way, why are you copy the dict? I'd be ok with the map changing after I modify a dict I passed to it. It'd actually surprise me less.

BibMartin:
@themiurgo After our discussion, I removed the line `json.loads(json.dumps(data))` from the code : after all, there's no reason to do a copy. I have been conservative on this point.

@themiurgo @ocefpaf : can you tell me if you're okay with the PR now ; I'll squash it asap.

themiurgo:
Go from my side.

folium/features.py
```diff
@@ -193,22 +195,41 @@ def __init__(self, data):
...
+            self.data = json.loads(data)
+
+        if 'features' not in self.data.keys():
+            # Catch case when GeoJSON is just a single Feature or a geometry.
+            if not (isinstance(self.data, dict) and 'geometry' in self.data.keys()):
+                # Catch case when GeoJSON is just a geometry.
+                self.data = {'type' : 'Feature', 'geometry' : self.data}
+            self.data = {'type' : 'FeatureCollection', 'features' : [self.data]}
+
+        self.style_function = style_function if style_function is not None else (lambda x: {})
```
themiurgo: Three remarks (before @ocefpaf will pep8 this) :smile: 

- the ( ) are redundant around the lambda function, they should go away;
- besides the fact this line is well over 80 lines it's anyways better to make it explicit:
```Python
if not style_function:
    style_function = lambda x: {}
self.style_funciton = style_function
```
- Finally, why did you not use a default a lambda function in the method signature? Is there some Python gotcha I'm not aware of, such as the default [] / {}? Unless there is, it's the cleanest way to do this.

BibMartin:
@themiurgo 
> Three remarks (before @ocefpaf will pep8 this) :smile:


> the ( ) are redundant around the lambda function, they should go away;
besides the fact this line is well over 80 lines it's anyways better to make it explicit:
>        if not style_function:
>            style_function = lambda x: {}
>        self.style_funciton = style_function

Ok, I'll apply the changes

> Finally, why did you not use a default a lambda function in the method signature? Is there some Python gotcha I'm not aware of, such as the default [] / {}? Unless there is, it's the cleanest way to do this.

Just because it makes long lines in the docstring ; you get something like :

    Signature: f(x, foo=<function <lambda> at 0x7fca98079c08>)

ocefpaf:
Awesome work guys! :100: 

@BibMartin can you squash it?
@themiurgo can you merge after that?

Thanks!


> Three remarks (before @ocefpaf will pep8 this) :smile:

My pep8ing, breaking, and scrambling coding skills are everywhere!  No code line is safe :ghost:

themiurgo:
@BibMartin Could you please squash everything, so that I can merge? :)

BibMartin:
> Could you please squash everything, so that I can merge? :)

Done.

> > Three remarks (before @ocefpaf will pep8 this) :smile:

> My pep8ing, breaking, and scrambling coding skills are everywhere! No code line is safe :ghost: 

Yes, that's why challenging PEP8 before merge is a good habit. Maybe one day I'll be able to write PEP8-complianty code directly. Thanks to both of you for review, and thanks @ocefpaf for :ghost:

ocefpaf:
@BibMartin can we do something about the `None` in the colorbar?

folium

folium/folium.py
```diff
@@ -438,7 +438,12 @@ def add_plugin(self, plugin):
 #
 #        self.template_vars.update({'fit_bounds': fit_bounds_str.strip()})
 
-    def geo_json(self, geo_path=None, geo_str=None, data_out='data.json',
+    def geo_json(self, *args, **kwargs):
+        """This method is deprecated, see `Map.choropleth` instead."""
+        warnings.warn('This method is deprecated. Please use Map.choropleth instead.')
```
ocefpaf: Let's add `and will be removed in v0.2.1`

folium/folium.py
```diff
@@ -438,7 +438,12 @@ def add_plugin(self, plugin):
...
+        return self.choropleth(*args, **kwargs)
+
+    def choropleth(self, geo_path=None, geo_str=None, data_out='data.json',
```
ocefpaf: The docstrings need to be updated.

ocefpaf: `m.geo_json` to `m.choropleth`

ocefpaf: We should also remove the deprecation warning.  `Choropleth` will stay.

BibMartin: Ok. What about the deprecation warning in `choropleth` ?

BibMartin: ok.

ocefpaf: I meant the warning in `choropleth`.

ocefpaf:
To deal with the `None` let's just do

```python
        legend_name: string, default empty string
```

Copy-and-paste docstring update too :wink:

BibMartin:
I had it for `None-> ""` but forgot about the docstring

ocefpaf:
Let me update my branch and try again.  I got that `None` yesterday.

PS: do you think you can rebase this to get the `_repr_html()` fix?

BibMartin:
If you're finished, I do an extra commit and wait for your feedback before re-squashing

What _repr_html_ fix ?

ocefpaf:
From #275.  Just because I am lazy and it will make it easier to see the results on the notebook.

ocefpaf:
Just re-run all my notebooks that uses choropleths and they look awesome!

@BibMartin great work!  And sorry for the last minute burst review. Squash?
@themiurgo Now am I done :flushed: I promise no more teasing for a merge and then asking for changes :stuck_out_tongue_winking_eye:

themiurgo:
It'd be great if this could be squashed again. Sorry @BibMartin you can blameit on @ocefpaf :stuck_out_tongue_winking_eye:

ocefpaf:
> Sorry @BibMartin you can blame it on @ocefpaf :stuck_out_tongue_winking_eye:

I take all the blame as long as @BibMartin keeps sending awesome PRs :smile_cat:

BibMartin:
Squashed (sorry for the 2 hours delay ; I was in a meeting)

themiurgo:
Merge! Congrats to @BibMartin and thanks to @ocefpaf for the comments.

ocefpaf:
:tada:

Issue #268, opened 2015-12-01
Control scale on map


stleon:
Hi, I want to add on map control scale, that shows km and miles.

In **js** I use:
```
L.control.scale().addTo(map); 
```

But in **Folium** I can't find any parameter for map ([code](https://github.com/python-visualization/folium/blob/master/folium/map.py#L20)), that can build control scale. 

Pull request needed)

ocefpaf:
Implemented in #355

stleon:
Cool work) thanks

PR #267, opened 2015-12-01
Docs updates (incomplete)

jelmelk:
Just a starting point. Leave feedback/comments/complaints

PR #270, opened 2015-12-02
scale control

jelmelk:
https://github.com/python-visualization/folium/issues/268

Added it to the `Map` class. Shows up on bottom left.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -137,6 +137,8 @@ class Map(MacroElement):
                                            maxBounds: bounds,
                                            layers: []
                                          });
+
+            var scale = L.control.scale().addTo({{this.get_name()}});
         {% endmacro %}
         """)
 


BibMartin:
Wow @jelmelk ; a PR just a few minutes after the issue's report !

I think this would worth being in a separate plugin ; otherwise you force *everyone* to have the control.scale on his map. You can find inspiration in other plugins (like https://github.com/python-visualization/folium/blob/master/folium/plugins/terminator.py).

Provided that you're also working on #265 and #267 ; tell me if you want I do a part of the job :wink:

ocefpaf:
> Wow @jelmelk ; a PR just a few minutes after the issue's report !

Awesome!  Thanks for the PR @jelmelk!

> I think this would worth being in a separate plugin

:+1: 

@jelmelk the key is the [template](https://github.com/python-visualization/folium/blob/master/folium/plugins/terminator.py#L32-L36).  Note that you do not have any extra `JavascriptLink` in this case.

> tell me if you want I do a part of the job

@jelmelk I am sure @BibMartin would do this in quickly, but we (myself included) most learn the new "pluginified way" to do things in folium `v0.2.0`.

So I recommend you take your time and ask for help as much as you need to get this ready on this PR.  It will be worth the effort in the long run :smile:

jelmelk:
> I think this would worth being in a separate plugin ; otherwise you force everyone to have the control.scale on his map. 

It seemed that it's something minor, same as the zoom control. But if it needs to be a separate plugin so be it, if you don't mind tackling that @BibMartin then I'll pay attention on how you do it as a future reference.

>@jelmelk I am sure @BibMartin would do this in quickly

Isn't that the truth :laughing: 

As far as https://github.com/python-visualization/folium/pull/265, not only I ruined the PR by overlapping additional ones it'll take me longer to get something like that done, so it's up to you @BibMartin.

For https://github.com/python-visualization/folium/pull/267 I'm 100% on board.

BibMartin:
I'm in line with @ocefpaf's concern
> but we (myself included) most learn the new "pluginified way" to do things in folium v0.2.0.

But you're doing a great doc job in #267 and I guess this is the main priority before `v0.2.0` release.

I'll finish up the #265 to help, and *leave this PR to you as an exercise* (I've never been teaching, but always dreamt to say that) ; provided there's no hurry on it.

jelmelk:
@BibMartin thank you for your help. I'll take it easy on PR for adding features while I learn the process.

ocefpaf:
> @BibMartin thank you for your help. I'll take it easy on PR for adding features while I learn the process.

:+1:

ocefpaf:
@BibMartin coming back to this I don't think we should add this as a plugin.  In the [`L.control.scale`](http://leafletjs.com/reference.html#control-scale) has no customization options.  It is simple a True/False thing.

Maybe this a case where we could overload the Map() initialization with a new `kwarg` (`control_scale=False/True`).  What do you think?  If so, @jelmelk are you wiling to resurrect this PR, implement the `kwarg`, and add a test for both cases?

ocefpaf:
@BibMartin bringing this back to your radar.  What do you think of my comment above?  If you agree we can add this to `v0.2.0`.

BibMartin:
Ok, see #355

ocefpaf:
Closing in favor of #355.  Thanks @jelmelk!

berrfred:
It looks like L.control.scale() now has several useful options  (e.g. position. metric, imperial), wouldn't it make sense to have it as a separate plugin as initially suggested by @BibMartin ?

I am interested in showing only the metric scale at bottomrigth ... and maybe trick the css to integrate it in the attributions but this is another story.  Today it would require some dirty css and js to achieve it.

berrfred:
This is where I got with my usual script approach, moving control-scale to bottomright before attribution and assuming imperial is the second child, then delete it:

```python
map_script_scale_control = '<script>document.addEventListener("DOMContentLoaded", function() {\
	var leafletControlScale = document.querySelector("div.leaflet-control-scale");\
	document.querySelector("div.leaflet-bottom.leaflet-right").insertBefore(\
		leafletControlScale, document.querySelector("div.leaflet-control-attribution")\
	);\
	leafletControlScale.removeChild(leafletControlScale.children[1]);\
});</script>'

map.get_root().html.add_child(folium.Element(map_script_scale_control))
```

image

Issue #269, opened 2015-12-02
line and multiline deprecation warning


jelmelk:
[`line`](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L129) and [`multiline`](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L175) don't have the usual deprecation warning. 

Can I pull a PR for that?

ocefpaf:
> [`line`](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L129) and [`multiline`](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L175) don't have the usual deprecation warning.

We missed that.

> Can I pull a PR for that?

Please!

PR #271, opened 2015-12-02
deprecation added

jelmelk:
To address this https://github.com/python-visualization/folium/issues/269

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -159,7 +159,9 @@ class Map(_Map):
                     line_color='red', line_opacity=1.0)
 
         """
-
+        warnings.warn("%s is deprecated. Use %s instead" %
+                      ("line", "add_children(PolyLine)"),
+                      FutureWarning, stacklevel=2)
         p = PolyLine(locations,
                      color=line_color,
                      weight=line_weight,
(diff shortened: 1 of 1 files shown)


ocefpaf:
Thanks @jelmelk!

ocefpaf:
@BibMartin unrelated to this PR I just noticed that the `_repr_html_` is broken when we use `line` or `multiline`.  I will update all the notebooks [here](https://github.com/ocefpaf/folium_notebooks) with latest master and report an issue later.

BibMartin:
Great fast job. Congrats @jelmelk 

@ocefpaf 
>  I will update all the notebooks here with latest master and report an issue later.

:+1: btw, I'm thinking we may incorporate a few example notebooks into the tests, to make sure they still run. Why not using something like <a href="https://pypi.python.org/pypi/runipy">this</a> ?

ocefpaf:
I used `runipy` in the past to test notebook output, but the problem, in this case, are that the changes will be hard to make sense (like diffs of base64 encoded PNGs).

I think we can do an OK job by strengthen our tests as their are and increase html comparisons.  I am sure that the `_repr_html_` in that case has a flaw that we can test.

Or maybe (crazy idea alert) we can add runipy, generate HTMLs, publish them, and inspect them before merging a PR.

BibMartin:
Let's continue this discussion in a dedicated issue #272

PR #274, opened 2015-12-02
Fix WMS only layer bug

ocefpaf:
I guess that the `overlay` property is not always present.

@BibMartin tiny one to relax after the #266 rebase :stuck_out_tongue_winking_eye: 

PS: The `LayerControl` seems to be swamping `base_layers` and `overlays`.  I will take a look into that.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -18,7 +18,7 @@ from .map import TileLayer, Icon
 class WmsTileLayer(TileLayer):
     def __init__(self, url, name=None,
                  format=None, layers=None, transparent=True,
-                 attr=None):
+                 attr=None, overlay=True):
         """
         TODO docstring here
 
@@ -29,6 +29,7 @@ class WmsTileLayer(TileLayer):
         self.url = url
         self.format = format
         self.layers = layers
+        self.overlay = overlay
         self.transparent = transparent
         self.attr = attr
 
--- a/folium/map.py
+++ b/folium/map.py
@@ -289,11 +289,11 @@ class LayerControl(MacroElement):
         self.base_layers = OrderedDict(
             [(val.tile_name, val.get_name()) for key, val in
              self._parent._children.items() if
-             isinstance(val, TileLayer) and not val.overlay])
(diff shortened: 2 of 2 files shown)


BibMartin:
The code is ok for me : it can only be better.
But I do you have cases when `overlay` property is not present ?

ocefpaf:
This one:

http://nbviewer.ipython.org/gist/ocefpaf/fd50c22dcf9cf1756235

ocefpaf:
> But I do you have cases when `overlay` property is not present ?

Reading the code closely I think that I am either doing something wrong or the bug is somewhere else.

When I do,

```python
m = folium.Map([40,-100], zoom_start=4)

url = 'http://hfrnet.ucsd.edu/thredds/wms/HFRNet/USEGC/6km/hourly/RTV'

w = folium.WmsTileLayer(url,
                        name='HF Radar',
                        format='image/png',
                        layers='surface_sea_water_velocity',
                        attribution='HFRNet',
                        transparent=True)

w.add_to(m)
m.add_children(folium.LayerControl())
m
```

the WMS layer should be added as an `overlay`, but it is being added as a `base_layer` instead of the underlying `openstreetmap` layer.

overlay


@BibMartin any ideas?

ocefpaf:
I think I am starting to get your latest changes.  (I am slow :grin:)

The latest commit adds a missing `overlay=True` to the `WMS` layer.  Let me know if you think we should get rid of the `hasattr(val, 'overlay')` or if we should leave it here.

PS: The `openstreetmaps` is still being added as an `overlay` instead of the `base_layer`.  I will take a look and fix this in another PR though.

ocefpaf:
@BibMartin Let me know if this is OK now.  I want to take another look at the layers control (making the initial tiles the `base_layer` and disabling the `overlay` by default), but I will send those in another PR to avoid confusion.

BibMartin:
@ocefpaf 
It's okay for me.
Squash and go ?!

ocefpaf:
Done!

PR #275, opened 2015-12-02
Fix #273

### LINKED ISSUE #273: Broken _repr_html_
I mentioned in #271 a problem with the `_repr_html_`.  NB this is not related to `line` and `multiline` as I mentioned in #271.  It is a problem introduce in the recent PRs :unamused: 

@BibMartin This really urges for #272 

See http://nbviewer.ipython.org/gist/ocefpaf/931066a4402320cf872b

ocefpaf:
@BibMartin can you take a quick look?

PS0: We need to test the `_repr_html_` to avoid this

PS1: As a self punishment I will watch [wf-BqAjZb8M](https://www.youtube.com/watch?v=wf-BqAjZb8M) a 100 times :wink:

### DIFF
--- a/folium/element.py
+++ b/folium/element.py
@@ -306,20 +306,21 @@ class Figure(Element):
 
         Parameters
         ----------
+
         """
         html = self.render(**kwargs)
         html = "data:text/html;base64," + base64.b64encode(html.encode('utf8')).decode('utf8')  # noqa
 
         if self.height is None:
-            iframe = ("""
+            iframe = """
             <div style="width:{width};">
             <div style="position:relative;width:100%;height:0;padding-bottom:{ratio};">
             <iframe src="{html}" style="position:absolute;width:100%;height:100%;left:0;top:0;">
             </iframe>
-            </div></div>""")  # noqa
-            iframe.format(html=html,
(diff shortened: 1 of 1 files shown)


jelmelk:
:laughing: that is a great talk tho

BibMartin:
Merged. Thanks @ocefpaf.

BTW : we're using triple quote because we are lazy ; I wonder whether we shall not one day think about moving that kind of code into :

    iframe = (
        '<div style="width:{width};">'
        '<div style="position:relative;width:100%;height:0;padding-bottom:{ratio};">'
        '<iframe src="{html}" style="position:absolute;width:100%;height:100%;left:0;top:0;">'
        '</iframe>'
        '</div></div>')

That way, the outputed HTML would not *benefit* of our python indentations.

But this is a *PEP8-style* PR that shall be launched when the repo is more quiet.

ocefpaf:
> That way, the outputed HTML would not benefit of our python indentations.

The output HTML does need some cleanups.  Tabs, white-spaces, empty lines...

> But this is a PEP8-style PR that shall be launched when the repo is more quiet.

:+1:

PR #276, opened 2015-12-02
Consistent use of the kw attr

ocefpaf:
We have `attr` originally in the `Map()` class.  This PR makes all others `kwargs` `attr` instead of `attribution`.

I am not sure if we should take this opportunity and change everything to `attribution` instead of `attr`.

(At least now it is consistent.)

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -18,7 +18,7 @@ from .map import TileLayer, Icon
 class WmsTileLayer(TileLayer):
     def __init__(self, url, name=None,
                  format=None, layers=None, transparent=True,
-                 attribution=None):
+                 attr=None):
         """
         TODO docstring here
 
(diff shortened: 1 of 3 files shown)


ocefpaf:
@andrewgiessel Can you review and merge this one?  @BibMartin and @themiurgo are you both OK with `attr`?

BibMartin:
> I am not sure if we should take this opportunity and change everything to attribution instead of attr.
> (At least now it is consistent.)

:+1: Let's go like this for the moment ; that's already a good thing

andrewgiessel:
Yes, I will review and do the merge!

ocefpaf:
> Yes, I will review and do the merge!

Thanks @andrewgiessel!  We already have green from @BibMartin.

andrewgiessel:
Looks good to me, merging.

folium/features.py
```diff
@@ -18,7 +18,7 @@
 class WmsTileLayer(TileLayer):
     def __init__(self, url, name=None,
                  format=None, layers=None, transparent=True,
-                 attribution=None):
+                 attr=None):
```
ocefpaf: Did I just break my "consistency" in #348?  I guess I did!

Issue #241, opened 2015-11-10
What are folium natural dependencies ?


BibMartin:
Just to open the question :

today, folium is *officially* depending on `Jinja2`, and the tests require `vincent`, `pandas` and eventually `mock` to run.

In practice, I guess we indirectly require `numpy` and maybe `matplotlib` or `IPython`, because we handle interaction with arrays, colormaps, notebook...

I wonder whether we shall extend the *official* dependency list ; provided that things like `numpy` are installed almost everywhere (?)

ocefpaf:
:+1:  for adding `numpy` (and maybe `matplotlib`+`pandas`) as core dependencies.
:-1: for `vincent` and `IPython` even though the latter is ubiquitous in everybody's python nowadays.

themiurgo:
I would try to keep dependencies to a very minimum, so that people can generate maps even in low power environments. Installing `matplotlib` + `numpy` + `pandas` can be a pain on RPis / smartphones.

As long as one library is only needed for tests/examples, it's not a strong dependency for folium, as one can use it without testing it.

I made an extensive search on these imports and here are the results.
- `pandas`: various examples and documentation files, `test_folium.py`, `utilities.py`;
- `numpy`: examples and docs/docstrings, some functionalities in `utilities.py`;
- `matplotlib`: only examples and docstrings;

In `utilities.py` there is a way already to gracefully degrade in case there is no `pandas` or `numpy`.

```Python
try:
    import pandas as pd
except ImportError:
    pd = None

try:
    import numpy as np
except ImportError:
    np = None
```

So it seems at the moment none of these libraries are strong dependencies.

ocefpaf:
My argument for adding `numpy`, `matplotlib` and `pandas` is to avoid all the workarounds we do to keep them as optional dependencies when most of the users already have them installed.

@andrewgiessel We need more votes.  What do you think? 


PS: 
> can be a pain on RPis / smartphones.

I had both `numpy` and `matplotlib` in my old N900.  Pandas did not even existed at that time!  It is sad to see the support for low power platform is going down with time.

themiurgo:
> My argument for adding numpy, matplotlib and pandas is to avoid all the workarounds we do to keep them as optional dependencies when most of the users already have them installed.

True, but from what I've seen they are all used in areas that are not essential to creating a map. For example, matplotlib, here's what a search returns:

```
➜  folium git:(master) ✗ ack matplotlib
element.py
[...]

PR #280, opened 2015-12-04
Less boilerplate for optional dependencies

ocefpaf:
This PR reduces the amount of code when adding an optional dependency.  I do not think we have anything like this outside `folium/utilities.py`.

I think we should follow this simpler pattern in the future.

### DIFF
--- a/folium/utilities.py
+++ b/folium/utilities.py
@@ -19,16 +19,6 @@ import json
 import base64
 from jinja2 import Environment, PackageLoader
 
-try:
-    import pandas as pd
-except ImportError:
-    pd = None
-
-try:
-    import numpy as np
-except ImportError:
-    np = None
-
 from folium.six import iteritems, text_type, binary_type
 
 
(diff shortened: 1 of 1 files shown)


ocefpaf:
@BibMartin this is not aimed to close #241 per se. But let me know if you are OK with this as a strategy deal with optional dependencies.

BibMartin:
@ocefpaf 
I'm not fully okay with this strategy. I'd prefer if the *buggy* functions were isolated into a plugin.
For example, I guess `ImageOverlay` would need to be a plugin (or at least the `array->image` part), so that the plugins would be tagged **numpy dependent**, while there's no risk of facing a bug when using *classical folium*.

ocefpaf:
> I'm not fully okay with this strategy. I'd prefer if the buggy functions were isolated into a plugin.
For example, I guess ImageOverlay would need to be a plugin (or at least the array->image part), so that the plugins would be tagged numpy dependent, while there's no risk of facing a bug when using classical folium.

That is better.  Still, even as isolated plugins, I'd rather we import inside the function and let the `ImportError` be raised, than suffering the risk of a rogue import somewhere triggering the optimal dependency.

We can close this, leave as-is, and open another issue to isolated things.  Or we can merge this, ship `v0.2.0` and work on isolating things in `v0.2.1`.  Your call.  I am fine either way.

themiurgo:
I'm with @BibMartin for long-term strategy on this.

ocefpaf:
I closed this. We'll follow with @BibMartin's strategy.

Issue #285, opened 2015-12-04
Loading data for the GeoJson


ocefpaf:
The [`GeoJson()`](https://github.com/python-visualization/folium/blob/master/folium/features.py#L228-L234) can be instantiated with a `file-object`, a JSON string, or a JSON-like python dict.

```python
        if 'read' in dir(data):
            self.data = json.load(data)
        elif type(data) is dict:
            self.data = data
        else:
            self.data = json.loads(data)
```

However, we lost the ability to load it by passing a path string and to not embed the data in the HTML. The former can be addressed by adding an extra logic there. The latter needs some thinking:

- To embed or not the data should be an option.
- Few people would want to not embed the JSON data, so that should not be default.
- It should be explicit in the docstring that by embedding is required to properly display standalone maps, like a notebook on nbviewer.

PS: Here is the code I imagine to start addressing this.

```python
if embed:
    if not os.path.isfile(data):
        msg = 'Cannot find file {}'.format
        raise IOError(msg(data))
    self.data = data  # We probably need to change more things downstream.
else:
    if isinstance(data, dict):
        self.data = data
    elif hasattr(f, 'read'):
        self.data = json.load(f)
    elif os.path.isfile(data):
        with open(data) as f:
            self.data = json.load(f)
    else:
        self.data = json.loads(data)
```

BibMartin:
Fixed by #290. Closing this

PR #290, opened 2015-12-05
Add possibility to have unembedded GeoJSON.

BibMartin:
Addresses #285.
Instead of having an `embed` option as @ocefpaf suggested, the user shall pass `data=open(filename)` if he wants to have the file embedded, and `data=filename` otherwise.

Note that the `style_function` cannot be applyed in the latter case, because the file won't be openned.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -228,10 +228,10 @@ class GeoJson(MacroElement):
         """
         super(GeoJson, self).__init__()
         self._name = 'GeoJson'
-        if 'read' in dir(data):
+        if hasattr(data,'read'):
             self.embed = True
             self.data = json.load(data)
-        elif type(data) is dict:
+        elif isinstance(data,dict):
             self.embed = True
             self.data = data
         elif isinstance(data, text_type) or isinstance(data, binary_type):
--- a/folium/features.py
+++ b/folium/features.py
@@ -9,7 +9,8 @@ from jinja2 import Template
 import json
 
 from .utilities import (color_brewer, _parse_size, legend_scaler,
-                        _locations_mirror, _locations_tolist, image_to_url)
+                        _locations_mirror, _locations_tolist, image_to_url,
+                        text_type, binary_type)
 
 from .element import Element, Figure, JavascriptLink, CssLink, MacroElement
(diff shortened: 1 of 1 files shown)


ocefpaf:
> Note that the `style_function` cannot be applied in the latter case, because the file won't be opened.

How difficult would it be to read the style from the file and still not embed?  We do not need to do this in this PR, but I want to keep that idea around.

folium/features.py
```diff
@@ -227,18 +229,20 @@ def __init__(self, data, style_function=None):
         super(GeoJson, self).__init__()
         self._name = 'GeoJson'
         if 'read' in dir(data):
```
ocefpaf: I am not opposed to this pattern: `read in dir(x)`, but I think `hasattr(object, name)` is more common?  Any special reason for doing that?

folium/features.py
```diff
@@ -227,18 +229,20 @@ def __init__(self, data, style_function=None):
...
+            self.embed = True
             self.data = json.load(data)
         elif type(data) is dict:
```
ocefpaf: Same as above.  I see `isinstance(object, dict)` more that `type(object) is dict`.  Again, not opposed, just curious.

BibMartin:
@ocefpaf 
> I am not opposed to this pattern: read in dir(x), but I think hasattr(object, name) is more common? Any special reason for doing that?

I guess this has few importance ; I'm just a goldfish who forgot the clean syntax. Corrected.

> Same as above. I see isinstance(object, dict) more that type(object) is dict. Again, not opposed, just curious.

This one is more important ; for instance, someone may be willing to provide an `OrderedDict` instead of a `dict` ; yet `type(OrderedDict()) is dict` is False while `isinstance(OrderedDict(),dict)` is `True`. So, your (official) syntax is better.

I've corrected both lines.

ocefpaf:
> I guess this has few importance

I just asked for shear curiosity.  Sometimes I learn new things by asking :wink: 

> So, your (official) syntax is better.

I suspected so, but haven't tested it.  Good to know!

Thanks!  Merging this...

Issue #89, opened 2015-03-03
Support for categorical/discrete colour maps in map.geo_json()


psychemedia:
The colour mapping when generating choropleth style maps is based around d3.js colour brewer sequential scales.

When generating thematic maps based on a categorical variable (such as election prediction or outcomes by party), it would be convenient to be able to pass in a colour map as a dict that maps `{'colvar1':color1, 'colvar2':color2, ..., 'colvarN':colorN}`

One option would be to support colorbrewer qualitative scales, though it would also be useful to be able to define color maps for cases where conventional colorings are used (as for political parties, for example).

I can't offhand see how to do that currently?

I note that is a *quantize_fill* parameter in `json_style` within the `geo_json` function. Is that provided as a stub for such a feature?

ocefpaf:
@BibMartin I did not try your hint for using mpl stuff as arg.  Would that fix this issue?

BibMartin:
I guess it's possible. I have to build an example/test.

BibMartin:
Sorry, it's not possible for the moment. The 'hint' trick is for `ImageOverlay` not for `GeoJson`. We definitely have to do something for simplifying choropleth stuff.

ocefpaf:
> We definitely have to do something for simplifying choropleth stuff.

:+1:

BibMartin:
@ocefpaf 
Can we close this based on #266 merge, or do you want to wait for an example in the docs ?

ocefpaf:
> Can we close this based on #266 merge, or do you want to wait for an example in the docs ?

Let's write an example first.  @psychemedia if you beat us to it please share here :wink:

BibMartin:
To keep track of things, we need to review #292 before closing this issue.
@ocefpaf can you assign yourself to this please ? (I cannot review myself)

psychemedia:
@ocefpaf 
Exciting:-) I will try to have a look at this - but it probably won't be before the end of this week.

ocefpaf:
I think #300 covered all that.

PR #292, opened 2015-12-06
Create GeoJSON example notebook

BibMartin:
Adresses #89 and centralize GeoJSON examples.


ocefpaf:
A few points:

- Beautiful!  We need to find a way to auto-generate docs from notebooks like this.
- It is hard to review the notebook on GitHub, so I am merging as-is and review it later.
- We all need to take a look at at to point people to this in issues.  I am sure that notebook will answer most of the questions we will get once we release.

Thanks @BibMartin!

Issue #291, opened 2015-12-05
We have a bug in `mercator_transform`


ocefpaf:
See cell [5]: http://nbviewer.ipython.org/gist/ocefpaf/2c35a679f333c57ecd1d

andrewgiessel:
Not sure what is going on here, because the nbviewer link isn't working well.  When I go to the gist and run it locally, the "bad" cell looks right and the "good" cell looks wrong.  I'll dig into it more

ocefpaf:
@andrewgiessel nbviewer is back online.

The **Bad** (`cell [4]`) you can see the wave north of the US (wrong).

The **Good** (`cell [5]`) shows the bug, but the results should be like those from `cell [6]` (compare with the image on `cell [7]`).

BibMartin:
Can we close this one ?

ocefpaf:
Yep.

PR #293, opened 2015-12-07
Update on ImageOverlay, and example nb

BibMartin:
I hope this addresses #291.

What's new:

* Bugfixes in `ImageOverlay` and it's `utilities`.
* Creates a `Layer` object behind `TileLayer` so that `TileLayer` is not the common parent of all layers (for `LayerControl`)
* Creates an example notebook for `ImageOverlay`. For reviewing this part, I guess it's simpler to go on http://nbviewer.ipython.org/github/bibmartin/folium/blob/image_overlay/examples/ImageOverlay.ipynb

At the cost of doing this, I wonder whether I shall not move `ImageOverlay` to plugins, as discussed in #280.


ocefpaf:
> At the cost of doing this, I wonder whether I shall not move ImageOverlay to plugins, as discussed in #280.

:+1:

BibMartin:
@ocefpaf 
It would require to drop `map.image_overlay` method, and force the users to do

    map.add_children(ImageOverlay(...))

Otherwise, the main `folium` script will depend on `plugins` submodule.
Two questions:

* What shall be done for `v0.2.0` ? Drop, deprecation,nothing ?
* What's the target for later ?

I'd vote for: 
* embedding the `import folium.plugins.ImageOverlay` inside the method `Map.image_overlay`
* add a deprecation warning in the method.
* drom this methods in `v0.2.1`.

@andrewgiessel your opinion would be welcome ; `ImageOverlay` is your baby :wink:

ocefpaf:
I am OK with your strategy.

BibMartin:
Done.

We still have two functions in `utilities.py` that require numpy :

* `split_six` : that needs a rewrite anyway.
* `write_png` : that cannot be moved to plugins, because it's used by several Elements. But it's possible to rewrite it without numpy.

andrewgiessel:
This looks great!  Thanks guys!  In brief, what was the error with the
mercator projection?  you have my :+1

BibMartin:
Squashed.

@andrewgiessel I guess the problem was (and still is) that you cannot draw an image from lat=-90 to lat=+90 ; because the mercator projection square has limits at -85.0513 to 85.0513.

If you replace 

    min_lat=lat.min(), max_lat=lat.max()
by

    min_lat=-85.0513, max_lat=85.0513
then you get the expected graph.

ocefpaf:
I am :+1: to the plugin approach.

I am on the fence about the mercator transform here. My problem with this solution is that, in the previous transform and in the cartopy `warp_array`, we could leave the limits alone and still get the correct result [1].  By forcing the user to change to `min_lat=-85.0513, max_lat=85.0513` when the limits are -90, 90 we add an extra step.  I think we should cover that be covered in the transformation operation in some way.  Maybe clipping the lat to 85.0513 when it goes over?

[1] http://nbviewer.ipython.org/gist/ocefpaf/954e65fa83e116eedd3e

BibMartin:
@ocefpaf 
:+1: 
I guess 72586fe does the job, right ?

folium/plugins/image_overlay.py
```diff
@@ -0,0 +1,134 @@
...
+        If None, the height of the input is used.
+    """
+    import numpy as np
+
+    mercator = lambda x: np.arcsinh(np.tan(x*np.pi/180.))*180./np.pi
+
+    array = np.atleast_3d(data).copy()
+    height, width, nblayers = array.shape
+
+    lat_min = max(lat_bounds[0], -85.051128779806589)
```
ocefpaf: This looks correct to me in this example: http://nbviewer.ipython.org/gist/ocefpaf/4e30b6b3d2426e6294d1

I do note that warping the image, both cartopy and the old function we had, looks "different" (not as smooth as the current version).  Not sure why and I haven't investigate what goes under the cover.

I will try to "beef up" the tests to check this and avoid future regressions.

folium/plugins/image_overlay.py
```diff
@@ -0,0 +1,134 @@
...
+            out[:, i, j] = np.interp(latslats, mercator(lats),  array[:, i, j])
+
+    # Eventually flip the image.
+    if origin == 'upper':
+        out = out[::-1, :, :]
+    return out
+
+class ImageOverlay(Layer):
+    def __init__(self, image, bounds, opacity=1., attr=None,
+                 origin='upper', colormap=None, mercator_project=False,
```
ocefpaf: Maybe we should be more specific and call this [`web_mercator`](https://en.wikipedia.org/wiki/Web_Mercator).

Just for curiosity check [this](https://en.wikipedia.org/wiki/File:Pieni_2_0791.jpg) map out.
I had no idea that there were some [wild truncations](https://en.wikipedia.org/wiki/Mercator_projection#Truncation_and_aspect_ratio ) out there!

folium/plugins/image_overlay.py
```diff
@@ -0,0 +1,134 @@
...
+                overlay=True, control=True):
```
ocefpaf: We need to add `control=True` to the docstrings.  Let's open an issue and address that in another PR to avoid confusion here.  I think that needs to be don in several places.

folium/plugins/image_overlay.py
```diff
@@ -0,0 +1,134 @@
...
+the map, implements ILayer interface.
+
+"""
+import json
+from jinja2 import Template
+
+from folium.utilities import image_to_url
+from folium.map import Layer
+
+def mercator_transform(data, lat_bounds, origin='upper', height_out=None):
```
ocefpaf: Web mercator? (Worth adding the reference to https://en.wikipedia.org/wiki/Web_Mercator in the docstring)

folium/plugins/image_overlay.py
BibMartin: Yes, `Layer` and `control` are something new ; the code needs to re-scanned to apply this.

For example, `Marker`, `LineString` and so on can be considered as `Layer`s.

I've openned issue #295 for this

folium/plugins/image_overlay.py
BibMartin: True. I do it immediatly.

ocefpaf:
@BibMartin I tuned out for a while.  (Sorry... Day job calling.)

Is this PR OK for another round of review?

BibMartin:
@ocefpaf 
> Is this PR OK for another round of review?

Yes, I guess (hope) it's okay.

ocefpaf:
Thanks @BibMartin!

Folium `v0.2.0` is almost ready...

PR #296, opened 2015-12-10
Enable GeoPandas drawing

BibMartin:
Addresses #34.

Note that it only works with `GeoDataFrame`, because `GeoSeries.to_json` method fails (it's inherited from `pandas.Series` and does not link to the feature transformation code).

As for example notebook, the link is here :
http://nbviewer.ipython.org/github/bibmartin/folium/blob/geopandas/examples/GeoPandas.ipynb


folium/features.py
```diff
@@ -241,6 +241,14 @@ def __init__(self, data, style_function=None):
             else:  # This is a filename
                 self.embed = False
                 self.data = data
+        elif hasattr(data, 'to_crs'):
```
ocefpaf: Let's use `isinstance`  I have a few custom object with `to_crs()` defined :wink: 
Also, if we use `geopandas.geodataframe.GeoDataFrame` we can probably drop the next `if` below.

folium/features.py
```diff
@@ -241,6 +241,14 @@ def __init__(self, data, style_function=None):
...
+            # We have a GeoPandas object
+            if hasattr(data, 'columns'):
+                # We have a GeoDataFrame
+                self.embed = True
+                self.data = json.loads(data.to_crs(epsg='4326').to_json())
```
ocefpaf: `GeoDataFrame`s have a `df.crs` that is empty when it it not defined.  We might take advantage of that to do some magic here.  The problem is that we might need `pyproj` for that...

folium/features.py
BibMartin: My goal was to avoid importing geopandas explicitely in folium.
Otherwise, we'll have another dependency.

Maybe there's some other way to filter that.

    'geopandas' in data.__class__().__str__()

is ugly, but may work.

ocefpaf: Don't know what would be the best approach :expressionless: 

Sometimes I think that empowering users by showing how to convert a `GeoDataFrame` to `GeoJSON`, and its pyproj like crs to EPSG would be better than wrapping all that in folium...

ocefpaf:
How about we do something like this:

```python
if/elif(s) and at the else:
    
else:
    from geopandas import GeoDataFrame, GeoSeries
    self.embed = True
    if isinstance(data, GeoDataFrame):
        self.data = json.loads(data.to_crs(epsg='4326').to_json())
    elif isinstance(data, GeoSeries):
        self.data = json.loads(data.__geo_interface__)
    else:
        raise ValueError('Unhandled data type {!r}.'.format(data))
```

This is untested, but I prefer this patter because the `ImportError` has the information we need and the code reads better (IMO :wink:).

If you disagree let's at least put the `{!r}` when raising the ValueError.  That might help people to debug  what object triggered the problem.

BibMartin:
@ocefpaf 
If I try to pass an unappropriate object to `GeoJson`, I shall get

    ValueError('Unhandled data type {!r}.'.format(data))

Yet with your  piece of code, if I have not installed `geopandas`, I will get 

    ImportError

Well, this is a matter of readability ; I'll try to think to it further in the coming days.

ocefpaf:
@BibMartin you are right.  My way can hinder some other failure unrelated to the optional dependency.

However, one might argue that whatever we supported was already tested.  An alternative would be to use a `try/except ImportError` and re-raise it with a new message. "Need optional dependecy `geopandas`.  Could not parse object {!r}."

Anyway, this is not crucial to this PR.  I guess that the main goal is to avoid the need of Geopandas as a core dependency.  If that starts to get in the way we might try to make this external in a different way.  Not sure how yet.  I trust your judgement here.  Just let me know when you are OK with this so I can merge it.

BibMartin:
@ocefpaf 
Thanks to https://github.com/geopandas/geopandas/issues/262, I've realized that `geopandas` objects now have a `__geo_interface__` attribute that does (almost) all the job.
So here is another suggestion to handle the pb...

Please tell me if you're okay with this ; I'll squash commits then.

folium/features.py
```diff
@@ -241,8 +241,19 @@ def __init__(self, data, style_function=None):
             else:  # This is a filename
                 self.embed = False
                 self.data = data
+        elif data.__class__.__name__ in ['GeoDataFrame','GeoSeries']:
+            if hasattr(data, '__geo_interface__'):
+                # We have a GeoPandas 0.2 object
+                self.embed = True
+                self.data = json.loads(json.dumps(data.to_crs(epsg='4326').__geo_interface__))
```
ocefpaf: The `self.embed = True` and `data.to_crs(epsg='4326')` could be up one level to avoid repetition.

PS: `__geo_interface__` is a `dict`, right?  Just making sure that we need  the `json.dumps`.

folium/features.py
```diff
@@ -241,8 +241,19 @@ def __init__(self, data, style_function=None):
...
+            elif hasattr(data, 'columns'):
```
ocefpaf: We really need a new release of GeoPandas!  This is very awkward...

folium/features.py
```diff
@@ -241,8 +241,19 @@ def __init__(self, data, style_function=None):
...
+                # We have a GeoDataFrame 0.1
+                self.embed = True
+                self.data = json.loads(data.to_crs(epsg='4326').to_json())
+            else:
+                raise ValueError('Unable to transform this object to a GeoJSON.')
         else:
-            raise ValueError('Unhandled data type.')
+            raise ValueError('Unhandled data type {!r}.'.format(data))
```
ocefpaf: Let's change the `data type` to `object`, since we are printing the `__repr__` and not the `type`.

folium/features.py
```diff
@@ -241,8 +241,19 @@ def __init__(self, data, style_function=None):
             else:  # This is a filename
                 self.embed = False
                 self.data = data
+        elif data.__class__.__name__ in ['GeoDataFrame','GeoSeries']:
```
ocefpaf: I made my peace with this.  I did not like it at first but now I see it is best.

PS: Can you add a space after the comma?  To avoid pep8 issues later on.

ocefpaf:
Some minor comments.  Will merge as soon as you ping me back.

folium/features.py
BibMartin: > The `self.embed = True` could be up one level to avoid repetition

True. I do it.

>  `data.to_crs(epsg='4326')`  could be up one level to avoid repetition

No, otherwise, the Exception `'Unable to transform this object to a GeoJSON.'` will be replaced by `Unknown method 'to_crs'` (or something like this).

> PS: __geo_interface__ is a dict, right? Just making sure that we need the json.dumps.

`__geo_interface__` is a dict. But we need to do a deep copy of the object here ; otherwise, we'll have unintended side effects : further folium computation (like styling) may modify the ` geopandas`  object...

folium/features.py
BibMartin: ok

folium/features.py
BibMartin: True. The :ghost: is back !!

BibMartin:
@ocefpaf 
Thanks for your feedback.

Answers are in the code.

Squashed and ready for merge if all is ok to you.

folium/features.py
ocefpaf: > No, otherwise, the Exception 'Unable to transform this object to a GeoJSON.' will be replaced by Unknown method 'to_crs' (or something like this).

I was thinking of a copy like:

`a = data.to_crs(epsg='4326')`

 and then

`a.__geo_interface__` or  `a.to_json()`

But that way is fine.

ocefpaf:
Awesome @BibMartin!

I always learn a lot by reviewing your PRs.  Thanks!

folium/features.py
ocefpaf: I need a pep8 Halloween costume.

PR #297, opened 2015-12-10
Add CRS map option

BibMartin:
addresses #8 

See http://nbviewer.ipython.org/github/bibmartin/folium/blob/crs/examples/CRS%20comparison.ipynb


ocefpaf:
Nice!

ocefpaf:
I need to read the docs on the CRS use in leaflet to see if we can add that option to other objects.

Like a GeoJSON that is not on `EPSG3857`, but can be re-project to it on the fly.  (Not sure if that is possible in leaflet though!)

BibMartin:
Seems it's possible with GeoPandas ; but it looks quite buggy for the moment.
At least, it would add dependency to `pyproj` and I guess it's a good reason to let this ouside of folium.

ocefpaf:
Merging this.

Thanks @BibMartin!

Issue #289, opened 2015-12-04
New split_six is not working as expected


ocefpaf:
When using `folium.utilities.split_six` to get the old behavior we are getting a different scale than before:

`split_six`:
![wrong](https://cloud.githubusercontent.com/assets/950575/11601551/b693771e-9ab2-11e5-838e-22efde8e4006.png)

old behavior:
![correct](https://cloud.githubusercontent.com/assets/950575/11601554/b852acbe-9ab2-11e5-9bad-c7099287eab0.png)

Note0: We also want integer number when using `split_six` or any other classification scale.
Note1: We also want to add some of `pysal.esda.mapclassify` so people can go beyond `split_six`.

ocefpaf:
Need to test it against #300 before closing.

ocefpaf:
I am not sure if we can handle the previous behavior in an elegant form.  `split_six` is working properly, but we are constructing the legend in a different way:


Before

```js
var x = d3.scale.linear()
    .domain([0, 9])

var color = d3.scale.threshold()
      .domain([3.0, 7.0, 8.0, 9.0, 9.0])
      .range(['#FFFFCC', '#D9F0A3', '#ADDD8E', '#78C679', '#41AB5D', '#238443']);
```

Now

```js
var x = d3.scale.linear()
    .domain([3, 9])

var color = d3.scale.threshold()
      .domain([3.0, 7.0, 8.0, 9.0, 9.0])
      .range(['#FFFFCC', '#D9F0A3', '#ADDD8E', '#78C679', '#41AB5D', '#238443', '#005A32']);
```

Note that we can just add a zero at the beginning of tge  `threshold_scale` returned from `split_six` to get something similar to the old behavior. But we would still have an extra color entry in the range shifting the colors, and the domain and the ticks will have the zero too.  (Note above that the old behavior does not show the zero in the tick-marks.)

@BibMartin I am OK closing this then.  I don't think it is worth mimicking the old behavior.

PS: when people ask about this we should guide them to [pysal.esda.mapclassify](http://pysal.org/1.2/library/esda/mapclassify.html) :stuck_out_tongue_closed_eyes:

PR #251, opened 2015-11-16
Create crossfilter plugin

BibMartin:
*not ready*

Inspired by @austinlyons 's awesome <a href="https://github.com/austinlyons/dcjs-leaflet-untappd">tutorial</a>, provide an API to create crossfilter maps in folium.

See example at http://nbviewer.ipython.org/github/bibmartin/folium/blob/crossfilter/examples/Crossfilter%20example.ipynb

Issue #195, opened 2015-08-20
Raster/vector image overlay and legend


ocefpaf:
(Summarizing a discussion from #193)

We nee to refactor the `geo_json` (and the legend),  `image_overlay` methods.

~~Possible names: `overlay_raster`, `overlay_vector`, `add_legend`;~~

- [X] "Plugnify" these methods;
- [X] Legend default should be a free scale
- [ ] Legend should have a `quants` with 'auto' or user defined quantiles. (See #255)
- [ ]  See PySAL for many ways to computed this and lets create some examples. (See #255)

ocefpaf:
The these overlay should also have a popup option (see https://github.com/python-visualization/folium/issues/6).

For the GeoJSON case see we need to support external data as popup (see https://github.com/python-visualization/folium/pull/102).

koldunovn:
Are there any plans to actually use data from the GeoJSON file itself (from 'properties'), not only from pandas?

ocefpaf:
Like this: http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/blob/master/test_geojson_popup.ipynb

ocefpaf:
@koldunovn's https://github.com/python-visualization/folium/issues/195#issuecomment-133105426 deserves another issue, but I think we should avoid adding non-standard ((simple-spec) ) geojson features by default.

PR #300, opened 2015-12-14
Colormap

BibMartin:
Create a colormap module, to help generating choropleth and other colouring jobs (like ImageOverlay ?).

It can address #289 (?)

I know there are plenty existing libs doing this (among which matplotlib), but I guess this will be rather simple to maintain and brings no new dependency.

Please see http://nbviewer.ipython.org/github/bibmartin/folium/blob/colormap/examples/Colormaps.ipynb for examples.


ocefpaf:
@BibMartin this looks great.  I will try to review it as soon as possible.  (Unless someone beats me to it.)

BibMartin:
@ocefpaf @themiurgo 
I've realized that colormaps can do much more than what I expected before.

We have 3 commits here:

* ae0bf14 creates `ColorMap` objects
* aafec58 transforms it into a `MacroElement` so that it can replace `ColorScale`
* 373364e expands the method `LinearColorscale.to_step` so that it can replace `split_six`.

I'm not really sure about the third point, but I guess this can help refactoring former `utilities` methods.
You points of view will be more than welcome.

Please try some examples in a notebook:

    import folium.colormap
    import numpy as np
    some_list = np.random.normal(loc=35000, scale=10000, size=300).tolist()

    lc = folium.colormap.linear.YlOrRd
    lc

    lc.to_step(n=12)

    lc.to_step(index=[0,2,4,6,8,10])

    lc.to_step(data=some_list, n=12)

    lc.to_step(data=some_list, n=12, method='linear')

    lc.to_step(data=some_list, n=12, method='log')

    lc.to_step(data=some_list, n=30, method='quantiles')

    lc.to_step(data=some_list, quantiles=[0,0.3,0.7,1])

    lc.to_step(data=some_list, quantiles=[0,0.3,0.7,1], round=True)

    lc.to_step(data=some_list, quantiles=[0,0.3,0.7,1], base_round=True)

ocefpaf:
There is a type in the notebook `hugh` &rarr; `high`

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+Utility module for dealing with colormaps.
+
+"""
+import math
+from jinja2 import Template
+from folium.six import text_type, binary_type
+from folium.element import MacroElement, Figure, JavascriptLink
+from folium.utilities import legend_scaler
+
+_cnames = {
```
ocefpaf: I wonder if someone should create a tiny library with that...

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        else:
+            color_code = _cnames.get(x.lower(),None)
+            if color_code is None:
+                raise ValueError('Unknown color.')
+            color_tuple = (int(color_code[1:3],16), int(color_code[3:5],16), int(color_code[5:7],16))
+    if max(color_tuple)>1.:
+        color_tuple = tuple(u/255. for u in color_tuple)
+    return tuple(map(float,(color_tuple+(1.,))[:4]))
+
+_round = round
```
ocefpaf: If I got this you are keep the built-in function `round` as `_round` to protect it from being clobbered later on, right?
If so can we avoid clobbering it?

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+           'Set2': ['#66c2a5', '#fc8d62', '#8da0cb',
+                    '#e78ac3', '#a6d854', '#ffd92f'],
+           'Set3': ['#8dd3c7', '#ffffb3', '#bebada',
+                    '#fb8072', '#80b1d3', '#fdb462'],
+          }
+
+def _parse_color(x):
+    """
+    """
+    if isinstance(x, tuple) or isinstance(x, list):
```
ocefpaf: This can be `if isinstance(x, (tuple, list))`

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        color_tuple = tuple(x)[:4]
+    elif isinstance(x, text_type) or isinstance(x, binary_type):
```
ocefpaf: Same as above.

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        if x[0]=="#" and len(x)==7:
```
ocefpaf: I think that `.startswith('#')` is more readable.

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+            # Color of the for #RRGGBB
```
ocefpaf: I guess something is missing the the comment above.

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+            color_tuple = (int(x[1:3],16), int(x[3:5],16), int(x[5:7],16))
```
ocefpaf: :ghost: space after comma :ghost:

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        else:
+            color_code = _cnames.get(x.lower(),None)
+            if color_code is None:
+                raise ValueError('Unknown color.')
```
ocefpaf: It always help people to debug if we print the `{!r}` of the offending object.  How about:
`'Unknown color {!r}.'.format(x)`

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+            color_tuple = (int(color_code[1:3],16), int(color_code[3:5],16), int(color_code[5:7],16))
```
ocefpaf: :ghost:

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        self.caption = caption
+        self.index = [min, max]
+
+        self._template = self._env.get_template('color_scale.js')
+
+    def render(self, **kwargs):
+        self.color_domain = [self.min + (self.max-self.min)*i/499. for i in range(500)]
+        self.color_range = [self.__call__(x) for x in self.color_domain]
+        self.tick_labels = legend_scaler(self.index)
+        self.caption = ""
```
ocefpaf: Is this intended?

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+class ColorMap(MacroElement):
+    """A generic class for creating colormaps."""
+
+    def __init__(self, min=0., max=1., caption=""):
+        """
+        """
+        super(ColorMap, self).__init__()
+        self._name = 'ColorMap'
+
+        self.min = min
```
ocefpaf: How about `vmin`, `vmax`?

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+
+        super(ColorMap, self).render(**kwargs)
+
+        figure = self.get_root()
+        assert isinstance(figure, Figure), ("You cannot render this Element "
+                                            "if it's not in a Figure.")
+
+        figure.header.add_children(
+            JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js"),  # noqa
+            name='d3')
```
ocefpaf: if you are using `#noqa` you can make this 1-line.  It those moments where pep8 hurts readability... :arrow_down_small: :ghost: (Could not find an upside down ghost emoji.)

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+
+    def _rgba_floats_tuple(self, x):
```
ocefpaf: I imagine this will go away and the code will be inside `rgba_floats_tuple()`, right?

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+        """
+        return tuple(int(u*255.9999) for u in self.rgba_floats_tuple(x))
+
+    def rgb_bytes_tuple(self, x):
+        """Provides the color corresponding to value `x` in the
+        form of a tuple (R,G,B) with int values between 0 and 255.
+        """
+        return self.rgba_bytes_tuple(x)[:3]
+
+    def rgb_hex_str(self, x):
```
ocefpaf: When I see all these methods I cannot stop thinking about a tiny lib for colormaps... I've seen implementations like this in so many places now...

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+                        i=i*1,
+                        color = self.rgb_hex_str(self.min + (self.max-self.min)*i/499.))
+                    for i in range(500)
+                    ])
+            + '<text x="0" y="35">{}</text>'.format(self.min)
+            + '<text x="500" y="35" style="text-anchor:end;">{}</text>'.format(self.max)
+            + '</svg>')
+
+class LinearColormap(ColorMap):
+    def __init__(self, colors, index=None, min=0., max=1., caption=""):
```
ocefpaf: Again let's consider changing all `min`/`max` to `vmin`/`vmax` to avoid name clashing.

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+            The minimal value for the colormap.
+            Values lower than `min` will be bound directly to `colors[0]`.
+        max : float, default 1.
+            The maximal value for the colormap.
+            Values higher than `max` will be bound directly to `colors[-1]`.
+        """
+        super(LinearColormap, self).__init__(min=min, max=max, caption=caption)
+
+        n = len(colors)
+        if n<2:
```
ocefpaf: :ghost:

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+            self.index = list(index).copy()
+        self.colors = [_parse_color(x) for x in colors]
+
+    def _rgba_floats_tuple(self, x):
+        """Provides the color corresponding to value `x` in the
+        form of a tuple (R,G,B,A) with float values between 0. and 1.
+        """
+        if x<= self.index[0]:
+            return self.colors[0]
+        if x>= self.index[-1]:
```
ocefpaf: :ghost: (also above)

folium/colormap.py
```diff
@@ -0,0 +1,594 @@
...
+
+class _LinearColormaps(object):
+    """A class for hosting the list of built-in linear colormaps.
+    """
+    def __init__(self):
+        self._schemes = _schemes.copy()
+        self._colormaps = {key : LinearColormap(val) for key,val in _schemes.items()}
+        for key,val in _schemes.items():
+            setattr(self,key, LinearColormap(val))
+    def _repr_html_(self):
```
ocefpaf: space

ocefpaf:
@BibMartin I made some minor comments.  I'd really appreciate if someone else could take a look too.

Even though I really like this PR a lot I cannot shake the feeling that this is too much for `folium`.  Maybe it is unavoidable...  let's see what others have to say.

BibMartin:
@ocefpaf 
> I wonder if someone should create a tiny library with that...

I was thinking of it when programming the first commit ae0bf14 ; in fact in this commit the module `colormap.py` is almost independent of folium (only imports `text_type` and `binary_type`).

It could be another (tiny) library supported by `python-visualization` community.

But on the other hand:

* It means more docs, releases, etc
* It would be another dependency for folium (with risks of inconsistencies)
* The other commits (aafec58 and 373364e) take advantage of having this module *inside* and have a colormap object that is also the colorscale `Element`.

Finally, I'm quite open to discuss that point, and would also appreciate other inputs (ping @themiurgo, @andrewgiessel  ?)

> If I got this you are keep the built-in function round as _round to protect it from being clobbered later on, right?
> If so can we avoid clobbering it?
> Again let's consider changing all min/max to vmin/vmax to avoid name clashing.

I like having simple and explicit kwarg names ; but I will trust your wise opinion.

As for your other comments, I'm okay and will apply changes asap.

ocefpaf:
> But on the other hand ...

I get that.  That is why I said someone else. You **not** included :wink:

> I like having simple and explicit kwarg names ; but I will trust your wise opinion.

Wise? No so sure :smile:

Note that I am biased here.  I guess that matplotlib uses `vmin`/`vmax`, right?

themiurgo:
I agree with the wise @ocefpaf on this. This functionality is great, but it feels a bit out of place here. Why don't you make a small library out of it?

Something that *seems* to be missing from scientific python community is a d3-like python library. There is d3py but the project looks abandoned. This library would be a good building block for something like that, which converts values to other scales or to colors.

ocefpaf:
> I agree with the wise @ocefpaf on this. This functionality is great, but it feels a bit out of place here. Why don't you make a small library out of it?

Just to be clear I am OK merging it.  I would like to see this code as external library, but the maintenance burden would be too much on @BibMartin shoulders.  (And I selfish!  I want to keep him focused on folium :stuck_out_tongue_closed_eyes:)

> Something that *seems* to be missing from scientific python community is a d3-like python library.

Indeed!

> There is d3py but the project looks abandoned. This library would be a good building block for something like that, which converts values to other scales or to colors.

I guess we should make some twitter noise: "folium submodule for sale!"

themiurgo:
Let's merge it then!

BibMartin:
> Something that seems to be missing from scientific python community is a d3-like python library

Yes, but it cannot be restricted to d3 ; d3 is often a brick among others to create a nice report (another example could be #251). This would be a *python-visualization* library based on HTML and JS.

I'm split between 3 scenarios:

#### 1. Folium is *python-visualization*

As a folium user and contributor, I'm glad not needing to install *fancy* dependencies with folium. By *fancy*, I mean something that risk to be buggy, not to evolve anymore or evolve too fast etc.
I guess it's painful to have to wait for another library to move before fixing a bug/enhancement.
Plus the complexity of tracking which versions of which lib are compatible...

In a word, I would keep everything in folium as much as possible ; at the cost of doing a big split when mature.

#### 2. Folium is the map plugin of *python-visualization*

Folium's goal is to enable easy map generation with Leaflet. We could factor-out everything that's not map-specific: `Element`, `Figure`, `Link`, `Div`, `VegaDiv` etc, and `Colormap`.

Finally, *python-visualization* would implement the core of the mindset (objects added the one to the other, render methods outputing result in a file or an iframe). And folium would focus on providing the leaflet-specific objects : `Map`, `Marker`, etc.

Other libraries could be created based on *python-visualization* (GeoJson, Crossfilter, NVD3, bootstrap...) but without involving folium.

#### 3. Folium depends on an existing visualization library

There are several visualization libraies available in python community (matplotlib, mpld3, bokeh, plot.ly...). No one is perfect today, nor dominating (except for matplotlib ; but not really notebook-interactive), but I'm not sure it worth trying to create another competitor. We can decide to let folium depend on one of these and use their colormap function, ...

Anyway, I think these reflections are useful (and necessary), but they are middle or long-term thoughts. I would vote for merging colormap as-is today and finish `v0.2` and tackle the refactoring in `v0.3`.


After that piece of (silly?) text, I looking forward for your reactions. :anguished:

andrewgiessel:
Sorry I didn't get a chance to look at this over the week- these larger issues deserve some thought.  I think colormap is a clear case of something that can be (and should be) either refactored out to another python-visualization project or use an existing library.

:+1 to merge for now, and include in v0.2

themiurgo:
> Yes, but it cannot be restricted to d3 ; d3 is often a brick among others to create a nice report (another example could be #251). This would be a python-visualization library based on HTML and JS.

I might have been unclear. What I meant by d3-like library is "something that converts values and ranges into geometrical objects". It might not be at all related to JS, HTML and it might use matplotlib, folium or anything as a backend.

ocefpaf:
@BibMartin I see a future where we are 90% (2) + 10% (3).  Folium is not python-visualization.  I guess that, when @wrobstory created the organization, he wanted to add his other projects here too (vincent, bearcart, etc).

> We could factor-out everything that's not map-specific: `Element`, `Figure`, `Link`, `Div`, `VegaDiv` etc, and `Colormap`.

Maybe yes.  Definitely not now!  Bare in mind that, if we re-factor, then folium would depend on those and we would start entering the 10% of (3) I mentioned above.

PS: Bokeh is improving their map support.  However, I believe folium is very different from bokeh by providing easy maps (just a few lines of code) and close integration with Leaflet.

BibMartin:
Thank you @ocefpaf @themiurgo @andrewgiessel for your comments.
I've created issue #320 to keep track of this discussion and think of factoring `colormap` (at least) out of folium.

For now it seems we are close to <a href="https://github.com/python-visualization/folium/issues?q=is%3Aopen+is%3Aissue+milestone%3Av0.2.0">`v0.2` release</a>...
Can we close #289, #195 and #89 based on this PR ?

wrobstory:
> Folium is not python-visualization. I guess that, when @wrobstory created the organization, he wanted to add his other projects here too (vincent, bearcart, etc).

Hi everybody!

My biggest goal was largely to move Folium _out_ of the `github.com/wrobstory` namespace when ownership was handed over to y'all. I had anticipated moving other projects into the `python-visualization` organization as well, but (obviously) have not done so. 

One option is to rename the organization from `python-visualization` to `folium`, and then sub-libraries that Folium depends on could be added to that organization. I'm totally fine with doing that- the only pain is updating your local remote pointers for the new name.

Issue #225, opened 2015-10-31
Improve plugin tests


ocefpaf:
We can do a better in doing the same thing as `test_folium.py`: compare the output of `.render()` with code generated from a validation template.

BibMartin:
Just a question of organisation : where should be stored each plugin components (code, test code, template, eventual data files) ?

An approach would be to host everything in the plugin folder with something like this:

    folium/plugins
        ./image_overlay
            ./__init__.py
            ./somecode.py
            ./test.py
            ./image_overlay.template
            ./some_data.json

then we would just have to import all the tests in `folium/tests/test_plugins.py`

ocefpaf:
I am not OK with that pattern.  Mainly because it is uncommon and not because I see a problem right away.  I'd rather have the code for the tests inside a `test/plugin` folder.  We can do a file for each plugin.

BibMartin:
ok. To sum up :

* The code is in `folium/plugins/image_overlay.py` (or a submodule)
* The test code is in `tests/plugins/test_image_overlay.py`
* The test data are in `tests/plugins`
* The test templates are inline, or in `tests/plugins`

ocefpaf:
I like that.  We can put the test data and templates in `tests/plugins/{data,templates}` if you think it will be better organized that way.  Ideally we could try to make the data and template light enough to be inside the `test_{plugin}.py` file to make each test file self-contained.

BibMartin:
pre {display: block;font-family: Helvetica Neue, Helvetica, Arial, sans-serif;white-space: pre-wrap;margin: 1em 0;}> we could try to make the data and template light enough to be inside thetest_{plugin}.py file to make each test file self-contained.

:+1:

Envoyé depuis AndroidFilipe --- [Marketing Mail] Re: [folium] Improve plugin tests (#225) --- De :"Filipe" <notifications@github.com>A :"python-visualization/folium" <folium@noreply.github.com>Cc :"Martin Journois" <martin.journois@fr.michelin-solutions.com>Date :mer., 16 déc. 2015 14:55Objet :[Marketing Mail] Re: [folium] Improve plugin tests (#225)I like that.  We can put the test data and templates in tests/plugins/{data,templates} if you think it will be better organized that way.  Ideally we could try to make the data and template light enough to be inside the test_{plugin}.py file to make each test file self-contained.

BibMartin:
See #304

BibMartin:
@ocefpaf
Can we close this ?

[...]

PR #304, opened 2015-12-20
Moved plugins test to tests/plugins

BibMartin:
Not finished. Adresses #225 .


BibMartin:
There remains 3 tests to improve. @ocefpaf you can already tell me whether the mindset is good.

BibMartin:
Finished.
@ocefpaf if you can review this PR first ; it's moving several files and I'd like to avoid interactions with other PR as much as possible.

ocefpaf:
This one looks awesome!  Thanks @BibMartin!!

folium/plugins/boat_marker.py
```diff
@@ -53,8 +53,7 @@ def __init__(self, position=None, heading=0,
             {% endmacro %}
             """)  # noqa
 
-
-def render(self, **kwargs):
+    def render(self, **kwargs):
```
ocefpaf: I guess that the tests found this bug :stuck_out_tongue_winking_eye:

tests/plugins/test_image_overlay.py
```diff
@@ -0,0 +1,37 @@
...
+import folium
+from folium import plugins
+
+def test_image_overlay():
+    """Test image overlay."""
+    data = [[[1, 0, 0, 1], [0, 0, 0, 0], [0, 0, 0, 0]],
+            [[1, 1, 0, 0.5], [0, 0, 1, 1], [0, 0, 1, 1]]]
+
+    m = folium.Map()
+    io = plugins.ImageOverlay(data,[[0,-180],[90,180]], mercator_project=True)
```
ocefpaf: Spaces...

tests/plugins/test_terminator.py
```diff
@@ -0,0 +1,22 @@
...
+---------------
+"""
+
+from jinja2 import Template
+
+import folium
+from folium import plugins
+
+def test_terminator():
+    m = folium.Map([45., 3.], zoom_start=1)
```
ocefpaf: Virtually all tests start by importing, `jinja2.Template`, `folium`, and `folium.plugins` and creating a map.  What do you think of an external `util.py` with this a simple function to do that?

ocefpaf:
I made a few minor comments.  Ping me back when you are done.  (Thanks again.  We really needed this!)

ocefpaf:
BTW.  I am OK merging this as is if you are busy.  I can make the modifications later.

tests/plugins/test_terminator.py
BibMartin: I would be okay, but I would keep it for another PR (maybe `v0.3`).

At the cost of doing this, I'd like to have an object like `PluginTester` so that we do something like :

    class TestTerminator(PluginTester):
        def test_main(self):
            ...

but `PluginTester` would also have a method we could use to generate all the test examples for the doc. I think it's too bad to implement nice maps in the test and have to re-implement it in notebooks for docs and examples.

ocefpaf: Sounds like a plan.  Can I merge it then?

BibMartin: Appart from this, I've applied you comments (spaces), and rebased.
Ready for merge.
Thanks @ocefpaf

ocefpaf:
Thanks @BibMartin!

Issue #139, opened 2015-05-26
Fit bounds to specific Features or FeatureGroups.


themiurgo:
As suggested by @ocefpaf in #134 , it would be nice to fit bounds to a group of features. Implementation should follow closely Leaflet, in terms of fitting bounds to a specific `Feature` or `FeatureGroup`/`Layer`.

ocefpaf:
See #305

Issue #244, opened 2015-11-13
Autobounds not working


themiurgo:
Since 95a8c0f281f2e28c135288ca0a8d5c48eb9819c0 automatic bounds are not working (the method `_auto_bounds` was commented out and never reintroduced).

ocefpaf:
Reviewer :gun: 

Oh no wait!! I was the Reviewer!!!

themiurgo:
:v:

themiurgo:
@BibMartin Any particilar reason for this or you just forgot? :)

BibMartin:
No reason for this ; I hope there's no other *collateral effect* of the refactoring.
@themiurgo thanks for having noticed this.

themiurgo:
It's alright, if not I'll just reimplement it or fix it. Just wondering if I had to be aware of any particular reason why this might not work. :smile:

ocefpaf:
See #305

PR #305, opened 2015-12-22
Bounds

BibMartin:
Not finished. Adresses #139

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -126,7 +126,6 @@ class RegularPolygonMarker(Marker):
             JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.2/leaflet-dvf.markers.min.js"),  # noqa
             name='dvf_js')
 
-
 class Vega(Element):
     def __init__(self, data, width=None, height=None,
                  left="0%", top="0%", position='relative'):
@@ -277,6 +276,12 @@ class GeoJson(MacroElement):
                 self.style_function(feature))
         return json.dumps(self.data)
 
+    def _get_self_bounds(self):
+        """Computes the bounds of the object itself (not including it's children)
+        in the form [[lat_min, lon_min], [lat_max, lon_max]]
+        """
+        raise NotImplementedError
+
 class TopoJson(MacroElement):
     def __init__(self, data, object_path):
         """
@@ -315,6 +320,13 @@ class TopoJson(MacroElement):
             JavascriptLink("https://cdnjs.cloudflare.com/ajax/libs/topojson/1.6.9/topojson.min.js"),  # noqa
             name='topojson')
 
+    def _get_self_bounds(self):
+        """Computes the bounds of the object itself (not including it's children)
+        in the form [[lat_min, lon_min], [lat_max, lon_max]]
+        """
+        raise NotImplementedError
+        return [[self.location[0],self.location[1]],[self.location[0],self.location[1]]]
+
 class ColorScale(MacroElement):
     def __init__(self, color_domain, color_code, caption=""):
         """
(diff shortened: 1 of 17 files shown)


folium/utilities.py
```diff
@@ -428,3 +428,37 @@ def _locations_tolist(x):
...
+                out = []
+                for y in x:
+                    out += iter_points(y)
+                return out
+            else:
+                return [x]
+        else:
+            return []
+    else:
+        raise ValueError('List type expected {!r}'.format(x))
```
ocefpaf: Maybe it is better: `List type expected.  Got {!r}.`

folium/utilities.py
```diff
@@ -428,3 +428,37 @@ def _locations_tolist(x):
...
+    else:
+        return max(x,y)
+
+def iter_points(x):
+    """Iterates over a list representing a feature, and returns a list of points,
+    whatever the shape of the array (Point, MultiPolyline, etc).
+    """
+    if isinstance(x,list):
+        if len(x):
+            if isinstance(x[0],list):
```
ocefpaf: space after the `,` :ghost:

folium/map.py
```diff
@@ -421,6 +417,11 @@ def __init__(self, location, popup=None, icon=None):
             {% endmacro %}
             """)
 
+    def _get_self_bounds(self):
+        """Computes the bounds of the object itself (not including it's children)
+        in the form [[lat_min, lon_min], [lat_max, lon_max]]
+        """
+        return [[self.location[0],self.location[1]],[self.location[0],self.location[1]]]
```
ocefpaf: Spaces after the `,`

folium/features.py
```diff
@@ -616,6 +669,23 @@ def __init__(self, locations, color=None, weight=None,
                 {{this._parent.get_name()}}.addLayer({{this.get_name()}});
             {% endmacro %}
             """)  # noqa
+    def _get_self_bounds(self):
+        """Computes the bounds of the object itself (not including it's children)
+        in the form [[lat_min, lon_min], [lat_max, lon_max]]
+        """
+        bounds = [[None,None],[None,None]]
```
ocefpaf: Spaces...

folium/features.py
```diff
@@ -616,6 +669,23 @@ def __init__(self, locations, color=None, weight=None,
...
+        for point in iter_points(self.data):
+            bounds = [
+                [
+                    none_min(bounds[0][0], point[0]),
```
ocefpaf: I see what you are trying to do here (and I do not have a better suggestion right now), but it looks like we could do something simpler...

ocefpaf:
Thanks @BibMartin! I made a few minor comments.  Let me know once you are done so I can review it again.

folium/utilities.py
BibMartin: ok, I'll do it

BibMartin:
Thanks @ocefpaf for first review. I'm okay with your comments and will apply them asap.
Before merging, I would like to:

* add `get_bounds` call in most tests (see `test_topo_json` for example)
* plug back something *like* `Map._auto_bounds` that uses this method.

BibMartin:
@ocefpaf 
I'm finished with this one. After rethinking, I guess it's no need to re-create the `_auto_bounds` method, provided everyone can create his own bounds in doing :

    map.fit_bounds(my_feature.get_bounds())

We can eventually add something in `Map.render` so that if no `FitBounds` has been defined, we do

    map.fit_bounds(map.get_bounds())

Tell me if you want I add it before squashing.

BibMartin:
@themiurgo I would be interested to have your point of view also ; I realize this addresses #244 in the same time as #139 .

ocefpaf:
Awesome!  

http://nbviewer.ipython.org/github/ocefpaf/folium_notebooks/blob/master/test_fit_bounds.ipynb

Thanks @BibMartin!!!

themiurgo:
Sorry, coming late for this. Looks great though! :)

PR #308, opened 2015-12-24
Fix #307  dependency problem when installing from pip

### LINKED ISSUE #307: pip installing from pypi doesn't install jinja
Hi

    ╰─$ pip install folium
    Collecting folium
    Installing collected packages: folium
    Running setup.py install for folium
    Successfully installed folium-0.1.6

    ╰─$ python -c 'import folium'
    Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/home/zalew/.virtualenvs/playground3/local/lib/python2.7/site-packages/folium/__init__.py", line 6, in <module>
    from folium.folium import Map
    File "/home/zalew/.virtualenvs/playground3/local/lib/python2.7/site-packages/folium/folium.py", line 19, in <module>
    from jinja2 import Environment, PackageLoader
    ImportError: No module named jinja2

The problem lies here:

    64: with open('requirements.txt') as f:
    65:     tests_require = f.readlines()
    66: install_requires = [t.strip() for t in tests_require]

Submitting fix.

Thank you for your work on this application!

zalew:


BibMartin:
Hi @zalew 
Thanks for the PR.

I'm sorry I know very few about how pip works ; I guess @ocefpaf will have to help on this (please).

But I've tried to undestand what the PR brings, and in both cases (before and after) the variable `install_requires` is equal to `['Jinja2']` in the end. Is there some specific version of python that led you to need this change ?

ocefpaf:
This PR is not necessary.  The issue does exist on `v01.6` source dist at PyPI, but we fixed that a long time ago.  (The only problem we do have is the bad naming `tests_require` :wink:)

Issue #316, opened 2016-01-04
Data not binding to choropleth in latest folium 0.2.0.dev (but worked in 0.1.6)


alexandercbooth:
Hi @ocefpaf & @BibMartin et al., 

First, thank you all for making such a wonderful tool!

I recently updated from the folium version currently on pip (0.1.6) to the current master (0.2.0.dev). A choropleth bound with data in 0.1.6 is no longer rendering in 0.2.0.dev. My code and images are displayed below. 

# In folium 0.1.6
```
zip_geo = r'newzip4.json'
m = folium.Map(location=[45,-72], zoom_start=4, tiles='cartodbpositron')
m.geo_json(geo_path=zip_geo, data_out='data4.json', data=df,
             columns=['ZIP3', 'count'],
             key_on='feature.properties.ZIP3',
             fill_color='YlGnBu', fill_opacity=0.95, line_opacity=0.9, 
             legend_name='Count by Location', topojson='objects.newzip3_4326', threshold_scale = [1,2500,5000,7500, 12000])
m
```

### This produces the following map:

![screen shot 2016-01-04 at 12 20 16 pm](https://cloud.githubusercontent.com/assets/15242567/12095590/9f71df74-b2dd-11e5-8594-dbbeacc0c20a.png)

# In folium 0.2.0.dev
```
zip_geo = r'newzip4.json'
m = folium.Map(location=[45,-72], zoom_start=4, tiles='cartodbpositron')
m.choropleth(geo_path=zip_geo, data_out='data4.json', data=df,
             columns=['ZIP3', 'count'],
             key_on='feature.properties.ZIP3',
             fill_color='YlGnBu', fill_opacity=0.95, line_opacity=0.9, 
             legend_name='Count by Location', topojson='objects.newzip3_4326', threshold_scale = [1,2500,5000,7500, 12000])
m
```

### This produces the same map, but without the data bound to it:

![screen shot 2016-01-04 at 12 22 25 pm](https://cloud.githubusercontent.com/assets/15242567/12095623/e048e358-b2dd-11e5-97d9-5ce183121696.png)
 
I apologize if this is just due to my misunderstanding of binding data to choropleths in the latest version. Any help would be greatly appreciated. 

Thanks so much!

Alex

ocefpaf:
> I apologize if this is just due to my misunderstanding of binding data to choropleths in the latest version. 

It seems OK to me.  Can you share your data so we can investigate this further?

alexandercbooth:
[...]

PR #317, opened 2016-01-04
Add style_function to TopoJson

BibMartin:
addresses #316 

Not finished ; I guess tests will need an update and an example could be good.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -308,7 +308,7 @@ class GeoJson(MacroElement):
         return bounds
 
 class TopoJson(MacroElement):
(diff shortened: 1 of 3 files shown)


ocefpaf:
> Not finished ; I guess tests will need an update

I guess you updated them, right?  Or do you want to add something extra?

>  and an example could be good.

That can be added later if you think this is ready to be merged.

ocefpaf:
BTW here is @alexandercbooth example using this PR:

http://nbviewer.ipython.org/gist/ocefpaf/ae22bd2ee2d4f86abffd

PR #321, opened 2016-01-10
Fix colormap import error in python2

BibMartin:
When trying to play with colormap in python2 ; I got bad surprises.

This is to fix this.

BTW ; this shows that we need tests for colormap. I'll add them in this PR soon.


ocefpaf:
> BTW ; this shows that we need tests for colormap. I'll add them in this PR soon.

And a better reviewer.  That @ocefpaf guy did not test this right...

PR #322, opened 2016-01-11
Add test_colormap.py

BibMartin:
To avoid #321 happen again.


ocefpaf:
:+1: (I see some pep8 violation, but I will merge with them to test my next PR.)

PR #326, opened 2016-01-14
Add period in plugins.timestamped_geojson

### LINKED ISSUE #325: Add the period option to TimestampedGeoJson (and/or fix the timestamps?)
Even when providing an array of times (in milliseconds) it seems that TimestampedGeoJson is defaulting to show the data daily. (I believe that the default is `period: "P1D"`).

I thought the `period` would be used only if `times` was `null`.  So maybe this is the sign of a deeper problem.

BibMartin:
Fix #325.

I had it and just forgot to do the PR.

### DIFF
--- a/folium/plugins/timestamped_geo_json.py
+++ b/folium/plugins/timestamped_geo_json.py
@@ -26,7 +26,8 @@ from folium.element import MacroElement, Figure, JavascriptLink, CssLink
 from folium.utilities import none_min, none_max, iter_points
 
 class TimestampedGeoJson(MacroElement):
-    def __init__(self, data, transition_time=200, loop=True, auto_play=True):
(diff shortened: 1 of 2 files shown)


ocefpaf:
You are fast!  I was still reading `Leaflet.TimeDimension` docs :smile_cat:

BibMartin:
@ocefpaf 
Squashed and ready to go if you're okay.

ocefpaf:
Awesome!  BTW here is how I found the bug:

http://nbviewer.ipython.org/gist/ocefpaf/60ae6dd499381b107dcc?flush_cache=true

My first attempt to use `timestamped_geojson`.  (It does nor work on Chrome BTW.  I tested it only on Chrome and Firefox.)

Issue #249, opened 2015-11-15
Masking map areas without overlay


methodds:
The possibility to use JSON data for map overlays is easily the best feature of folium. For many visualizations, non-overlay parts of a map are not important and disturb viewers from focusing on the main information.  Masking map areas (e.g. to white, black, or transparent) without overlay would be a very nice addition to folium, so that in combination with #210, one can create maps like this: 

![alt image](http://ichef.bbci.co.uk/news/660/cpsprodpb/F429/production/_85450526_asylum_seekers_germany_624v.png)

themiurgo:
Great suggestion! :+1: 

Would you like to contribute this enhancement by writing it yourself? It would mean a lot to the rest of the folium users. :smile:

methodds:
I would love to contribute to folium but unfortunately have no free time left in 2015 :disappointed: . If this enhancement and #210 won't be implemented by the beginning of the next year, I will try to write them.

ocefpaf:
It turns out that implement this was relatively easy.  See #330

http://nbviewer.ipython.org/gist/ocefpaf/a502162cf7a472577033

ocefpaf:
@methodds I am guessing #210 is also a low hanging fruit.  Want to try :wink:

nurmians:
nbviewer links are broken, I'm looking for a way to mask map areas without overlay?

Conengmo:
The suggested approach here is to not include any tileset, only your data visualization.

```
m = Map(tiles=None)
# add whatever Choropleth or something you want to add
Choropleth(...).add_to(m)
```

PR #330, opened 2016-01-15
Add a "no tiles" option

ocefpaf:
See #249.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -50,7 +50,8 @@ class Map(MacroElement):
         height: pixel int or percentage string (default: '100%')
             Height of the map.
         tiles: str, default 'OpenStreetMap'
-            Map tileset to use. Can use defaults or pass a custom URL.
+            Map tileset to use. Can choose from a list of built-in tiles,
+            pass a custom URL or pass `None` to create a map without tiles.
         API_key: str, default None
             API key for Cloudmade or Mapbox tiles.
         max_zoom: int, default 18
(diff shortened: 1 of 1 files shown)


BibMartin:
Simple, clean. Thanks @ocefpaf .

ocefpaf:
I am not sure if we can change the grey background to another color though.  But this will be a niche feature anyways.

BibMartin:
An ugly hack: if you want a white background, you can create a Layer pointing to white tiles.

```python
# Create a white image of 4 pixels, and embed it in a url.
white_tile = folium.utilities.image_to_url([[1,1],[1,1]])

# Create a map using this url for each tile.
m = folium.Map(tiles=white_tile, attr='white tile')
```

BibMartin:
I don't think it shall be embeded in folium, but it's good to remember.

ocefpaf:
> I don't think it shall be embeded in folium,

Agree.

> but it's good to remember.

:+1: (That actually gave me a few crazy ideas :wink:)

BibMartin:
>  (That actually gave me a few crazy ideas :wink:)

Yeah ; like:

```python
tile = folium.utilities.image_to_url([[(-1)**((i+j)//30) for i in range(300)] for j in range(300)])
tile = folium.utilities.image_to_url([[(-1)**((i//30+j//30)) for i in range(300)] for j in range(300)])
```

ocefpaf:
I really like the second one:

http://nbviewer.ipython.org/gist/ocefpaf/efcf54d0a3710b324b60

PR #336, opened 2016-01-19
<<Strip>> Figure._repr_html_ output

BibMartin:
A simple PR to avoid that Figure._repr_html_ produces bad-indented-multilined output.

It's ugly, and it breaks the rendering when a notebook is *nbconvert*ed to markdown.


ocefpaf:
> It's ugly, and it breaks the rendering when a notebook is nbconverted to markdown.

Cool!

PR #277, opened 2015-12-02
Added all WMS options and fleshed out the docs.

ocefpaf:
**This is incomplete!  Do not merge!!**

I am waiting for #274  and #276 to finish this one.

PR #348, opened 2016-02-10
WMS options

ocefpaf:
Work in progress to replace #277 

See: http://nbviewer.jupyter.org/gist/ocefpaf/190dd0792ea05c0e5d3d

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -23,45 +23,57 @@ class WmsTileLayer(Layer):
 
     Parameters
     ----------
-    url: str
+    url : str
         The url of the WMS server.
-    format: str, default None
-        The format of the service output.
-        Ex: 'iamge/png'
-    layers: str, default None
-        The names of the layers to be displayed.
-    transparent: bool, default True
-        Whether the layer shall allow transparency.
-    attr: str, default None
-        The attribution of the service. Will be displayed in the bottom right corner.
     name : string, default None
         The name of the Layer, as it will appear in LayerControls
+    layers : str, default None
+        The names of the layers to be displayed.
+    styles : str, default None
+        Comma-separated list of WMS styles.
+    format : str, default None
+        The format of the service output.
+        Ex: 'iamge/png'
+    transparent: bool, default True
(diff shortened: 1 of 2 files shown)


BibMartin:
Happy to see the PR back.

In my mind, we were about to release `v0.2` ; do you want to incorporate this PR in it ?

ocefpaf:
Yep.  I will finish this today and focus on the release.  I don't want to delay it more, but I meed WMS styles there for my own selfish reasons :grin:

ocefpaf:
@BibMartin I removed the legacy WMS method test.  That test was already a little flaky and we will remove method in after `v0.2.0`.  However, I did not add a proper unittest for the new options. 

The current test checks if the HTML is rendered and for the existence of bounds. The bounds are `None` now but I have a plan to use OWSLib to feed that information there.  OWSLib will be an option dependency and I will leave that implementation to `v0.3.0`.

Can you review and merge this one?

folium/features.py
```diff
@@ -23,45 +23,57 @@ class WmsTileLayer(Layer):
...
+        # Options.
+        self.layers = layers if layers else ''
+        self.styles = styles if styles else ''
+        self.format = format if format else 'image/jpeg'
         self.transparent = transparent
-        self.attr = attr
-
+        self.version = version
+        # FIXME: Should be map CRS!
+        # self.crs = crs if crs else 'null
```
BibMartin: Does it mean that you would like to use `this._parent.crs` in the template to set the CRS ?

BibMartin:
OK, go!

folium/features.py
ocefpaf: I understood leaflet correct the WMS service might be in a different `crs` than the map and this variable is used to re-project it to the base map.

ocefpaf:
Thanks @BibMartin!  Preparing the  release now...

BibMartin:
Don't you want to finish #352 before ?

ocefpaf:
Working on that too.

PR #359, opened 2016-02-13
Try branca dependency

BibMartin:
Do not merge ; this is to test folium with a dependency to branca.

But the interest of the split has to be confirmed.


ocefpaf:
> But the interest of the split has to be confirmed.

I thought that :ship: had already :boat: ?

:+1:

ocefpaf:
This re-factor will also be a good test for our tests :wink: 

Have fun making Travis-CI happy!

BibMartin:
> Have fun making Travis-CI happy!

Yes ; I think the bug is on branca's side. I'll put an issue to them...

BibMartin:
And it works!
There are still a few things to move (tests, example notebooks...) but I think it's enough for a first shot.

ocefpaf:
> There are still a few things to move (tests, example notebooks...) but I think it's enough for a first shot.

:+1: 

However, let's hold the merge until we have a first release of `branca`.  When projects are tied together like this the first dependency dictates the releases.

PS: I assign myself for reviewing this and testing the notebook later today or early tomorrow. Ping me if I fail to do so :wink:

.travis.yml
```diff
@@ -16,6 +16,7 @@ before_install:
     - travis_retry conda create --yes -n test python=$PYTHON --file requirements.txt
     - source activate test
     - conda install --yes --file requirements-dev.txt;
+    - pip install git+https://github.com/python-visualization/branca.git
```
BibMartin: @ocefpaf 
> However, let's hold the merge until we have a first release of branca. When projects are tied together like this the first dependency dictates the releases.

This line downloads and install the master of branca, meaning folium tests will run with branca's present master.
I guess this is too dangerous because it breaks tests reproducibility. But I know it's possible to ask pip for a specific commit, so that we could live (on folium's master) on a version of branca that's not necessarily released.

Depending on your reply, I'll either freeze a branca commit into this line, or create a `v0.0.0` of branca and ask for it in this line.

ocefpaf: > Depending on your reply, I'll either freeze a branca commit into this line, or create a v0.0.0 of branca and ask for it in this line.

Actually that depends on your plans.  Even though I recommend a release (`v0.1.0` at least), we can deal with Travis-CI to pointing `master` until the dust settles.  I just don't want that to become a permanent thing.

So,

- if you see a fast development pace in the next few weeks leave it as-is and let's open an issue to remind us to change it later;
- if you are not going to be developing `branca` for a while and/or if you thing other might be interested into using this ASAP, then let's tag a release.

BibMartin:
Finally, I've put a commit number in the `.travis.yml`, for two reasons:

* In short term, it is a good compromise between safety (in ensures test reproducibility) and agility (no need to release branca each time folium wants to benefit from it's changes).
* In longer term, I think we can self-apply these rules:
    * At any time, Folium has to point to a precise commit/tag/release of branca
    * Any release of folium has to point to a precise release of branca

Would it be good enough ?

ocefpaf:
> Would it be good enough?

Yep.  Sounds good.

Another idea for the long term is to split Travis-CI into two tests.  One for the latest branca `master` and another one to a stable release.  I never did that but I saw a few examples.  It is not too complicated.

PS: Ready for merging?

BibMartin:
> Another idea for the long term is to split Travis-CI into two tests. One for the latest branca master and another one to a stable release. I never did that but I saw a few examples. It is not too complicated.

Yes, it can be interesting.It will widely depend on how we organize release cycles. For the moment I imagine to have a release of branca almost each time we release folium. It's not really clear in my mind today.


> PS: Ready for merging?

Yes

ocefpaf:
BTW this PR just appeared under my radar:

https://github.com/rhattersley/pyugrid/blob/separate-coding-checks/.travis.yml

That is an example of context separation we need.

PR #368, opened 2016-02-15
Put explicit links to nbviewer/v0.2.0

BibMartin:
When linking from readthedocs to nbviewer, we arrive on the master version of notebooks (with branca in the code).
This changes the docs **for branch v0.2.0 only** for targetting the proper notebooks in nbviewer.

@ocefpaf Shall we open an issue to automate these chages at each version change ?


ocefpaf:
Thanks!

PR #374, opened 2016-02-27
Try to fix test_geo_pandas

BibMartin:


BibMartin:
@ocefpaf I think this one is the first priority: it solves the problem of dependency in geopandas.

ocefpaf:
The long term solution is to fix gdal in the ioos channel. For now that is
fine. If tests are passing go ahead and merge. (Or I'll do it next week.)

BibMartin:
OK, so self-merging this.

BTW: it seems the problem comes from missing `krb5` dependency.

PR #380, opened 2016-03-09
HeatMap should pass "name" to super

eddies:
HeatMap wasn't passing the name argument, so the label for a heatmap in the LayerControl was defaulting to a uuid-based name.

Also, I "fixed" some dependency issues I ran into so that I could build & test folium:
* `No matching distribution found for krb5 (from -r requirements-dev.txt (line 6))`
* `ImportError: No module named 'branca'`

### DIFF
--- a/folium/plugins/heat_map.py
+++ b/folium/plugins/heat_map.py
@@ -41,7 +41,7 @@ class HeatMap(TileLayer):
         gradient : dict, default None
             Color gradient config. e.g. {0.4: 'blue', 0.65: 'lime', 1: 'red'}
         """
-        super(TileLayer, self).__init__()
+        super(TileLayer, self).__init__(name=name)
         self._name = 'HeatMap'
         self.tile_name = name if name is not None else self.get_name()
 


eddies:
hrm. I see from the Travis failure that conda doesn't know about branca. I don't use/am only passingly familiar with conda. And perhaps krb5 is available via conda--I was only checking pypi. But in any case, all tests passed without `krb5`. And I did see @BibMartin's commit 080fc2f6271dd3bef4f2ebc98efb7149a0d3d1cb referencing `test_geo_pandas` but couldn't find said test.

BibMartin:
Thanks @eddies for your contribution.
(btw your generator-dashi looks great)

> I can toast f66ba6e if that's preferred, since it's not actually necessary for the fix to HeatMap.

Yes, please remove this commit ; we're pointing to a branca commit for the moment because branca is brand new (and instable), but we'll change it before next release.

`krb5` used to be necessary for installing `geopandas` with `ioos` repository. It may not be required now (if `ioos` fixed their `geopandas` distro), but I would keep it for the moment (it's installed anyway as it's a geopandas requirement).

`test_geo_pandas` is *automatically generated* by the script [test_notebook.py](https://github.com/python-visualization/folium/blob/master/tests/test_notebooks.py) that creates one test for each notebook in the [examples](https://github.com/python-visualization/folium/blob/master/examples) folder.

eddies:
Thank @BibMartin for the patient explanation. I removed f66ba6e and Travis appears happy again.

BibMartin:
Awesome; Thanks @eddies

ocefpaf:
:+1: Thanks!

eddies:
@ocefpaf if you're cutting a 0.2.1 release, would you please consider including this as well?

ocefpaf:
> @ocefpaf if you're cutting a 0.2.1 release, would you please consider including this as well?

That is the plan :wink:

Issue #381, opened 2016-03-11
Missing legend in Map.cholopleth()


mojones:
Hi folks, 

First time using Folium and I am blown away by how easy it is to use, fantastic work. I'm trying to create  a cholorpleth map using the function of the same name, and the legend isn't displaying. Looking at the code, I cant see the `legend_name` variable being used anywhere. Is there some other way to attach a description to the legend?

BibMartin:
Well seen @mojones,
At the end of the `colormap` definition (in `folium.py`), we read:
```python
            color_scale = StepColormap(
                brewed[1:len(color_domain)],
                index=color_domain,
                vmin=color_domain[0],
                vmax=color_domain[-1],
                )
            self.add_children(color_scale)
```
and we shall add `caption=legend_name,` like this:
```python
            color_scale = StepColormap(
                brewed[1:len(color_domain)],
                index=color_domain,
                vmin=color_domain[0],
                vmax=color_domain[-1],
                caption=legend_name,
                )
            self.add_children(color_scale)
```
If you have time to do the PR, I ca

ocefpaf:
Fixed in #389

PR #389, opened 2016-03-15
added legend to choropleth colour map

mojones:
as per discussion here: https://github.com/python-visualization/folium/issues/381


ocefpaf:
Ouch! This one escaped me entirely! Thanks @mojones!!

PR #404, opened 2016-03-28
Fix WMS

ocefpaf:
An extra comma was prevent maps from being rendered. Not sure how that was introduced there.


ocefpaf:
@BibMartin easy review :wink:

PR #417, opened 2016-04-13
Add Rectangle Marker and Polygon.

matsuken92:
Hi! I added a 2 features (Rectangle Marker and Polygon).
I would appreciate it if you could accept my pull request.


ocefpaf:
@matsuken92 this one escape me entirely! Sorry. I am merging as-is but there is a big change it will be broken Travis-CI due to the new tests. But I will take care of that.

However, it would be nice if you could send a new PR with an entry to the changelog and a notebook with example on how to use this new functionality.

PR #420, opened 2016-04-26
Allow non-standard params on WmsTileLayer

bielfrontera:
I've tried to add a WMS layer with some non-standard params, but it failed because `WmsTileLayer` got an unexpected keyword argument.

I propose to pass any other kwarg as an option of the WMS layer that is created by folium.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -53,7 +53,7 @@ class WmsTileLayer(Layer):
     """
     def __init__(self, url, name=None, layers=None, styles=None, format=None,
                  transparent=True, version='1.1.1', attr=None, overlay=True,
-                 control=True):
+                 control=True, **kwargs):
         super(WmsTileLayer, self).__init__(overlay=overlay, control=control, name=name)  # noqa
         self.url = url
         self.attribution = attr if attr is not None else ''
@@ -63,6 +63,9 @@ class WmsTileLayer(Layer):
         self.format = format if format else 'image/jpeg'
         self.transparent = transparent
         self.version = version
+        self.extra_params = []
+        for name, value in kwargs.items():
+            self.extra_params.append((name, value))
         # FIXME: Should be map CRS!
         # self.crs = crs if crs else 'null
         self._template = Template(u"""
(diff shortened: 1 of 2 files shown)


folium/features.py
```diff
@@ -73,8 +76,9 @@ def __init__(self, url, name=None, layers=None, styles=None, format=None,
                     styles: '{{ this.styles }}',
                     format: '{{ this.format }}',
                     transparent: {{ this.transparent.__str__().lower() }},
-                    version: '{{ this.version }}',
-                    {% if this.attribution %} attribution: '{{this.attribution}}'{% endif %}
+                    version: '{{ this.version }}'
+                    {% if this.attribution %}, attribution: '{{this.attribution}}'{% endif %}
+                    {% for param in this.extra_params %}, {{param.0}}: '{{param.1}}'{% endfor %}
```
ocefpaf: @bielfrontera thanks for the PR!

I wonder if you should be explicit here about the extra parameter instead of looping over the `kwargs `. What do you think @BibMartin?

bielfrontera: Some of the extra parameters are those included in [ncWMS](http://www.resc.rdg.ac.uk/trac/ncWMS/wiki/WmsExtensions).

But at SOCIB we have extended that library with parameters as markerscale and markerspacing.

ocefpaf: @bielfrontera I am :+1: to all that. I am just concerned if we should be explicit there, bu naming every option, instead of looping over the `kwargs `.

ocefpaf:
@bielfrontera do you mind rebasing this so we can merge it?

bielfrontera:
Hi @ocefpaf , rebase done!

ocefpaf:
> Hi @ocefpaf , rebase done!

Thanks!

folium/features.py
ocefpaf: @bielfrontera where do you stand on adding those parameters explicitly instead of looping over them? We also need to update the docs to describe them. At least the most important subset you expect to use.

ocefpaf:
@bielfrontera I have to apologize but this PR feel under the cracks and I merged #644 which is quite similar. Closing this but we can work out the details to get the same functionalities you expected from this one.

PR #436, opened 2016-05-22
Split branca test

ocefpaf:
This is a test to split travis into a stable released version of branca and the development version. BTW we can tell travis that one of them is "expected to fail." Not sure if that is useful here though.


ocefpaf:
This PR is ending a conda package. See https://github.com/conda-forge/staged-recipes/pull/652

ocefpaf:
Getting there. Testing both `branca` latest and release works. Now I need to fix the notebook conversion issue :grimacing:

ocefpaf:
I believe the kernel specs in some of those notebook might be calling `Travis-CI`'s Python instead of `miniconda`. I will refresh all the notebooks and test this...

ocefpaf:
I had to move the notebooks tests to `Python 3.5` because, for some unknown, reason the notebook tester was calling the root` Python` instead of the Python in the envs. Since the env is created with `Python 3.5` things are OK in this setup. One way around it would be to download the right `miniconda` for each `Python` version, but I don't think that is worth it. Running the notebook tests once is enough.

BTW only `tests/test_notebooks.py::TestNotebooks::test_quickstart` in the Python 3.5 `TEST_TARGET=branca_latest` is failing. We still need to investigate that failure, but failures in `TEST_TARGET=branca_latest` are fair game and the build matrix should "allow" for that.

@BibMartin this is ready for review. I will add more notebooks to the example folder later and see if we can convert them o `HTML` and publish them somewhere for a visual inspection of the tests.

PR #440, opened 2016-06-01
Change links from raw.githubusercontent.com to cdn.rawgit.com.

talespaiva:
Change links from `raw.githubusercontent.com` to `cdn.rawgit.com` to avoid the following error: 

> Refused to execute script from ... because its MIME type (text/plain) is not executable, and strict MIME type checking is enabled

Following the advice in [this question](http://stackoverflow.com/questions/17341122/link-and-execute-external-javascript-file-hosted-on-github) on StackOverflow.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -51,7 +51,7 @@ _default_css = [
     ("marker_cluster_css",
      "https://cdnjs.cloudflare.com/ajax/libs/leaflet.markercluster/0.4.0/MarkerCluster.css"),  # noqa
     ("awesome_rotate_css",
-     "https://raw.githubusercontent.com/python-visualization/folium/master/folium/templates/leaflet.awesome.rotate.css"),  # noqa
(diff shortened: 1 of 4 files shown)


ocefpaf:
Awesome! Thanks @talespaiva!!

talespaiva:
I just noticed that this modification may cause future problems. `cnd.rawgit.com` points to a permanent version of the file that is created in the CDN once the file is first accessed. So, if the file is changed in the project, the CDN version is not updated. Source: [this thread](http://stackoverflow.com/questions/23537239/how-can-i-force-cdn-rawgit-com-use-updated-file) and some tests).

The fix for this is just removing the  `cdn.` portion of the URL.

talespaiva:
@ocefpaf Should I open a new PR?

ocefpaf:
> @ocefpaf Should I open a new PR?

Yes please :grimacing:

PR #448, opened 2016-06-11
Removing `.cdn` from rawgit URLs

talespaiva:
This avoids future bugs caused by #440.


ocefpaf:
Thanks @talespaiva!

QuLogic:
This might get throttled or blacklisted, though. Maybe you can use a version-specific URL?

PR #452, opened 2016-06-21
Smooth factor example

JamesGardiner:
@ocefpaf here's a smooth factor example and an updated CHANGES.txt.


ocefpaf:
Thanks @JamesGardiner!

PR #471, opened 2016-07-21
Added more classifiers to setup.py

timtroendle:
I was just browsing through pypi to find this package and didn't find it among the GIS packages. Don't you want to add these classifiers?


ocefpaf:
No need to wait on the CIs for this one. Thanks @timtroendle!

PR #487, opened 2016-08-18
Fix typo in style_function

jdoerrie:
This typo renders the `line_opacity` attribute of `Map.choropleth` useless. The error is propagated to the generated JavaScript, where Leaflet ignores it.

This one line fix corrects the error. Consider the attached screenshots for a demonstration of the fix applied to US Unemployment example.

### Prior to Fix:
us_states

### After Fix:
us_states_fix

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -673,7 +673,7 @@ class Map(LegacyMap):
         def style_function(x):
             return {
                 "weight": line_weight,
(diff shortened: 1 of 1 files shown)


ocefpaf:
Thanks @jdoerrie! The failure here is unrelated to your PR. So merging this.

PR #514, opened 2016-09-23
Pep8 notebooks

ocefpaf:
This test always return `0` exit code because it is inside a loop. That is OK for now b/c we do not want to enforce this but just use it as a guide to get better code in the notebooks.


ocefpaf:
As expected the tests pass but the lints are logged for inspection:

https://travis-ci.org/python-visualization/folium/jobs/162192922#L779

Issue #319, opened 2016-01-08
Add PNG/SVG images/legend using Leaflet's L.control


ocefpaf:
We could try to implement adding objects using relative map positions like this:

![map_rose](https://cloud.githubusercontent.com/assets/950575/12208569/55c11956-b62c-11e5-92dc-4092fc0e66e9.png)

Here is the full notebook:

http://nbviewer.ipython.org/urls/gist.githubusercontent.com/ocefpaf/3c03e06a2d5347934699/raw/948d4ebbc5e2354df4ce9d3c6f68035f3032e90d/map_legend.ipynb

Note that the rose will always stay at the top right when interacting with the map.

Things that we need to consider:
- Is the `L.control` the only one to do this?  If so can we avoid clobbering it when we use `L.control.layer`?
- SVGs could probably be more powerful and added via `.css`, but sure if that is worth though...

ocefpaf:
Implemented in #537

PR #537, opened 2016-10-21
Add a float image plugin

juoceano:
Implements #319 with a logic similar to `ScrollZoomToggler` instead of using `L.DomUtil`. It is simpler, but the relative position is not against the actual map and may be misleading.

### DIFF
--- a/tests/plugins/test_float_image.py
+++ b/tests/plugins/test_float_image.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+"""
+Test FloatImage
+---------------
+"""
+
+from jinja2 import Template
+
+import folium
+from folium import plugins
+
+
+def test_float_image():
+    m = folium.Map([45., 3.], zoom_start=4)
+    url = 'https://raw.githubusercontent.com/SECOORA/static_assets/master/maps/img/rose.png'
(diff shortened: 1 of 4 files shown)


ocefpaf:
Nice one!

> Implements #319 with a logic similar to `ScrollZoomToggler` instead of using `L.DomUtil`. It is simpler, but the relative position is not against the actual map and may be misleading.

That is fine for now. At least people will be able to add extra image elements easier now.

Can you add a test for the new feature?

tests/plugins/test_float_image.py
```diff
@@ -0,0 +1,44 @@
...
+    szt = plugins.FloatImage(url, bottom=60, left=70)
+    m.add_child(szt)
+    m._repr_html_()
+
+    out = m._parent.render()
+
+    # Verify that the div has been created.
+    tmpl = Template("""
+        <img id="{{this.get_name()}}" alt="float_image"
+        src="https://raw.githubusercontent.com/SECOORA/static_assets/master/maps/img/rose.png"
```
ocefpaf: We should probably copy that image here.

ocefpaf:
Awesome!  Another one bites the dust!

PR #547, opened 2016-10-24
Update map.py

mbudda:


ocefpaf:
@mbudda I am closing this for now since you seems to be away. Let me know if you want to pursue this further as a plugin and I will re-open this.

Issue #557, opened 2016-10-26
Add heatmap notebook


ocefpaf:
https://github.com/ocefpaf/folium_notebooks/blob/master/test_heatmap.ipynb

ocefpaf:
Donein #558.

PR #558, opened 2016-10-26
added heatmap example

jsignell:
Addresses #557


ocefpaf:
Minor code standard issue. I will solved that later b/c I am in the process of revising the pep8 rules I want to check.

Thanks!

Issue #556, opened 2016-10-26
Add tiles notebook example/test


ocefpaf:
https://github.com/ocefpaf/folium_notebooks/blob/master/test_tiles.ipynb

HashCode55:
Do I like just need to port that file in this repository or am I getting it wrong?

ocefpaf:
> Do I like just need to port that file in this repository or am I getting it wrong?

Kind of. Make sure the notebook runs and the tiles are properly display. Feel free to improve the text too :smile: 

I believe that MapQuest tiles needs addressing to. Not sure if there is a similar source for it or if we need to just remove it from the list.

jsignell:
What functionality was MapQuest supposed to provide that was different or better than OpenStreetMap

ocefpaf:
No functionality just a different "look." Some people liked it I guess... I actually never used it!

ocefpaf:
Now that I think about it let's just remove the `MapQuest` options/code everywhere. See https://github.com/python-visualization/folium/issues/559

ocefpaf:
Done in #560.

@HashCode55 #559 is a low hanging fruit. Do you want to try that one?

HashCode55:
yeah I'll take that one

PR #560, opened 2016-10-26
added tile example notebook

jsignell:
Addresses #556. I removed references to MapQuest and services that require API keys.


ocefpaf:
> I removed references to MapQuest and services that require API keys.

:+1:

PR #567, opened 2016-10-30
Heat map with time

Padarn:
Added a plugin to create a HeatMapWithTime layer.


examples/HeatMapWithTime example.ipynb
```diff
@@ -0,0 +1,71 @@
...
+     "execution_count": 1,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "import folium\n",
+    "import folium.plugins as plugins\n",
+    "\n",
+    "x , y = -35.19925725627544, 149.0817133320111\n",
```
BibMartin: We've moved to Cambera :wink: It seems to be a nice place.

folium/plugins/heat_map_withtime.py
```diff
@@ -0,0 +1,245 @@
...
+        self.style_NS = style_NS
+        self.time_slider = "true" if time_slider else "false"
+        self.time_slider_drap_update = "true" if time_slider_drap_update else "false"
+        self.time_steps = time_steps
+        self.time_control_title = time_control_title
+
+        self._template = Template(u"""
+        {% macro script(this, kwargs) %}
+
+            var times = {{this.times}};
```
BibMartin: In the gist I did, it was :
```javascript
var times = [];
for (var i=0; i<data.length; i++) {times.push(1+i);}
```
The interest was to let the user manage the timeline through the `index` attribute instead of `times`.
Your code is obviously even more agile, but I wonder if it will not seem too complex to the final user.
It may depend on the docstring.

It's just a question ; I let you take the decision.

Padarn:
Thanks for the feedback @BibMartin. I will address your specific concerns as I fix them, and will add the docs etc.

Just a question: Do you think we should make sure all of the customisation options are exposed for the heat map and control, or maybe simplify direct exposure? There are for example some settings I can see here:
https://github.com/pa7/heatmap.js/blob/master/src/config.js
that cannot currently be changed in the plugin. 

re input format: I agree keeping it simple makes sense, but I would like to keep the possibility of more user control. I was hoping to also add some helper functions to turn a Panda's DataFrame with a time index into a data set broken into steps too.

folium/plugins/heat_map_withtime.py
Padarn: I've decided to go with what you've suggested and removed 'time'. I've tried to make it clear through my doc-string and the other options that index is essentially the time step though.

Padarn:
I've pushed a new version - I renamed (and changed) the example, which can now be found here: https://github.com/Padarn/folium/blob/HeatMapWithTime/examples/HeatMapWithTime.ipynb

xen0l:
@ocefpaf @Padarn I used this and it works as expected. Though, I had to make small changes:

* switch http links to https, so I avoid warnings about mixed content
* switch from rawgit.com to cdn.rawgit.com, so we leverage CDN

I have changes available in my fork located at https://github.com/xen0l/folium.

ocefpaf:
I overlooked this PR...  And many others :flushed: 

I'll try to revive this over the weekend and get some work done for `folium`. Thanks @Padarn and @xen0l!

xen0l:
@ocefpaf cool, do you think is there any other work to be done apart from reported issues? I would like to have this merged, so I can get rid of my fork.

xen0l:
@ocefpaf @Padarn one more issue not (directly) related to this:

* heatmap doesn't work with adblockers (I'm using ublock origin) as it blocks heatmap.js

You can check that out on https://bicing.xenol.eu/graphs/experiments/heatmap.

xen0l:
@ocefpaf I ran pep8 on this PR and outcome is here: https://github.com/xen0l/folium/commit/fa5f17ff6b1b448f2299a60784023e84e30688a4. I would also recommend increase number of maximum lines from 100 to 120 in .stickler.yml as PEP8 recommends this as well. With that there will be only one issue left to address. 

What do you think?

ocefpaf:
> @ocefpaf I ran pep8 on this PR and outcome is here: https://github.com/xen0l/folium/commit/fa5f17ff6b1b448f2299a60784023e84e30688a4.

Don't worry about it right now. I'll try to review this ASAP and I'll ignore the minor pep8 stuff at the moment.

> I would also recommend increase number of maximum lines from 100 to 120 in .stickler.yml as PEP8 recommends this as well. With that there will be only one issue left to address. 

Agreed. If you want to send another PR with that please do so :grin:

ocefpaf:
The errors are unrelated to this PR and I'll fix them in the next one. Thanks @Padarn! This is an awesome addition to `folium`.

PR #568, opened 2016-10-30
remove src file

birdage:
@ocefpaf we might not need the explicit src file, lets see how the build does


ocefpaf:
We are relying on that for the tests. But you are correct, we should re-fator the tests and get rid of this file.

I want to clean the HTML first before re-factoring the tests. There are many blank lines, whitespaces, etc. It creates a dirty and hard to edit HTML.

birdage:
@ocefpaf sounds like a good plan!

ocefpaf:
I'll fix the use of that in the code in another PR. Merging it. Thanks @birdage!

Issue #524, opened 2016-10-11
Additional parameters for CircleMarker


carderne:
I just started using Folium and it's blowing me away. I spent ages googling looking for Python mapping packages, but don't know JS at all so avoided that end and somehow never found this project.

I have quite a strange use-case, that is plotting hundreds of circles on an even grid. How they look is then quite important. How easy is it to add some further parameters for CircleMarker? For my application I'm trying to control the stroke (or remove it). But also to add customised click/hover actions.

I tried to go through your code, but I'm not sure where exactly the border between folium and leaflet is, so I don't know how simple this is to implement. I'm also interested what if any communication there is between folium and ipyleaflet?

olibchr:
Check out my issue request #523

carderne:
Great! Could you share your new features.py or should I just wait for it to merge?

ocefpaf:
Just implemented that in #683

PR #569, opened 2016-10-31
Add class definition

birdage:
@ocefpaf what do you think about adding a class definition to the markers. ive put an example of the proposed changes below for just the `circle`. Wed have to figure where to put the css, maybe `folium/templates/user_defined.css`? i can squash the multiple changes if we want to go down this path.

i.e we have a circle marker but we want to leverage css styling a bit more, this would aid in the following issue https://github.com/python-visualization/folium/issues/524

> leaflet js example
> 
> ``` js
>     L.circle([51.509, -0.11], 1500, {
>         className:" sample-circle",
>         color: 'red',
>         fillColor: '#f03',
>         fillOpacity: 0.5
>     }).addTo(mymap).bindPopup("I am a circle.");
> ```
> 
> css style
> 
> ``` css
> .sample-circle:hover{
>     stroke: blue;
> }
> ```
> 
> normal circle
> screen shot 2016-10-30 at 8 15 13 pm
> 
> on hover
> screen shot 2016-10-30 at 8 15 16 pm

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -712,18 +712,19 @@ class Circle(Marker):
         The fill opacity of the marker, between 0. and 1.
     popup: string or folium.Popup, default None
         Input text or visualization for object.
+    class_name: string, default is an empty string
+        Class name to associate with the circle, known as ClassName in the leaflet spec
 
     """
     def __init__(self, location, radius=500, color='black',
-                 fill_color='black', fill_opacity=0.6, popup=None, class_name=None):
+                 fill_color='black', fill_opacity=0.6, popup=None, class_name=''):
         super(Circle, self).__init__(location, popup=popup)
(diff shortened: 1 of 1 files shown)


ocefpaf:
This looks nice. I am unfamiliar with the leaflet counter part to review this right now but I will take a look at it ASAP.  Meanwhile we should add a notebook example and tests.

I am not sure about providing the `css`. Maybe the end user should ensure that the CSS is loaded in their application. I do not think we should try to cover all the possibilities... But I am open to suggestions. What to you think @BibMartin?

birdage:
@ocefpaf i just think we could add an empty css file that the user could add their css too 😄

ocefpaf:
@birdage I took leaflet's approach and added the default options. See #683

PR #574, opened 2016-11-20
added multi option circle marker

olibchr:


folium/features.py
```diff
@@ -787,6 +787,134 @@ def __init__(self, location, radius=500, color='black',
...
+
+            info.addTo({{this._parent.get_name()}});
+
+
+
+        {% endmacro %}
+        """)
+
+
+class CustomCircleMarker(Marker):
```
ocefpaf: @birdage this is related to your work in #524, right? Can you take a look at this part and join forces with @olibchr.

folium/features.py
```diff
@@ -787,6 +787,134 @@ def __init__(self, location, radius=500, color='black',
             """)
 
 
+class CustomLegend(Layer):
```
ocefpaf: @olibchr Maybe it is better if you send this in a separate PR. It will be easier to merge.

ocefpaf:
@olibchr awesome additions!

Can you add tests, a changelog entry, and examples in notebooks gallery?

CHANGES.txt
```diff
@@ -20,6 +20,8 @@ Bug Fixes
 - Fixed MIME type (text/plain) is not executable (talespaiva #440)
 - Update Travis-CI testing to incorporate branca and fix notebook tests (ocefpaf #436)
 - Removed MultiPolyLine and MultiPolygon, both are handled by PolyLine and PolyLine in leaflet 1.0.* (ocefpaf #554)
+- Added CustomCircleMarker (olibchr #574)
```
ocefpaf: I don't think these are bugfixes. They belong in the section above.

folium/features.py
ocefpaf: What do you think about my comment above?

ocefpaf:
Please remove the `.DS_Store` file and add an example, preferably in one of the existing notebooks, for the new feature.

ocefpaf:
There are some interesting ideas here but I don't think we should implement those as features.
They should be either a plugin or an example on how to extend `folium` to other needs.

I am re-factoring `features.py` to reflect leaflet's docs, so this does not fit there. Closing for the moment.

PR #579, opened 2016-12-14
added support to bind popups to a field on the feature in a GeoJson Layer

kputland:
eg. 

GeoJson(popup='fields.properties.popupContent')

--Karl

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -309,7 +309,7 @@ class GeoJson(Layer):
     """
     def __init__(self, data, style_function=None, name=None,
                  overlay=True, control=True, smooth_factor=None,
-                 highlight_function=None):
+                 highlight_function=None, popup=None):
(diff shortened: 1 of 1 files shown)


ocefpaf:
I like this implementation but I would like to get the old behavior, of reading the `popupContent` from the JSON, back too. That is unrelated to this PR though, but maybe you are interested in tackling that issue...

folium/features.py
```diff
@@ -309,7 +309,7 @@ class GeoJson(Layer):
     """
     def __init__(self, data, style_function=None, name=None,
                  overlay=True, control=True, smooth_factor=None,
-                 highlight_function=None):
+                 highlight_function=None, popup=None):
```
ocefpaf: This needs a test and an example but looks good to go.

kputland:
Take it or leave it... I've change to google maps api for now as there was a bug in leaflet-realtime that I couldn't work around.

PR #581, opened 2016-12-31
Added weight option to CircleMarker

palewire:


ocefpaf:
Looks good! Thanks!!

I guess you will need to update the test to incorporate the new `kwarg`.

palewire:
Yes. Sorry for the testing oversight. I am still getting the hang of this system. I will have a fix up shortly.

ocefpaf:
> Yes. Sorry for the testing oversight. I am still getting the hang of this system. I will have a fix up shortly.

No problem. It is the holidays and I finally have some time for `folium`. So you are in luck that your PR got a quick review :smile:

palewire:
The tests are now passing. Thank you for your patience.

folium/features.py
```diff
@@ -753,6 +753,8 @@ class CircleMarker(Marker):
         use Circle.
     color: str, default 'black'
         The color of the marker's edge in a HTML-compatible format.
+    weight: int, default 2
```
ocefpaf: `Leaflet` docs are hard to follow sometimes. It says that `CircleMarker` extends `Circle` that extends `Path` which has a `weight` default of `5`. What is the reason that lead you to choose `2` instead? (Just curios, you don't need to change to `5`. Images maybe help us decided here :smile: )

palewire:
The only reason I picked it was because I saw it as the default on another
class that already had the option. I'm fine with 5 if that fits better.

ocefpaf:
> I'm fine with 5 if that fits better

To be honest I never plotted the difference :grimacing: 

Let's merge with `2` for the sake of consistency and, if `5` looks better, we can change that in another PR for all the classes.

Thanks @palewire for the awesome addition!

PR #582, opened 2016-12-31
Add a .gitignore file with *.pyc

palewire:
I noticed this was missing when I prepared a more significant pull request. Any objections to adding it?


ocefpaf:
Hi @palewire I see why you want to add that based on https://github.com/python-visualization/folium/pull/581/commits/fcadba13b18eae35a2271d28d2a6f88d8413eaad but I like to leave the `.gitignore` up to each dev. So you can create your own and add `.gitignore` to `.gitignore` as well :smile:  (That is what I do BTW.)

palewire:
No problem!

PR #585, opened 2017-01-06
Add Fast marker cluster to plugins

### LINKED ISSUE #416: Add a script version of MarkerCluster to speedup creating many markers.
For example following code costs 5 seconds to add 1000 markers:

```
import folium
import numpy as np
from folium.plugins import MarkerCluster
N = 1000
fig = folium.element.Figure()
map_ = folium.Map(location=[20, 50], zoom_start=3)
cluster = MarkerCluster(np.random.uniform(0, 90, (N, 2))).add_to(map_)
map_.add_to(fig)
```

and the output html contains code for 1000 markers.

I  tried a Javascript version to speedup this:

```
class MarkerClusterScript(MarkerCluster):
    def __init__(self, data, callback):
        from jinja2 import Template
        super(MarkerClusterScript, self).__init__([])
        self._name = 'Script'
        self._data = data
        if callable(callback):
            from flexx.pyscript import py2js
            self._callback =  py2js(callback, new_name="callback")
        else:
            self._callback = "var callback = {};".format(_callback)

        self._template = Template(u"""
(function(){
    var data = {{this._data}};
    var map = {{this._parent.get_name()}};
    var cluster = L.markerClusterGroup();
    {{this._callback}}

    for (var i = 0; i < data.length; i++) {
        var row = data[i];
        var marker = callback(row);
        marker.addTo(cluster);
    }

    cluster.addTo(map);
})();
{% endmacro %}
            """)
```

and then create the map by

```
import pandas as pd
N = 1000
df = pd.DataFrame(np.random.uniform(0, 90, (N, 2)), columns=list("xy"))
df["color"] = np.random.choice(["red", "green", "blue"], N)

fig = folium.element.Figure()
map_ = folium.Map(location=[20, 50], zoom_start=3)

def create_marker(row):
    icon = L.AwesomeMarkers.icon({markerColor: row.color})    
    marker = L.marker(L.LatLng(row.x, row.y))
    marker.setIcon(icon)
    return marker

MarkerClusterScript(df.to_json(orient="records"), callback=create_marker).add_to(map_)
map_.add_to(fig)
```

It only takes 0.3 seconds. All the data are saved in a javascript array, and the for-loop for creating markers occured in web brower. 

I suggest to add this kind of Classes for speedup.

JamesGardiner:
This should close #416. Adds a faster way to add marker clusters to the output map, using a javascript based callback to construct MarkerCluster objects in the browser. Script originally proposed by @ruoyu0088. I've modified that original proposal to remove the dependency on flexx and tornado.

Users can specify their own callback string should they want to by passing it as the callback kwarg when creating a FasterMarkerCluster object.


tests/plugins/test_fast_marker_cluster.py
```diff
@@ -0,0 +1,58 @@
...
+        np.random.uniform(low=-12, high=30, size=n),  # Random longitudes.
+        range(n),                                     # Popups.
+    ]).tolist()
+    m = folium.Map([45., 3.], zoom_start=4)
+    mc = plugins.FastMarkerCluster(data, callback=None)
+    m.add_child(mc)
+    m._repr_html_()
+
+    out = m._parent.render()
+    print(out)
```
ocefpaf: Please remove the debug `print`s.

folium/plugins/fast_marker_cluster.py
```diff
@@ -0,0 +1,68 @@
...
+                    var row = data[i];
+                    var marker = callback(row);
+                    marker.addTo(cluster);
+                }
+
+                cluster.addTo(map);
+            })();
+            {% endmacro %}""")
+
+    def create_marker(self):
```
ocefpaf: You probably do not need to do this as a function.

folium/plugins/__init__.py
```diff
@@ -19,6 +20,7 @@
 
 __all__ = [
     'MarkerCluster',
+    'FastMarkerCluster',
```
ocefpaf: How does this compares with our `MarkerCluster` besides speed? Can we replace the current one with this?

JamesGardiner: There's some significant differences - the main one being that the FastMarkerCluster object calls `super(FastMarkerCluster, self).__init__([])` - so no children are added and full functionality isn't available (for instance, calling `get_bounds()` returns [[None, None], [None, None]]`. I'd recommend keeping both and caveating FastMarkerCluster as being quick to draw 000's of points but loses functionality.

ocefpaf:
@JamesGardiner I am a little bit swamped right now to review this. But this change is very welcomed!
(Just letting you know that I did not abandon your PR but I will be slow to respond here.)

JamesGardiner:
No problem @ocefpaf - I completely understand. I'm finding it difficult to get time to review the notebook failure in any case. Hopefully I can get all tests passing this weekend.

CHANGES.txt
```diff
@@ -1,6 +1,7 @@
 0.3.0
 ~~~~~
 
+- Added `FastMarkerCluster` (James Gardiner #585 (proposed by @ruoyu0088))
```
ocefpaf: :+1:

folium/plugins/__init__.py
ocefpaf: Thanks for the clarification!

ocefpaf: I did not read the docstrings yet, but that information should be there.

folium/plugins/fast_marker_cluster.py
```diff
@@ -0,0 +1,71 @@
...
+
+class FastMarkerCluster(MarkerCluster):
+    """Add marker clusters to a map using in-browser rendering"""
+    def __init__(self, data, callback=None):
+        """Add marker clusters to a map using in-browser rendering.
+           Using FastMarkerCluster it is possible to render 000's of
+           points far quicker than the MarkerCluster class. Be aware
+           that the FastMarkerCluster class does not retain a
+           reference to any marker data, and therefore methods such as
+           get_bounds() are not available when using it.
```
ocefpaf: Can you expand on the difference from the "regular" MarkerCluster?

ocefpaf:
Ignore the Travis-CI coding standards failures as those are related to new and old notebooks. I will fix that later.

Can you just comment on the docstring so we can merge this?

Thanks for the effort!

ocefpaf:
> Can you just comment on the docstring so we can merge this?

@JamesGardiner what is the status on this? I am planning on issuing a new release and it would be nice to get this PR merged before that.

JamesGardiner:
@ocefpaf - I can add the docstrings in before the end of this week if that helps?

ocefpaf:
> @ocefpaf - I can add the docstrings in before the end of this week if that helps?

That works :smile: 

My original plan was to issue a new release before SciPy, but that is not going to happen. (Day job is keeping me busy.)

JamesGardiner:
@ocefpaf I've updated the docstrings in c642388

ocefpaf:
@JamesGardiner awesome! Thanks!!

Just a minor comment and we are good to go.

JamesGardiner:
No problem, I've pushed a fix for the typo :+1:

ocefpaf:
Awesome! Thanks!!

PR #601, opened 2017-03-06
Fix doc bug

### LINKED ISSUE #584: Remote GeoJSON URLs in Folium (https? http?)
Using release 0.2.1 there _seems_ to be either a documentation clarity issue or an actual support issue. As I understand it from reading the docs, the following code _should_ work:
```python
import folium  
json_url = "https://raw.githubusercontent.com/datasets/geo-boundaries-world-110m/master/countries.geojson"
m = folium.Map(location=(0,0), zoom_start=2)
m.choropleth(geo_path=json_url)
m
```
_Note: I've stripped out the choropleth styling parameters for simplicity's sake._

Instead, I get the following error:
```python
IOError: [Errno 2] No such file or directory: 'https://raw.githubusercontent.com/datasets/geo-boundaries-world-110m/master/countries.geojson'
```
This error implies that only locally-accessible files are supported, but the documentation states only that:
```
Parameters
    ----------
    geo_path: string, default None
        URL or File path to your GeoJSON data
```
I can obviously get around this using the `requests` library and passing `r.json()` to the `geo_str` parameter, but it would be more elegant and useful (as well as being consistent with the documentation) if that weren't actually necessary.

ocefpaf:
Closes #584

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -175,7 +175,7 @@ class Map(LegacyMap):
         Parameters
         ----------
         geo_path: string, default None
-            URL or File path to your GeoJSON data
+            File path to your GeoJSON data
         geo_str: string, default None
             String of GeoJSON, alternative to geo_path
         data_out: string, default 'data.json'


ocefpaf:
Closing in favor of #602

PR #605, opened 2017-03-13
Replace seawater with geographiclib.

QuLogic:
There's no real use of the seawater package for anything seawater-related, but instead it's used for calculating azimuths. GeographicLib can do this much more accurately (though it's not really a huge difference.)

Plus, its tests are rather heavy and I'm a bit wary of the license (cf. pyoceans/python-seawater#16).

This won't pass without conda-forge/staged-recipes#2586 though.


ocefpaf:
It is only used to build one of the examples. So you need to update the example too.

QuLogic:
It is already updated. Also, fixed a warning from ipykernel.

ocefpaf:
Missed that. But we don't have your alternative packaged. Could you try
https://github.com/axiom-data-science/pygc instead?

ocefpaf:
Thanks @QuLogic. Sorry for the delay to merge this.

PR #612, opened 2017-04-04
Fix highlight_function breaking popup in GeoJson

xinluh:
Current behavior:
`GeoJson(highlight_function=None).add_child(folium.Popup('some popup'))` => click on shape, popup opens
`GeoJson(highlight_function=lambda x: {'weight': 5}).add_child(folium.Popup('some popup'))` => hover on shape, line weight changes (as expected); click on shape, no popup

Highlight function overrides the on click behavior - this PR adds it back. Not sure if this is the best way to do it though...

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -365,6 +365,7 @@ class GeoJson(Layer):
                         mouseover: function(e) {
                             e.target.setStyle(e.target.feature.properties.highlight);},
                         click: function(e) {
+                            layer.openPopup();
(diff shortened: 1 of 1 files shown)


ocefpaf:
What version are you using? I was able to do this http://nbviewer.jupyter.org/gist/anonymous/45fa2b172d1f655a766e2e7567805213 with the latest version.

Closing this because I cannot reproduce it but let's keep the discussion alive until we figure out what happened with your installation there.

PR #623, opened 2017-05-25
Add subdomains option for TileLayer

damselem:
Some tile services use a different set of subdomains. For reference: http://leafletjs.com/reference-1.0.3.html#tilelayer-subdomains

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,7 @@
+0.4.0
+~~~~~
+- Added support for subdomains options in TileLayer (damselem #623)
+
 0.3.0
 ~~~~~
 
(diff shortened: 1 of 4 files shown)


ocefpaf:
I am 👍 to that change. Can you update the test template to reflect the extra line:
`subdomains: {{this.subdomains}}` and add a test case for something that is different from the default argument?

We need a changelog entry for that.

Bonus if you add a notebook to our gallery with a few use cases 😄

damselem:
Do you agree this should bump the minor version?

ocefpaf:
> Do you agree this should bump the minor version?

Makes sense. Don't forget to add a changelog entry.

damselem:
Is there anything else that needs to be done in order to get this merged?

ocefpaf:
> Is there anything else that needs to be done in order to get this merged?

Nope. I've just been awfully busy with the day job.

Thanks for the PR!

Issue #625, opened 2017-05-26
Value Check on Location parameter of the Map creator


radumas:
I accidentally tried calling `folium.Map([45.0 -79.0])` in a Notebook, and it failed silently leading me to tried to decode the indecipherable JS console output.

There could be a simple check to see, if `location` is a list, that it has 2 and only 2 values.

ocefpaf:
That is doable. Do you want to try a PR?

On May 26, 2017 4:40 PM, "Raphael Dumas" <notifications@github.com> wrote:

I accidentally tried calling folium.Map([45.0 -79.0]) in a Notebook, and it
failed silently leading me to tried to decode the indecipherable JS console
output.

There could be a simple check to see, if location is a list, that it has 2
and only 2 values.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/python-visualization/folium/issues/625>, or mute the
thread

radumas:
I'd be happy to!

ocefpaf:
Nice. Ping me if you need any help.

radumas:
I've written up the code. Just wondering how to write tests compatible with your current testing framework, I'm only familiar with using the built-in `unittest` library.

ocefpaf:
Start a PR and I can help you with the tests. I'm traveling at moment though and it may take a while for me to look at it.

radumas:
Just started that. Thanks!

ocefpaf:
Fixed in #626

PR #626, opened 2017-05-29
Add value checking for location #625

radumas:
Using property decorators.
Need to implement testing.


radumas:
I'll wait to fix that whitespace error until there are other comments on this PR

folium/map.py
```diff
@@ -297,6 +296,27 @@ def render(self, **kwargs):
             '</style>'), name='map_style')
 
         super(LegacyMap, self).render(**kwargs)
+    
+    @property
+    def location(self):
+        return self._location
```
ocefpaf: I am not sure we need a property and a setter for this. Can you explain what you have in mind?

I thought that we could have an internal function to check and raise a `ValueError` that we would call in the places we need to check for the two value list.

folium/map.py
```diff
@@ -297,6 +296,27 @@ def render(self, **kwargs):
...
+    
+    @location.setter
+    def location(self, value):
+        """Validates the location values before setting"""
+        if type(value) not in [list, tuple]:
+            raise TypeError("Location is not a list, expecting ex: location=[45.523, -122.675]")
+        
+        if len(value) != 2:
+            raise ValueError("Location should have two values, [lat, lon]")
```
ocefpaf: I like that message. Simple and clear!

folium/map.py
```diff
@@ -297,6 +296,27 @@ def render(self, **kwargs):
...
+        
+        for val in value:
+            try:
+                float(val)
+            except:
+                raise TypeError("Location values should be numeric, {} is not a number".format(val))
```
ocefpaf: In a way this is still a `ValueError` b/c some strings will pass your `float(val)`  line. But that is not too important.

PS: I never tested if

```python
location = ['12.5', 42.5]
```

would work on leaflet. But that will pass your test there.

folium/map.py
radumas: That's a fair point. I was more worried about changing anything in an `__init()__` function, since it seems like the property decorators leave those alone, but it does result in unnecessary code. I could swap out for a.... class function?

folium/map.py
radumas: oh that's an interesting point. I guess I could actually convert the values to float or number and then we wouldn't need to know whether it works in leaflet or not.

ocefpaf:
@radumas what is the status on this? (I am planning to isseu a new release.)

radumas:
@ocefpaf I was waiting for your response to my replies to your comments.

ocefpaf:
> @ocefpaf I was waiting for your response to my replies to your comments.

Sorry. My time for folium is so limited that I always need to start from zero when I get back to PRs...

folium/map.py
ocefpaf: Yep. I would go for that approach.

folium/map.py
ocefpaf: Not sure we need a class function. The location is used only once to create the map. We don't need to store it as a property b/c we won't do anything with it later. Just a private function to convert to float and issue a `ValueError` when that cannot be done would suffice IMO. Then you can run that function in `__init__()`, it would be just 1 extra line there. What do you think?

radumas:
Totally understandable. I've updated based on comments.

folium/map.py
```diff
@@ -60,6 +60,19 @@
...
+        if type(values) not in [list, tuple]:
+            raise TypeError("Location is not a list, expecting ex: location=[45.523, -122.675]")
+
+        if len(values) != 2:
+            raise ValueError("Location should have two values, [lat, lon]")
+
+        try:
+            values = [float(val) for val in values]
+        except:
+            raise ValueError("Location values should be numeric, {} is not a number".format(val))
```
ocefpaf: :+1:

folium/map.py
```diff
@@ -60,6 +60,19 @@
      'https://rawgit.com/python-visualization/folium/master/folium/templates/leaflet.awesome.rotate.css'),  # noqa
     ]
 
+def _format_lat_lon(values):
```
ocefpaf: How about naming it `_validate_location`?

folium/map.py
```diff
@@ -170,7 +184,7 @@ def __init__(self, location=None, width='100%', height='100%',
             self.location = [0, 0]
             self.zoom_start = min_zoom
         else:
-            self.location = location
+            self.location = _format_lat_lon(location)
```
radumas: Just as I pushed this I realized this mistake... sigh

folium/map.py
```diff
@@ -60,6 +60,20 @@
      'https://rawgit.com/python-visualization/folium/master/folium/templates/leaflet.awesome.rotate.css'),  # noqa
     ]
 
+def _validate_location(values):
```
ocefpaf: Just this one left to address. I'll merge as is and fix it later. Thanks @radumas!

PR #630, opened 2017-06-15
Added a minimum working example for the highlight_function

eHanseJoerg:
I added a Jupyter Notebook and some csv data to show how the highlight function works. 
The Jupyter NB depends on pandas, os, folium, json.

Additionally, I added two lines to the docstring in the GeoJson class because the paramter list was incomplete. I did not work on chloropleth class.


ocefpaf:
Thanks @eHanseJoerg!

GitHub makes it hard to review notebooks. So here is a `nbviewer` link to hlelp: http://nbviewer.jupyter.org/github/eHanseJoerg/folium/blob/master/examples/Highlight_Function.ipynb

joshuacano:
Thanks @ehansejorg! I won't be able to add the docstrings for another week when I'm back from vacation, I'd be glad to when I get back.

Joshua Cano

ocefpaf:
Thanks @eHanseJoerg!

elochk:
Thanks for the example! I am trying to plot polygons in the same way, but nothing displays at the map. When I change type "Polygon" to type "LineString" keeping the same coordinates, they are displayed. I would be grateful if you could help with this issue.

amirham:
In the same situation as @elochk, this doesn't seem to work for Polygons.

PR #641, opened 2017-07-08
Update map.py to set max_bounds to False as default

rdd9999:
Set max_bounds to False as default


ocefpaf:
Indeed leaflet's default is False. See http://leafletjs.com/reference-1.2.0.html#map-maxbounds
I'll merge this and fix the errors in another PR. Thanks!

PR #644, opened 2017-07-20
Adding Leaflet.timeDimension support for WmsTileLayers

acrosby:
...as well as additional kwargs as a way to support non-standard WMS request parameters in WmsTileLayers. Couldn't come up with a good way to setup a test, open to suggestions.


acrosby:
Actually, want to smooth some issues out first.

ocefpaf:
OK. No need to close, just add a `WIP` in the title.

acrosby:
Alright. I noticed that if the time units are not specified in the layers GetCapabilities, then the time control doesn't get populated with the available times, so I want to expose some additional arguments that will allow that to get set manually.

acrosby:
This should get the job done to support datasets that might have a problem with their time units in the GetCaps, also supports multiple time enabled wms layers using a single control

folium/features.py
```diff
@@ -49,14 +49,18 @@ class WmsTileLayer(Layer):
...
+        the WMS request.
 
     For more information see:
     http://leafletjs.com/reference.html#tilelayer-wms
 
     """
     def __init__(self, url, name=None, layers=None, styles=None, fmt=None,
                  transparent=True, version='1.1.1', attr=None, overlay=True,
-                 control=True):
+                 control=True, **kwargs):
```
ocefpaf: :+1:  ideally we should validade those kw args before passing to the render b/c they fail silently. But that can be done in another PR. Let us assume that people adding extra options know what they are doing.

folium/plugins/timestamped_wmstilelayer.py
```diff
@@ -0,0 +1,163 @@
...
+from jinja2 import Template
+
+from branca.element import Figure, JavascriptLink, CssLink
+from ..map import Layer
+from ..features import WmsTileLayer
+
+
+class TimestampedWmsTileLayers(Layer):
+    def __init__(self, data, transition_time=200, loop=False, auto_play=False,
+                 period="P1D", time_interval=False):
```
ocefpaf: In folium we prefer single quotes `'` instead of double `"` to make it clear to ready JSON and Python. (JSON are only double.)

acrosby:
Sure, I can do that--might be a few days.

ocefpaf:
> Sure, I can do that--might be a few days.

Maybe let us go for the "example only" then and I can merge this and open an issue for the rest.

I only have today for "folium work" and I would like to advance as much as possible. (No pressure though I have tons of other PRs and issues here :smile:)

acrosby:
Sure, I was using a notebook to test my changes anyway. I'll just find a public WMS server to point to, and fix the quotes style thing too.

acrosby:
im having some trouble updating this branch now, not sure what I did. I can add the example as an independent PR later.

ocefpaf:
You should be able to do a `git pull` to get the rebased version here.
But yes, adding it later sounds easier.

Issue #494, opened 2016-08-29
How do i customize my own jinja template?


afidegnum:
by default only plain maps are generated, i would like to create a custom jinja template on which the map will be rendered.

any idea?

kolewu:
Unfortunately there is no interface for this and since the templates are now embedded they cannot be easily overwritten.

But you can monkey patch the Figure class that is used internally as parent for the map:

```python
import os.path
from jinja2 import Template
from folium.element import Figure

tmplfname = "/path/to/your/jinja/template.j2"
if os.path.isfile(tmplfname):
    with open(tmplfname) as fo:
        tmpldata = fo.read()
else:
    tmpldata = u"""\
<!DOCTYPE html>
<!-- monkey patched Figure template -->
<head>
    {{this.header.render(**kwargs)}}
</head>
<body>
    {{this.html.render(**kwargs)}}
<script>
    {{this.script.render(**kwargs)}}
</script>
</body>
</html>
"""

Figure._mytemplate = Template(tmpldata)
def myFigureInit(self, *args, **kwargs):
    self.__init_orig__(*args, **kwargs)
    self._template = self._mytemplate
Figure.__init_orig__ = Figure.__init__
Figure.__init__ = myFigureInit
```

But beware: the map-div may interact badly with your template, especially when you are not using bootstrap.

This monkey patch works for the current stable version 0.2.1.

ocefpaf:
I would be open to a PR that implements that "officially." Such PR should target the `branca` module. More specifically https://github.com/python-visualization/branca/blob/master/branca/element.py#L230-L284

rl-utility-man:
what would be involved in invoking the branca figure.to_dict() function on a Folium map?  if we could do so, would that return a dictionary that would make it straight forward and reasonably safe to insert folium map(s) into an appropriately designed custom jinja template of the form:

```
<!DOCTYPE html>
<html>
<head>
<!--custom header components-->
{{ head }}
</head>
<body>
<h1>Map!</h1>
{{ html }}
<p>And here's some text after the graph.</p>
</body>
<script>
{{ script }}
</script>
</html>
```

PR #653, opened 2017-08-09
Allow map css customization

dirkvdb:
I needed to change the rendering of the image overlays, which requires css tweaks.
I therefore added a custom_css argument to the map constructor which applies the provided css rules on the current map id.

e.g.
```
m = folium.Map(crs='EPSG3857', custom_css=['.leaflet-image-layer { image-rendering: pixelated;}'])
```

This causes the image overlay to be renderered without blurring of the pixels, which is expected by gis people.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -88,6 +88,10 @@ class Map(LegacyMap):
         Forces Leaflet to not use hardware-accelerated CSS 3D
         transforms for positioning (which may cause glitches in some
         rare environments) even if they're supported.
+    custom_css : string array, default []
+        Additional css rules to allow customization of map display.
+        Each element in the array wil get prefixed with the current map id
+        so it only applies to this map instance
 
     Returns
     -------
--- a/folium/map.py
+++ b/folium/map.py
@@ -145,6 +145,10 @@ class LegacyMap(MacroElement):
         Forces Leaflet to not use hardware-accelerated CSS 3D
         transforms for positioning (which may cause glitches in some
         rare environments) even if they're supported.
+    custom_css : string array, default []
+        Additional css rules to allow customization of map display.
+        Each element in the array wil get prefixed with the current map id
+        so it only applies to this map instance
 
     Returns
     -------
@@ -171,7 +175,7 @@ class LegacyMap(MacroElement):
                  min_lon=-180, max_lon=180, max_bounds=True,
                  detect_retina=False, crs='EPSG3857', control_scale=False,
                  prefer_canvas=False, no_touch=False, disable_3d=False,
-                 subdomains='abc', png_enabled=False):
+                 subdomains='abc', png_enabled=False, custom_css=[]):
         super(LegacyMap, self).__init__()
         self._name = 'Map'
         self._env = ENV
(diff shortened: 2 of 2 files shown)


folium/folium.py
```diff
@@ -88,6 +88,10 @@ class Map(LegacyMap):
         Forces Leaflet to not use hardware-accelerated CSS 3D
         transforms for positioning (which may cause glitches in some
         rare environments) even if they're supported.
+    custom_css : string array, default []
```
ocefpaf: I wonder if overloading the `Map` class is the best way to do it.
Also, why do you need an `[]` as default? I'd rather use `None`.

folium/map.py
```diff
@@ -227,6 +232,9 @@ def __init__(self, location=None, width='100%', height='100%',
                 left: {{this.left[0]}}{{this.left[1]}};
                 top: {{this.top[0]}}{{this.top[1]}};
                 }
+                {% for rule in this.custom_css %}
```
ocefpaf: If you change `custom_css` to `None` we can something like:

```python
if custom_css:
  <jinja template code>
```

ocefpaf:
> This causes the image overlay to be renderered without blurring of the pixels, which is expected by gis people.

This comment actually make me wonder if we should not make that default instead of adding options to handle it. What do you think @dirkvdb?

dirkvdb:
> I am on the fence on this one so if you can make a case it can help me decide

I understand that it feels like violating the abstractions. But hey, sometimes a little css tweak can make a big difference :-). I think users who use custom css will understand that things can go wrong. In my case the pixelated raster rendering is really required.

>This comment actually make me wonder if we should not make that default instead of adding options to handle it. What do you think @dirkvdb?

I think that it indeed makes sense to have this as the default behaviour. At least it is what my gis colleagues expect from a raster view. The image-rendering css property seams pretty recent however, not all browsers support it at the moment so it is hard to provide the same experience on all browsers. [https://developer.mozilla.org/en/docs/Web/CSS/image-rendering](url)

>PS: can you add a changelog entry, and a simple test case.

I will put in the extra effort if you decide the PR will be accepted :-)
Changing the default to be pixelated is also fine by me, then I don't need the css backdoor. But somehow I feel that the css backdoor will be useful someday.

ocefpaf:
Let me read on https://developer.mozilla.org/en/docs/Web/CSS/image-rendering and I'll get back to you.

At the moment I am mor inclined to add it by default and implement the CSS backdoor in a different way in the future.

ocefpaf:
@dirkvdb after giving this some I would like to:

1. allow for `css` override with a full jinja/augmented jinja templating in branca. See  https://github.com/python-visualization/folium/issues/494#issuecomment-324461161
2. add the option from this PR as the default "now" without the custom templating.

Note that the custom template would allow for much more than just extra `css`.

I am closing this but I welcome a PR to implement `2`.

dirkvdb:
Sounds good. I will try to create a PR for option 2 sometime this week.

ocefpaf:
Awesome thanks!

jrborbars:
Hi, @ocefpaf and @dirkdvb. How this PR will be/was implemented? I have a fresh folium install in python 3.7 and trying to change the default folium templates (in the folder templates) without any success. My goal is to add some JS scripting to make an online view map. Best regards

PR #655, opened 2017-08-16
Update map.py to fix broken unpkg urls

steeznson:
Fixed broken unpkg url

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -28,7 +28,7 @@ ENV = Environment(loader=PackageLoader('folium', 'templates'))
 
 _default_js = [
     ('leaflet',
-     'https://unpkg.com/leaflet@1.1.0/dist/leaflet.js'),
+     'https://unpkg.com/leaflet@1.2.0/dist/leaflet.js'),
     ('jquery',
      'https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js'),
     ('bootstrap',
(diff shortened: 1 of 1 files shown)


MartinKolarik:
Just fyi, this might not fix the problem for everyone - lots of people have been complaining about random 404 and 503 unpkg responses recently (e.g. for me the 1.1.0 link works).

You can always use jsDelivr instead, which works in a similar way but is more reliable: https://cdn.jsdelivr.net/npm/leaflet@1.1.0/dist/leaflet.js

steeznson:
Huh I thought that unpkg might have deleted the old version seeing as 1.2.0 is the latest. Anyway thanks for the new url.

ocefpaf:
@steeznson do you to try @MartinKolarik's new URL but still pointing to `1.1.0` version? (I would like to keep the leaflet update to another PR so we can access what has changed first.)

steeznson:
Yes I think @MartinKolarik's url to Leaflet1.1.0 sounds like a good idea.

ocefpaf:
Thanks! I am reserving today for folium. If you make the changes soon I can merge it today. (No pressure though :wink:)

steeznson:
Ok all done! Thanks for reminding me to change it.

ocefpaf:
We need to also update the template used for testing, but I can do that. Thanks for the PR @steeznson!

Feel free to send another one updating folium to leaflet 1.2.0 :wink:

PR #656, opened 2017-08-19
Added HERE tiles template

deeplook:
This contains only the `attr.txt` and `tiles.txt` files. Not sure if something else is needed (apart from README, etc). In addition to the credentials (`APP_CODE` and `APP_ID`) many more items can be varied in the HERE tiles URL, but I'm not sure if the current design suggests to use multiple folders and names, like in the current `stamenterrain` and `stamentoner` folders, or if additional attributes could/should be rather passed to the `folium.Map()` constructor. (Sure, the user can use the his own tiles URL, anyway.)

### DIFF
--- a/folium/templates/tiles/here/attr.txt
+++ b/folium/templates/tiles/here/attr.txt
@@ -1,0 +1,1 @@
+Map tiles by <a href="https://www.here.com/">HERE Technologies</a>.
--- a/folium/templates/tiles/here/tiles.txt
+++ b/folium/templates/tiles/here/tiles.txt
@@ -1,0 +1,1 @@
+https://{s}.base.maps.api.here.com/maptile/2.1/maptile/newest/normal.day/{z}/{x}/{y}/256/png8?app_id={{ APP_ID }}&app_code={{ APP_CODE }}


ocefpaf:
@deeplook thanks for the PR!

I am sorry to say that after reading HERE website I am unsure if we can use the tiles programmatically even by requiring a key. (I am no lawyer so I may be wrong here.)

Also, we do provide an API to add them directly with very little code, and that won't characterize as programmatically b/c it is always on a case-by-case basis.

Maybe the best thing to here is to add it as an example, using the API that adds external tiles, instead of adding them as an built-in option.

deeplook:
I don't mind if this doesn't get merged, because as you say the existing API makes using HERE tiles easy enough. Otherwise, I'm also not a lawyer, but I'm curious to know what difference you see between HERE and other map providers using some sort of credentials, like Cloudmade and Mapbox?

ocefpaf:
> I don't mind if this doesn't get merged, because as you say the existing API makes using HERE tiles easy enough.

I still encourage you to submit an example to https://github.com/python-visualization/folium_contrib

I am planning to make that repo a live gallery of folium examples.

> Otherwise, I'm also not a lawyer, but I'm curious to know what difference/issues you see between HERE and other map providers using some sort of credentials, like Cloudmade and Mapbox?

I am planning on removing the `Mapbox` tiles code too. The wording there also seems to suggest that we cannot download data programmatically, maybe I am reading this wrong but I prefer to err on the side of simplicity.

To be honest I want to remove most of the tiles and add docs on how to manually add them. That way we are not liable for outdated attribution text, etc.

deeplook:
I see. Thanks for your explanation!

deeplook:
The easiest would be to make a PR with the existing notebook as-is, hoping it encourages others to write a bit more prose. I would number the static PNGs like `here<num>.png`, then. Ok?

ocefpaf:
> The easiest would be to make a PR with the existing notebook as-is, hoping it encourages others to write a bit more prose. I would number the static PNGs like `here<num>.png`, then. Ok?

Sounds like a plan!

PR #674, opened 2017-08-24
absolute imports

ocefpaf:
Just to avoid `ValueError: Attempted relative import in non-package` when hacking into the code locally.


ocefpaf:
I cannot reproduce that notebook failure locally :-/

PR #683, opened 2017-08-25
Re-factor Circle and CircleMarker

### LINKED ISSUE #430: CircleMarker line_color not recognised
It seems since the change from a the function (circlemarker) to a class def (CircleMarker) it has dropped the line_color attribute.

ocefpaf:
This PR is the first of many that:

- re-implements the features using leaflet's defaults
- factor out the parsing of the options so we can re-use (`Path`)
- eliminate the bogus tests that relies on updating the template (new tests are still pending!)

If this works nicely I plan to do the same for all `feature.py`.

Closes #430, #524, and #569


folium/features.py
```diff
@@ -809,95 +809,81 @@ class Circle(Marker):
...
+        If true the circle will be filled.
+    fill_color: str, default to the same as color
         The fill color of the marker in a HTML-compatible format.
-    fill_opacity: float, default 0.6
+    fill_opacity: float, default 0.2
         The fill opacity of the marker, between 0. and 1.
     popup: string or folium.Popup, default None
         Input text or visualization for object.
 
+    See http://leafletjs.com/reference-1.2.0.html#path for more otions.
```
ocefpaf: Instead of overloading the docstring I'll leave only the main options here and refer to the original page for more.

folium/features.py
```diff
@@ -809,95 +809,81 @@ class Circle(Marker):
...
-        )
-        self._name = 'Circle'
-        self.radius = radius
-        self.color = color
-        self.fill_color = fill_color
-        self.fill_opacity = fill_opacity
+    def __init__(self, location, radius=10, popup=None, **kw):
+        super(Circle, self).__init__(_validate_location(location), popup=popup)
+        self._name = 'circle'
+        options = _parse_path(**kw)
```
ocefpaf: I don't want to make this overly complex. Just a simple function to parse the options.
Ideally this function could validade the options as well.

folium/features.py
```diff
@@ -809,95 +809,81 @@ class Circle(Marker):
...
+
+        options.update({'radius': radius})
+        self.options = json.dumps(options, sort_keys=True, indent=2)
```
ocefpaf: First step towards a cleaner HTML. I want to eliminate blank spaces and bad indentations.

folium/features.py
```diff
@@ -809,95 +809,81 @@ class Circle(Marker):
...
                 [{{this.location[0]}},{{this.location[1]}}],
-                {{ this.radius }},
-                {
-                    color: '{{ this.color }}',
-                    fillColor: '{{ this.fill_color }}',
-                    fillOpacity: {{ this.fill_opacity }}
-                    }
-                )
-                .addTo({{this._parent.get_name()}});
+                {{ this.options }}
```
ocefpaf: We should use `json.dumps` everywhere to reduce the amount of code like `'true' if something else 'false'`.

folium/features.py
```diff
@@ -809,95 +809,81 @@ class Circle(Marker):
...
-        use Circle.
-    color: str, default 'black'
+        The radius of the circle in pixels.
+        For setting the radius in meter, use Circle.
+    color: str, default '#3388ff'
         The color of the marker's edge in a HTML-compatible format.
-    weight: int, default 2
-        Stroke weight in pixels
-    fill_color: str, default 'black'
+    fill: bool, default False
```
ocefpaf: That is different from the current defaults! Here I applied leaflet's defaults everywhere.

folium/templates/circle_marker.js
```diff
@@ -1,6 +0,0 @@
-var {{ circle }} = L.circleMarker([{{ lat }}, {{ lon }}], {
```
ocefpaf: First step to eliminate all these render templates.

tests/test_circle.py
```diff
@@ -0,0 +1,146 @@
...
+    }}
+    ).addTo({1});
+    """.format(circle.get_name(), m.get_name())
+
+    rendered = circle._template.module.script(circle)
+    assert rendered.strip().split() == expected_rendered.strip().split()
+    assert circle.get_bounds() == expected_bounds
+    assert json.dumps(circle.to_dict()) == circle.to_json()
+    assert circle.location == [-27.551667, -48.478889]
+    assert circle.options == json.dumps(options, sort_keys=True, indent=2)
```
ocefpaf: We should follow this example to re-factor the tests in `features.py`.

PR #684, opened 2017-08-26
Use pixelated image rendering for image overlays

dirkvdb:
As discusses previously. This changes the default css image-rendering style for image overlays to pixelated.

Tested on firefox and chrome

PR #686, opened 2017-08-26
Finishes #684

ocefpaf:
@dirkvdb I'd appreciate if you could take a look. Not that the `default` is "pixelated."

Check [this](http://nbviewer.jupyter.org/github/ocefpaf/folium/blob/0796237f08165a30b355f38d97b2082fa70c4ca5/examples/GulfStreamImageOverlay.ipynb) notebook for a before/after example.


dirkvdb:
Looks great! It's good you added the option to disable it.

PR #378, opened 2016-03-01
Have added support to simple_marker to allow script to be included in popups

samchorlton:
Have added functionality to branca to support html in popups and have amended map.py and folium.py so that this can be set as a flag.

PR #689, opened 2017-08-26
Simple HTML popup

ocefpaf:
Continues #378

@samchorlton I took a different approach and made HTML popups enable by default like in the early days of `folium`. This makes the most common use easier while still enabling us to render complex templates as popups.


ocefpaf:
Getting close to a release :sweat_smile:

PR #700, opened 2017-08-30
Refactor polyline

ocefpaf:
Similar to #683


ocefpaf:
TODO: write the tests for the new behavior and simplify the docs by using reference the option args they are using and documenting only the unique options.

PR #701, opened 2017-08-31
Custom iconCreateFunction in the constructor

odovad:
Possible to add a custom iconCreateFunction to the constructor.

Please see the example below :

```python
icon_create_function = """
function (cluster) {
    var childCount = cluster.getChildCount();
    var c = ' marker-cluster-small';
    return new L.DivIcon({ html: '<div><span>' + childCount + '</span></div>', className: 'marker-cluster' + c, iconSize: new L.Point(40, 40) });
}
"""
```

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -756,9 +756,10 @@ class MarkerCluster(Layer):
         Adds the layer as an optional overlay (True) or the base layer (False).
     control : bool, default True
         Whether the Layer will be included in LayerControls
-
+    icon_create_function : string, default None
+        Override the default behaviour, making possible to customize markers colors and sizes 
     """
-    def __init__(self, name=None, overlay=True, control=True, icon_create_function=""):
+    def __init__(self, name=None, overlay=True, control=True, icon_create_function=None):
         super(MarkerCluster, self).__init__(name=name, overlay=overlay,
                                             control=control)
         self._name = 'MarkerCluster'
@@ -766,7 +767,7 @@ class MarkerCluster(Layer):
         self._template = Template(u"""
             {% macro script(this, kwargs) %}
             var {{this.get_name()}} = L.markerClusterGroup({
-                {% if this._icon_create_function != "" %}
+                {% if this._icon_create_function %}
                    iconCreateFunction: {{this._icon_create_function}}
                 {% endif %}
             });
--- a/folium/features.py
+++ b/folium/features.py
@@ -758,13 +758,16 @@ class MarkerCluster(Layer):
         Whether the Layer will be included in LayerControls
 
     """
-    def __init__(self, name=None, overlay=True, control=True):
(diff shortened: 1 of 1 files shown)


folium/features.py
```diff
@@ -758,13 +758,18 @@ class MarkerCluster(Layer):
         Whether the Layer will be included in LayerControls
 
     """
-    def __init__(self, name=None, overlay=True, control=True):
+    def __init__(self, name=None, overlay=True, control=True, icon_create_function=""):
```
ocefpaf: please change  `icon_create_function=""`  to `icon_create_function=''` . Sounds silly, but with all the jinja templates represent JSONs it is nice to have single quotes for python code and double for JSON. (Makes it easier to read.)

folium/features.py
```diff
@@ -758,13 +758,18 @@ class MarkerCluster(Layer):
...
         super(MarkerCluster, self).__init__(name=name, overlay=overlay,
                                             control=control)
         self._name = 'MarkerCluster'
+        self._icon_create_function = icon_create_function.strip()
         self._template = Template(u"""
             {% macro script(this, kwargs) %}
-            var {{this.get_name()}} = L.markerClusterGroup();
+            var {{this.get_name()}} = L.markerClusterGroup({
+                {% if this._icon_create_function != "" %}
```
ocefpaf: I believe that `{% if this._icon_create_function %}` would work.
Also, maybe a default to `None` instead of `''`would be more readable.

ocefpaf:
@odovad maybe we could take this opportunity and upgrade to 1.1.0

See https://github.com/Leaflet/Leaflet.markercluster/blob/f175a7cc31322a194ecafc28bbf4fcfe953c476e/CHANGELOG.md#110-2017-08-27

folium/features.py
odovad: Ok, I try to do all you've said.

folium/features.py
odovad: Yes, you're right. It works.

ocefpaf:
@odovad are you done? LGTM!

Can you make and entry in the `CHANGES.txt` file so we can merge this :wink:

odovad:
@filipe I did not have time for the file with the test and the notebook. 
I try to do it as fast as I can 

Envoyé de mon iPhone

ocefpaf:
> I try to do it as fast as I can 

There is no rush :wink: 

> @filipe I did not have time for the file with the test and the notebook.

The notebook is optional and can he added in another PR. I'll add the changelog entry if you don't mind and get this merged.

The reason is b/c I am re-factoring a bunch of stuff and I don't want you to end up having to rebase this.

Thanks a lot for the contribution!

odovad:
👍🏻 Ok I thought I had to write at least the file for the test. 

Envoyé de mon iPhone

ocefpaf:
> at least the file for the test.

That will be nice to have too :wink:

Issue #702, opened 2017-08-31
On The Fly Marker return geolocation?


lockbucket:
I am new to Folium and saw the feature for on-the-fly placement of markers.  Is it possible to return those values in Folium?

What I am trying to do is if a person creates a region defined by markers, to return those values and then run a query against that particular area?

ocefpaf:
That is possible but not implemented. Take a look at ipyleaflet which I believe has that feature.

PR #720, opened 2017-09-12
add Draw plugin

ocefpaf:
@BibMartin I know you are quite busy but I'd appreciate if you take a look at this.
Specially if you think of a way to get the GeoJSON back as a python variable.

Right now if you click on the object you'll get an JS alert with the GeoJSON.

I tried:

```
var command = "coords='" + coords + "'"
var kernel = IPython.notebook.kernel;
kernel.execute(command);
```

To return a `coords` variable, but that seems to no longer work. It seems that they deprecated it and now the only way to do this is via widgets. Not sure...

See http://nbviewer.jupyter.org/github/ocefpaf/folium/blob/Draw/examples/Draw.ipynb

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,17 @@
+0.5.0
+~~~~~
+
+- Added `Draw` plugin (ocefpaf #720)
+- Better handling of URL input (ocefpaf #717)
+- Versioned docs! Visit http://python-visualization.github.io/folium/docs-v{{version}}
+  or simply http://python-visualization.github.io/folium/ for the latest version.
+
+Bug Fixes
+
+- Fix `VideoOverlay` import (ocefpaf #719)
+- Fix `choropleth` docstring (lsetiawan #713)
+- Fix `choropleth` name in `LayerControl` (ocefpaf #493)
+
 0.4.0
 ~~~~~
 - Optional `iconCreateFunction` for `MarkerCluster` to customize the icons (odovad #701)
--- a/examples/Draw.ipynb
+++ b/examples/Draw.ipynb
@@ -1,0 +1,79 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "0.4.0+15.ga36370e\n"
+     ]
+    }
+   ],
+   "source": [
+    "import os\n",
+    "import folium\n",
+    "\n",
+    "\n",
+    "print(folium.__version__)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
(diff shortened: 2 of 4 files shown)


BibMartin:
Hi @ocefpaf ,
The PR looks perfect, and the plugin is awesome.
I'll try to have a look about getting the data back in python ; but I'm a bit pessimistic: you have to tell the client browser to send data back to the server and there's no generic way of doing this.
Anyway, it would be notebook-specific.

ocefpaf:
> Anyway, it would be notebook-specific.

Yeah, that is the goal, but I want to avoid telling people to just use `ipyleaflet`  for that.

We could at least try to create a hook to save a GeoJSON file though... That would be helpful too.

ocefpaf:
BTW I used this in the past with success: http://jakevdp.github.io/blog/2013/06/01/ipython-notebook-javascript-python-communication/

BibMartin:
Yes, this has to be updated but may work. A difficulty lies in the fact that our figure is embedded into an Iframe; meaning that you don't have the same javascript environment as the main notebook. We did this to prevent folium and the notebook's javascript to break eachother...

Another workaround could be to go through the API and use something like:

    curl -X PUT --header 'Cookie: ...' -d '{"content":"someText", "type":"file", "format":"text"}' http://my-jupyter-server/api/contents/foo.txt
but you need to configure the auth cookie properly

ocefpaf:
> Yes, this has to be updated but may work. A difficulty lies in the fact that our figure is embedded into an Iframe; meaning that you don't have the same javascript environment as the main notebook.

Totally forgot about that!

> We did this to prevent folium and the notebook's javascript to break eachother...

We should take a look at how ipyleaflet is doing that. Maybe there is not much interference as we expect.

> Another workaround could be to go through the API and use something like:

Seems too hacky, no?

ocefpaf:
@BibMartin I'll merge this as-is now and we can revisit the Python variable integration later. There are people waiting on this functionality.

BibMartin:
:+1:

Shall we open a new issue for this or let #702 hold the question ?

ocefpaf:
#702 is fine. Thanks!

lockbucket:
I don't know if this is possible, but I was thinking of using this on a flask website.  The plugin works fantastic, but is it possible to load to a database instead of a jupyter notebook?  I think that would be an easier commit than a standard jupyter notebook?  I did a google search for pulling this data and found this bookmark which allows a download of the coordinates?

https://bl.ocks.org/danswick/d30c44b081be31aea483

ocefpaf:
> but is it possible to load to a database instead of a jupyter notebook?

Not sure what you mean. I am terrible with databases.

> I did a google search for pulling this data and found this bookmark which allows a download of the coordinates?

That looks awesome! I'll take a look if I can add that functionality.

Better than the current copy-n-paste and while `ipyleaflet` requires global variables this would require a known file name to deliver a similar functionality with the advantaged of having the data outside the notebook JSON.

lockbucket:
RE: Database

I guess what I am trying to figure out is illustrated here:
http://flask.pocoo.org/docs/0.12/patterns/jquery/

when a person clicks on the layer jquery will then run a python command to update a database.  I will try looking into this. 

Thank you for such a wonderful plugin!

ocefpaf:
PR #727 added the export button option. Not exactly what you want but that is a start in you want to create your own plugin based on that.

PR #727, opened 2017-09-20
export draw

ocefpaf:
Add export drawing layer.


ocefpaf:
See http://nbviewer.jupyter.org/gist/ocefpaf/288e713e05e3dca3d5dfbce129993294

folium_geojson_export

PR #756, opened 2017-10-23
TileLayer and WMSTileLayer are in __all__,

sanga:
therefore they should be imported.

Or if this was the wrong fix then they should be removed from `__all__` (note: I haven't tested this at all it's just a driveby fix)


sanga:
Please squash-merge as I didn't noticed when uncommenting that the uncommented line actaully had a syntax typo.

ocefpaf:
Thanks @sanga!

PR #764, opened 2017-10-31
Add a `time` named argument to `to_png`

hunterowens:
Since it might take longer than 3 seconds to render the map, this adds a time named arg that allows you to delay the screenshot.


ocefpaf:
> Since it might take longer than 3 seconds to render the map, this adds a time named arg that allows you to delay the screenshot.

Makes perfect sense. Thanks!

Issue #606, opened 2017-03-13
Show/Hide Selected Overlays (was question re: the layer checkbox)


cloud05:
Good day! I was reading one of @ocefpaf 's post and stumbled upon this. https://ocefpaf.github.io/python4oceanographers/blog/2015/03/23/wms_layers/ . How can I make a layer group unchecked on load like the example given on the post? Thanks!

ocefpaf:
That is a bug we introduced in the latest version that needs fixing. (PRs welcomed :wink:)

cloud05:
Can we have this as a feature? Haha Add a parameter to a layer group that will determine if it is checked/unchecked on load? Like here, on the Programmatic Display section: https://rstudio.github.io/leaflet/showhide.html.  I'll try to tinker around this. Thanks!

Conengmo:
I tried two different implementations for a default visibility setting for overlays:

1. Add this functionality to `class LayerControl` by accepting a list with names of overlays to hide on load. This works fine, but you have to remember which layers you want to hide until adding layer control.

2. Add a `hide` parameter to `class Layer`. Then each child class would also need to accept this extra parameter. 

Both end up being used in the template in `class LayerControl`. I add a `layer_object.remove();` line for each layer that should be hidden by default.

I'm new to this scene and didn't find any guidelines so please advice on how to proceed.

ocefpaf:
I am more inclined to solution number 2.

Conengmo:
Allright, I'll make a WIP pull request later for you to see. I'll wait with adding this to all the classes until we're sure this is the way to go.

Conengmo:
I was going through the issues and noticed this functionality is the same as asked for in #571.

Issue #571, opened 2016-11-09
Select visible layer from a feature group


ispmarin:
I´m using `FeatureGroup` and `LayerControl` to set a bunch of layers. When the map is drawn all layers are visible. Is there a way to show only one layer on rendering and keep `overlay == True`? 

The problem with using `overlay==False` is when more than one tile server is added: all options are the same, and now both the feature groups and the tile servers are options.

ocefpaf:
I am not sure I follow. Can you try to put that as an example, or better yet, a PR :wink:

battez:
Hi seconding this. A way to do it in Leaflet is given here:
[Toggle Initial Layer Visibiliity](http://stackoverflow.com/questions/27056581/how-to-toggle-initial-layer-visibility-in-leaflet-js
)
`var map = L.map('map').setView([52.814172, -2.079479], 9);

L.tileLayer('http://{s}.tile.stamen.com/toner/{z}/{x}/{y}.jpg', {
maxZoom: 18,
attribution: 'Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>.'
}).addTo(map);

// Left out layer-declarations, they are HUGE

anti_soc.addTo(map); // here it is

var overlayMaps = {
    "Anti-social behaviour" :anti_soc,
    "Criminal damage and arson": crim_dam,
    "Burglary": burg, 
    "Violence and sexual offences": viol,
    "Other theft": other_theft,
    "Vehicle crime": veh_crime,
    "Shoplifting": shoplift,
    "Public order": pub_ord, 
    "Robbery": robb,
    "Bicycle theft": bikes,
    "Drugs": drugs,
    "Theft from the person": theft_person,
    "Other crime": other_crime,
    "Possession of weapons": weap
};

L.control.layers(null,overlayMaps).addTo(map);`

Conengmo:
This has been added in #772

PR #772, opened 2017-11-22
Untoggle overlays and additional base layers

Conengmo:
I worked on solving issue #606 and #571. I added functionality in the `LayerControl` class to untoggle overlays that have `hidden=True`. While I was there I simplified the checks in render(). If an item is an instance of `Layer` it will always have the `overlay` and `control` attributes.

I added the `hide` parameter to some classes to test it out, but not to all children of `Layer` yet.

What do you guys think, is this a good way to do this?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,7 @@
 - Added `tooltip` support to `Marker`s (ocefpaf #724)
 - Added `tooltip` support to all vector layers (ocefpaf #722)
 - Added `TimeSliderChoropleth` plugin (halfdanrump #736)
+- Added `show` parameter to choose which overlays to show on opening (conengmo #772)
 
 API changes
 
--- a/folium/features.py
+++ b/folium/features.py
@@ -317,6 +317,8 @@ class GeoJson(Layer):
         Adds the layer as an optional overlay (True) or the base layer (False).
     control : bool, default True
         Whether the Layer will be included in LayerControls
+    hide: bool, default False
+        Whether the layer will be hidden by default, if it is an overlay.
     smooth_factor: float, default None
         How much to simplify the polyline on each zoom level. More means
         better performance and smoother look, and less means more accurate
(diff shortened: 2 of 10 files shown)


ocefpaf:
> What do you guys think, is this a good way to do this?

Yep. That fits `folium` perfectly. Thanks!

Can you:

- add the keyword to everything else that inherits from `Layer`;
- change the nome from `hide` to... Maybe `toggle`? I accept other suggestions but hide seems like something that won't be displayed ever;
- add an entry to the changelog/

Conengmo:
I added the `show` parameter to all classes that inherit from `Layer`. At least, I hope I got them all :)
`show` seemed a good choice, I saw it's also being used in #778.

I noticed `Heatmap` and `HeatMapWithTime` inherited from `TileLayer`, that didn't really make sense to me. They also didn't use anything from `TileLayer`. So for consistency I changed both classed to inherit from `Layer`.

I also did some work on the use of `name`, `overlay` and `control` in `__init__` statements and docstrings in files I touched. It should be fairly consistent now.

I didn't test all code I touched. I did test creating some maps with `GeoJson` objects, with `Featuregroup`s, `MarkerCluster`s and `HeatMap`s, and everything seemed to work alright.

Let me know if there's something you want different or added.

Conengmo:
I noticed that [according to Leaflet](http://leafletjs.com/reference-1.2.0.html#control-layers-l-control-layers), any additional base layers should not be added to the map. In Folium all base layers are added to the map. As a result the last base layer is selected in layer control, while the first base layer is displayed.

Fixing this by not adding additional base layers to the map is hard, since we then need to check this in every class that inherits from `Layer`. Removing the additional base layers in `LayerControl` also works, so I added it here.

ocefpaf:
Perfect! Thanks @Conengmo!!

carlospendola:
thanks master! @Conengmo

Issue #210, opened 2015-10-04
permanent popups


methodds:
Hi, I have a question regarding popups for markers. Is it possible to use the markers as a kind of label, such that the text for popups is already visible without having to click on the marker? I would like to use this functionality to label some cities on a map.

andrewgiessel:
Great idea. I bet it's possible. Folium uses leaflet under the hood, so I'd
check out the leaflet API (or plugins) and if it's there we can implement
it. Link us back with anything you find!

methodds:
Thanks for the fast response! Apparently there is a simple plugin for this functionality in leaflet:
https://github.com/Leaflet/Leaflet.label
Can you work with this?

themiurgo:
@methodds @raidoz Would any of you be interested in implementing it? It should be fairly easy to contribute it as a new class heavily similar to Popup.

cgarciaf:
Can you pls show some code line ass un example i need to use this feature....

ocefpaf:
@cgarciaf take a look at the plugins folder in `folium`.

cgarciaf:
Could you pls drive me a bit more i looked into plugin but i am lost...

ocefpaf:
> Could you pls drive me a bit more i looked into plugin but i am lost...

Sorry, but not right now. Folium is maintained by  the efforts of brave volunteers with demanding day jobs :crying_cat_face: 

As soon as I get some free time I will put a skeleton PR together and ping you.

cgarciaf:
You all have me recognition that before hand. Looking fwrd that skeleton, i am lost on the usage.. Thanks
C

pcohen89:
@cgarciaf did anyone ever create some mechanism to add static text (or always open popup text) to folium maps?

dunguyen84:
Dear all,

https://github.com/Leaflet/Leaflet.label   was promising, and after figuring out how to plot the polygon (somehow the methods and the tutorial are not exactly matching anymore), I would love to add the mouse over text poping up. 
I saw the Leaflet.label package but do not know how to integrate in my code. Nor could I add a popup as for folium.Polyline or folium.CircleMarker
I can share my bit of code if someone is willing to help me.

Many thanks in advance, and thank you for folium

ocefpaf:
> I can share my bit of code if someone is willing to help me.

If you can try to do that in the form of a PR I can take a look.

jwhendy:
[...]

PR #778, opened 2017-11-25
Allow popups to be open on page load

jwhendy:
This is a first stab at #210 , but should probably have some input from the community.

Currently, this creates a new argument, `folium.Popup(..., default_open=False)` which utilizes [`autoClose` option](http://leafletjs.com/reference-1.0.2.html#popup-autoclose) from `L.popup`. This was found on this [SO question](https://stackoverflow.com/questions/9047931/leaflet-js-open-all-popup-bubbles-on-page-load) in which a couple answers cited `autoClose: false`.

This [issue](https://github.com/Leaflet/Leaflet/issues/5719) is also enlightening, as there are some other nuanced arguments that control the behavior we might want to consider in `folium`. For example, the `autoClose` docs say:

>Set it to false if you want to override the default behavior of the popup closing when another popup is opened.

However, the popup will still close when you click another popup because you're also considered to be clicking on the *map* itself. Currently this PR assumes that if you want them to be open on load, you don't want them to close when clicking another, so [`closeOnClick`](http://leafletjs.com/reference-1.0.2.html#popup-closeonclick) is also applied if `default_open=True` is used.

Lastly, there's an argument to `L.map()` that could also be relevant, [`closePopupOnClick`](http://leafletjs.com/reference-1.0.2.html#map-closepopuponclick).

Some items I think should be reviewed:

- [x] the variable name, `default_open`
- [x] is the current approach sane with the inserted if statement (e.g. `{% if this.default_open %}, autoClose: false, closeOnClick: false{% endif %}});`)?
- [x] should other variables be added in order to allow users more granularity, giving access to `autoClose`, `closeOnClick`, and `closePopupOnClick`?
- [x] is `folium.Popup` the right place to add this, or does it make sense any higher up in the chain like `folium.Marker` or `folium.Map` (leaning toward it being correct as it is, which maps `leaflet` options directly to their `folium` analogs).

Anything else? I'm newer to `python` and know nearly zilch of `javascript`, so comments/feedback are welcome!


ocefpaf:
> I'm newer to python and know nearly zilch of javascript, so comments/feedback are welcome!

Everything looks good to go! Thanks!! Can you:

- add a note on the docstring about the new option;
- add an entry to the changelog file;
- add a test to ensure future changes won't break this.

Regarding your questions:

> the variable name, `default_open`

I guess we can keep it simple and just name it `open`.

> is the current approach sane with the inserted if statement (e.g. `{% if this.default_open %}, autoClose: false, closeOnClick: false{% endif %}});`)?

I like it. Simple and "hackable." I don't really want to wrap everything leaflet can do with popups b/c I am not sure the complexity is worth it. However, this extra keyword makes sense to me.

> should other variables be added in order to allow users more granularity, giving access to `autoClose`, `closeOnClick`, and `closePopupOnClick`?

This is what I am on the fence. I don't really want to make it overly complex. open by default is very useful, the other not so much. I can be convinced otherwise but I don't think it is worth investing in that.

> is 'folium.Popup` the right place to add this, or does it make sense any higher up in the chain like `folium.Marker` or `folium.Map` (leaning toward it being correct as it is, which maps `leaflet` options directly to their `folium` analogs).

I may be wrong but `Popup` does not inherit from `Maker`, so we should either re-factor to fix that or we should implement this functionality on both. Not sure about `folium.Map`, the init class there is already overloaded with options.

jwhendy:
@ocefpaf thanks for the feedback!

- [x] `default_open` is now `open` (does it bother you that `open` triggers syntax highlighting since it's a built-in?)
- [x] docstring added:
```
    open: bool, default False
        True renders the popup open on page load
```
- [ ] wrestling with what the test should be. I've only done one test before and it was checking function output; not sure what it means in this case? Would I just be making sure that I get this if `open=True` is passed?
```
var name = L.popup({maxWidth: 'max_width'
    , autoClose: false, closeOnClick: false});
```
I think what you're getting after is that if someone changes the template, maybe my `if` stuff gets mucked up. I'm not sure on a flexible way to check, though. Maybe someone puts an option between `max_width` and `open`; I'm trying to think of a flexible way not to break the test in that case.

>This is what I am on the fence. I don't really want to make it overly complex. open by default is very useful, the other not so much. I can be convinced otherwise but I don't think it is worth investing in that.

Sounds good to me, and I think `closePopupOnClick` from `L.map` is not that useful. It's just static labels that can't be closed. So, it's down to the other two.

- [ ] Can you confirm, then, that we should just bring in `closeOnClick: false` with the choice to render open on page load? You can compare the behavior of these to see the effect:
  - [this one](https://jsfiddle.net/tfz8c5en/) only uses `autoClose`. Click both pins to close them. Now click the left, then the right, back and forth. Each time you open the left, the right will close. After the initial load in which they're open, I think the behavior is undesirable.
  - [this one](https://jsfiddle.net/oyoz60en/) works, in my opinion, how people would expect. Popups only close when they themselves clicked, not when anything else happens.

Once we confirm on the few remaining items above, I'll re-push a [hopefully] final commit!

ocefpaf:
> - [x] `default_open` is now `open` (does it bother you that `open` triggers syntax highlighting since it's a built-in?)

Yes it does. Sorry. We should not shadow builtin and I an terrible with names. Maybe we should name it `toggle`?

> - [ ] wrestling with what the test should be. I've only done one test before and it was checking function output; not sure what it means in this case? Would I just be making sure that I get this if `open=True` is passed?

Yep. Unfortunately that is not much we can do beyond that. Still that kind of test help a lot when adding other features and ensuring that we don't break this one.

I'll get to the rest of your PR and message later. I have to go back to the day job now :grimacing:

jwhendy:
Taking a *break* from my day job :) 

> Maybe we should name it `toggle`

I thought about `open_on_load` initially, which actually struck me as pretty straightforward/explanatory. `toggle` is also pretty reasonable; it just doesn't convey that we're toggling them all *open*. `render_open`, `make_open`, `show_open`, or maybe just `show`? I actually like that... `show`. Passing true means show it by default.

>Still that kind of test help a lot when adding other features and ensuring that we don't break this one.

Sounds good. I'm a little swamped now that the weekend is over and I have to actually do work again. I'm probably good on input until I figure out how to write a test (and I'll just look at some of the other tests for ideas). Thanks!

jwhendy:
Update. Per discussion on #784 , I rather like the new idea that's come up to split the open on load and clicking behavior into separate options. I also like the names. The proposal:

- `show`: `autoClose` and `openPopup()`. Popups will start open on page load, but toggling one of them, or simply clicking the map will make them go away. That'd be [this behavior](https://jsfiddle.net/de8d1qux/1/)
- `sticky`: `autoClose` and `closeOnClick`. Clicking the map or other popups will not close the popup. That would be [this behavior](https://jsfiddle.net/upwv5ayw/1/)

I just made these changes and things seem to work nicely. Here's a [jupyter notebook](https://nbviewer.jupyter.org/gist/jwhendy/653c6bb9002e6edb291b932d2a9e8ac4) I used to run through the combos (to try you'll obviously need to install my branch).

Conengmo:
Hi @jwhendy, I want to continue on working to merge this PR. I think it's great, really simple and effective. Do you want to help with two final things?

* Add a line to the changelog
* Write a test. I think you already thought about this before. It's okay if it's 'just' a rigid check if the option got added, most folium tests are that way.

jwhendy:
I'll take a look at the tests again. It's been a while so it'll take me a little to look at some others for a good example. Chanelog seems easy enough :)

jwhendy:
I'll take a look at the tests again. It's been a while so it'll take me a little to look at some others for a good example. Chanelog seems easy enough :)

jwhendy:
Alright... finally started looking at tests. Can you point me in the right direction? I *think* this looks like it could fit in `test_map.py` and I'd be making sure that the `html` took in the popup code. That said, I'm a bit confused by the tests in general. [This test](https://github.com/python-visualization/folium/blob/master/tests/test_folium.py#L290) sounds like it makes sure custom icons work, but it only appears to check the bounds?

Conengmo:
I think you got the general idea of it: most tests check if the expected javascript/html is indeed in the rendered output. The current popup tests only check the inside html of the popup. Most other tests render a complete map and check if an expected snippet is included. 

Here is a suggestion. I tried to render the popup, so the output can be compared with an expected string. Note that it needs a parent, that's why I added a feature group (can be anything).
```
def test_popup_sticky():
    fg = FeatureGroup()
    popup = Popup('Some text.', sticky=True).add_to(fg)
    rendered = popup._template.render(this=popup, kwargs={})
    expected = u"""some javascript with the same name inserted""".format(popup.get_name())
    assert rendered == expected
```

There have been some changes to master, so the conflicts in this branch have to be resolved.

If you need help let me know :)

jwhendy:
@Conengmo many thanks for the pointer. I have my fingers crossed that the `_template` attribute (method?) was what I was likely looking for. In looking at the available elements via `dir()`, I was not seeing how to get at the html itself. This is most helpful.

I'll revisit soon!

jwhendy:
Pretty sure I'm rebased correctly on top of `master`, so that's one part resolved.

I'm still feeling quite dense on this test thing... from your example code, what would `rendered` be expected to contain? I get an empty string. I was assuming I'd get the raw `js` that gets injected into the page and then manually enter in what that should be. 

```
from folium.map import Popup, FeatureGroup

def test_popup_sticky():
    fg = FeatureGroup()
    popup = Popup('Some text.', sticky=True).add_to(fg)
    rendered = popup._template.render(this=popup, kwargs={})
    expected = u"""some javascript with the same name inserted""".format(popup.get_name())
    return rendered
    
test_popup_sticky()
>>> ''
```

jwhendy:
I think I'm good, or pretty darn close at this point. Sorry that the git commit history is pretty gross...

- [x] fixed conflicts with `master`
- [ ] tests added; is there a better way to get at the `html` id? I'm using `list(popup.html._children.items())[0][0]`. It's in the `rendered` javascript, so I needed some way to pass it to into `expected`
- [ ] added API change to changelog under 0.6.0. Let me know if that's not the right place.
- [x] the `stickler` thing appears new... I need the whitespace it's complaining about to make the templates work.

Conengmo:
Great, seems like we're almost there. Don't worry about the commit history, we'll squash it. 

> is there a better way to get at the html id?

What are doing seems fine. You can also directly get the keys instead of the items:

`list(popup.html._children.keys())[0]`

> added API change to changelog under 0.6.0

It's a new feature, not an API change that users will have to deal with. So you can add it in the main list.

> the stickler thing appears new... I need the whitespace it's complaining about to make the templates work.

Yeah that one's a bit annoying isn't it. We'll have to fix these lint errors though, otherwise the code will become a mess. What you have to do is make sure that there aren't any spaces on your blank lines. You don't have to add another blank line, just remove the spaces from it. For example in map.py there are now two blank lines before the init of Popup, remove one and make sure there are no spaces on the remaining line.

Finally, I have a suggestion for your tests. Right now the spaces and newlines have to be exactly the same in comparing rendered and expected. You can make it more robust by removing whitespaces and newlines before comparing. You could copy over the `_normalize()` function from test_vector_layers.py to test_map.py and normalize rendered and expected before comparing.

jwhendy:
@Conengmo All sounds great. I'll have this done tonight.

Appreciate the tips on the test (such a great idea and newlines/spaces *killed* me trying to get right!) and where to add the changelog bit.

I'm not surprised there were accidental spaces an will fix those. For this:

>For example in map.py there are now two blank lines before the init of Popup...

I actually thought `def`s should be preceded by two blank lines? I just re-educated myself with some PEP-8 action and see that's top level functions and classes... within a class it's just one. Oops!

jwhendy:
I think we're good now! 

- I modified the test slightly, using `''.join(rendered.split())`. This is nice as the line breaks don't even matter that way. Hope that's alright.
- I updated with named `format()` args, which was a nice touch I saw in a different test.
- change log entry moved, whitespace fixed, and updated to use `keys()`

---

Aaannnnd, while I was re-reading your comments for maybe the *fourth* time... how the heck did I miss this staring me in the face:

>First of all, it seems that the `openPopup()` call got lost.

Indeed it did! I blame rebasing on `master` but am not sure how! I guess that's pretty important. I've got that back in, updated the `_show` test to account for it, and also re-ran through all of [these](https://nbviewer.jupyter.org/gist/jwhendy/653c6bb9002e6edb291b932d2a9e8ac4) tests to make sure the map behavior is as-desired. I think we're *actually* good now!

Conengmo:
Merged! Congrats @jwhendy and thanks for all your efforts!

jwhendy:
Woo-hoo! And thanks to you for your patience/coaching! I learned quite a few things.

Conengmo:
Good to hear. And remember you're always welcome to pick up another issue, or contribute by commenting/reviewing other stuff.

phao5814:
Hey guys, any idea why I can't seem to use the new properties yet? Is it because it hasn't been released yet?
I'm getting the following error when using "show" as one of the properties in folium.Popup().

> TypeError: __init__() got an unexpected keyword argument 'show'

Conengmo:
Indeed it hasn't been released yet to Pypi. So if you're on version 0.5 it won't work. Try installing it from GitHub.

PR #791, opened 2017-12-07
Set min_zoom default to Leaflet value (0 i.o. 1)

Conengmo:
I noticed the default value for the min_zoom parameter is set to 1, while [in Leaflet it is 0](http://leafletjs.com/reference-1.2.0.html#tilelayer-minzoom). I couldn't find any documentation on this, so I assume it's a (tiny) mistake. 

I can imagine that users don't want the `zoom_start` to change from 1 to 0, so I set it to 1 instead of `min_zoom`. If `min_zoom` is higher than 1, Leaflet will automatically change the starting zoom level.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -101,8 +101,10 @@ class Map(MacroElement):
         pass a custom URL or pass `None` to create a map without tiles.
     API_key: str, default None
         API key for Cloudmade or Mapbox tiles.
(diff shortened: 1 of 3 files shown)


ocefpaf:
LGTM (and this one is an easy merge :wink:)

Thanks @Conengmo! I'll try to find some time to review your other PRs ASAP.

Issue #789, opened 2017-12-06
native zoom option on tiles missing?


AndrewAnnex:
```python
opm_attr='<a href="https://github.com/openplanetary/opm/wiki/OPM-Basemaps" target="blank">OpenPlanetaryMap</a>'

m = folium.Map(attr=f'NASA/MOLA |{opm_attr}', max_zoom=5, 
           tiles='https://s3-eu-west-1.amazonaws.com/whereonmars.cartodb.net/mola-color/{z}/{x}/{-y}.png')

m

```
#### Problem description

I want to be able to set the native zoom option in leaflet for a tile layer but I don't see a way to do it. I currently set the max zoom but If I disable it and zoom in the tiles are not interpolated and I am left with gray tiles.

#### Expected Output

I would expect to be able to zoom in past whatever native limit there is to a tile, sure it would be interpolated but that is ok for some uses.

#### Output of ``folium.__version__``
'0.5.0'

Conengmo:
I reproduced your problem and may have found what causes it. (Pretty cool map by the way!)

If I get to zoom level 7, the tileserver responds with HTTP status 403 Forbidden for each tile. So Leaflet expects to receive tiles, is unable to get them, and then displays grey tiles. 

The same happens if you use OpenStreetMaps and go past zoom level 18 or 19 (I think), you get a HTTP status 404 Not Found on each tile.

A solution is to provide a `maxNativeZoom` option to the `tileLayer` function. It tells Leaflet to not retrieve tiles for zoom levels beyond this value. This is not implemented in Folium, so you will have to do it in the javascript that is created by Folium. For example:
```
var tile_layer_1234 = L.tileLayer(
                'https://s3-eu-west-1.amazonaws.com/whereonmars.cartodb.net/mola-color/{z}/{x}/{-y}.png',
                {
  "attribution": "NASA/MOLA |<a href=\"https://github.com/openplanetary/opm/wiki/OPM-Basemaps\" target=\"blank\">OpenPlanetaryMap</a>",
  "maxNativeZoom": 6,
  "maxZoom": 18,
} ).addTo(map_1234);
```

AndrewAnnex:
It seems that would be the solution, additionally one could expose additional options like this via a kwargs parameter for greater flexibility.

ocefpaf:
> I reproduced your problem, pretty cool map!

[...]

PR #792, opened 2017-12-08
Tilesets native zoom

Conengmo:
I worked a bit on adding a parameter for the maximum native zoom levels of a tileset, based on #789.

The simplest implementation is just accepting the keyword and adding it to the options of a `TileLayer`. Since we don't know a reasonable default value, it shouldn't be added if it's `None`. I think up to here this is an uncontroversial addition.

Then I worked on finding and adding the maximum native zoom level of the tilesets mentioned in the `TileLayer` docstring. If these are used automatically we prevent users from encountering grayed out tiles.

For most of these tilesets you can check the maximum native zoom level automatically. So I wrote a test that verifies the stored maximum native zoom levels. The tileservers will return a HTTP 403 or 404 when requesting a tile with an unsupported zoom level. Except for the Stamen ones, I didn't find any logic there. It will return tiles for zoom levels up to 30, but will randomly return a HTTP 503 for some reason.

I'm not sure if these values should be hard-coded in `raster_layers.py`. But the tiles templates also didn't seem to fit right. What I was thinking now was to make the tiles folder into a single JSON file, where each tileset is an entry with an url template, attribution and maximum native zoom level.

I'm also not sure about the test. If the test fails because a tileserver changed, it may confuse a contributor who doesn't know about it.

- [x] Should folium fill in the maximum native zoom value for known tilesets?
- [x] Where should these values be stored?
- [x] Where should the script to find/verify these values be located, or should it be dismissed?

### DIFF
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -36,7 +36,7 @@ class TileLayer(Layer):
             - "Mapbox Bright", "Mapbox Control Room" (Limited zoom)
             - "Cloudmade" (Must pass API key)
             - "Mapbox" (Must pass API key)
-            
+
         You can pass a custom tileset to Folium by passing a Leaflet-style
         URL to the tiles parameter: ``http://{s}.yourtiles.com/{z}/{x}/{y}.png``
         You must then also provide attribution, use the `attr` keyword.
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -47,7 +47,6 @@ class TileLayer(Layer):
     max_native_zoom: int, default None
         The highest zoom level at which the tile server can provide tiles.
         If provided you can zoom in past this level. Else tiles will turn grey.
-        For the listed free tileset providers it will be set automatically.
     attr: string, default None
         Map tile attribution; only required if passing custom tile URL.
     API_key: str, default None
@@ -78,6 +77,13 @@ class TileLayer(Layer):
         self._name = 'TileLayer'
         self._env = ENV
 
+        options = {'minZoom': min_zoom, 'maxZoom': max_zoom, 'noWrap': no_wrap,
+                   'attribution': attr, 'subdomains': subdomains,
+                   'detectRetina': detect_retina}
+        if max_native_zoom:
+            options['maxNativeZoom'] = max_native_zoom
+        self.options = json.dumps(options, sort_keys=True, indent=2)
+
         tiles_flat = ''.join(tiles.lower().strip().split())
         if tiles_flat in ('cloudmade', 'mapbox') and not API_key:
             raise ValueError('You must pass an API key if using Cloudmade'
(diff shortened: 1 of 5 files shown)


ocefpaf:
> Should folium fill in the maximum native zoom value for known tilesets?

I don't think we should hardcode known values. The grey tiles are "expected" in the slippy mapping world when you go over the max zoom. People who want a "nice" zoom should do that explicitly.

> Where should these values be stored?

We can consolidate the tile templates, this is a good idea, but I don't think we should provide the maz values.

> Where should the script to find/verify these values be located, or should it be dismissed?

Dismissed. Having the option to add the value manually is good enough IMO.

Conengmo:
Thanks for the quick feedback! I can agree with sticking to manually adding a value. The whole storing and verifying of the values is a bit messy, though interesting to investigate.  I'll revert the changes past the keyword part.

folium/raster_layers.py
```diff
@@ -73,33 +76,29 @@ def __init__(self, tiles='OpenStreetMap', min_zoom=0, max_zoom=18,
...
-            'noWrap': no_wrap,
-            'attribution': attr,
-            'subdomains': subdomains,
-            'detectRetina': detect_retina,
-        }
+        options = {'minZoom': min_zoom, 'maxZoom': max_zoom, 'noWrap': no_wrap,
+                   'attribution': attr, 'subdomains': subdomains,
+                   'detectRetina': detect_retina}
+        if max_native_zoom:
+            options['maxNativeZoom'] = max_native_zoom
```
ocefpaf: This is fine but I am 99% sure you can just pass the option and, if `None`, the `json.dumps` will convert to `null` and everything will work as expected on the JS side. Then we'll have 1 new line in the options instead of the if clause.

folium/raster_layers.py
```diff
@@ -73,33 +76,29 @@ def __init__(self, tiles='OpenStreetMap', min_zoom=0, max_zoom=18,
...
-            'minZoom': min_zoom,
-            'maxZoom': max_zoom,
-            'noWrap': no_wrap,
-            'attribution': attr,
-            'subdomains': subdomains,
-            'detectRetina': detect_retina,
-        }
+        options = {'minZoom': min_zoom, 'maxZoom': max_zoom, 'noWrap': no_wrap,
+                   'attribution': attr, 'subdomains': subdomains,
+                   'detectRetina': detect_retina}
```
ocefpaf: The 1 option per line is by design, for readability. Do you mind reverting that change?

Conengmo: Sure no problem

folium/raster_layers.py
Conengmo: Unfortunately the map won't load with `"maxNativeZoom": null`. I understand you don't want the extra if clause. I solved it by using `'maxNativeZoom': max_native_zoom or max_zoom`.

ocefpaf: > Unfortunately the map won't load with `"maxNativeZoom": null`.

Ah! I did not test that. Thanks!

> I understand you don't want the extra if clause.

Actually the if-clause was OK in that case.

> I solved it by using `'maxNativeZoom': max_native_zoom or max_zoom`.

That is fine too. Thanks!

ocefpaf:
Thanks @Conengmo! The failure is b/c of our flaky tests that compare templates and results. We need to add the extra option there. But you did a lot already and I can solve this while reviewing the other PRs. Thanks!

Issue #788, opened 2017-12-06
Guidance on documentation


Conengmo:
There is documentation and examples in multiple places:
- There is the [docs folder](https://github.com/python-visualization/folium/tree/master/docs), which is build and shown on [github.io](http://python-visualization.github.io/folium/docs-master/index.html). It contain an installation guide, quick start examples and the API documentation. 
- Then there is the [examples](https://github.com/python-visualization/folium/tree/master/examples) folder, which is best viewed online with [Jupyter nbviewer](http://nbviewer.jupyter.org/github/python-visualization/folium/tree/master/examples/). 
- And there is a [folium_contrib](https://github.com/python-visualization/folium_contrib) repository, which contains a few user contributed examples. These are also best viewed with [Jupyter nbviewer](https://nbviewer.jupyter.org/github/python-visualization/folium_contrib/tree/master/notebooks/)

@ocefpaf can you share your plan for the structure of the documentation? I can spend some time on this, but I want to do it right :)

ocefpaf:
> @ocefpaf can you share your plan for the structure of the documentation?

To be honest I don't have a clear plan :grimacing: 

>  I can spend some time on this, but I want to do it right :)

That would be awesome!

Right now examples is a mixture of tests and demos I would like to disambiguate that.
Here is what I wanted to do some time ago:

- move most of the notebooks in the example folder to `python-visualization/folium_contrib`
- create a gallery webpage there, with the same infrastructure from `python-visualization/folium`, so we can either link to the `HTML` pages or to `nbviewer`
- adapt some key notebooks from `examples` to be tests and renamed that folder to make it "invisible" to users (users should use the gallery as examples and not the tests)

I am not sure you want to tackle that, it is kind of boring and it would demand some knowledge of which notebook we need as test and which we don't. Maybe a review of the current docs would be less boring?

Conengmo:
Thanks for explaining! I'll stay out of the examples then and look into the 'regular' documentation.

By the way, maybe it's a good idea to put the links to both of the nbviewer pages in the readme. Until you've created the full folium_contrib webpage. I know when I started using Folium it took me a while to sort out what was where, so it might help new users.

ocefpaf:
> By the way, maybe it's a good idea to put the links to both of the nbviewer pages in the readme.

Yep. We should do that!

AlJohri:
[...]

PR #794, opened 2017-12-09
List all docs in readme

Conengmo:
For new users it's probably good to list all available documentation and galleries in the readme. See #788. So I added links to the two example galleries. 

If you search for 'Folium documentation', the first hit is the Read the Docs page. That is documentation for 0.2.0, so that may confuse people. (It confused me at first :p ) Therefore I also included a line on that.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,10 +15,9 @@ folium
 Python Data, Leaflet.js Maps
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -38,10 +37,18 @@ Documentation
 
 http://python-visualization.github.io/folium/docs-master/
 
+Note that the documentation on Read the Docs is for the deprecated version
```
ocefpaf: Unfortunately I do not own the readthedocs account to remove that, so we need this note. Thanks!

Issue #793, opened 2017-12-08
folium.choropleth() does not use pandas series as key column correctly


krinsman:
**Set-up:** To give some idea of the set-up, I have a 2-column pandas DataFrame, called `german_cars_price`. The first column of this dataframe is labeled `state` and denotes the relevant German federal state. The second column is labeled `price` and refers to the average used car sales price in that German federal state from a sample of ads on eBay.

I also have a GeoJSON file, `data/federal_state_borders.geojson`, which gives the boundaries of the German federal states, and I have imported that into Python using `geopandas`, to get a (geo) DataFrame called `federal_state_borders`. The column in this DataFrame identifying the German federal states is called `VAR_NAME1`. It is a modified version of the geoJSON found [here](https://github.com/isellsoap/deutschlandGeoJSON) (my translations of the names into English are different).

For the full set-up of a minimal working example producing this problem, see here: https://github.com/krinsman/minimal_working_example
If nothing else, hopefully GitHub's display of the workbook can give someone a better idea of the structure of the relevant data, if that is necessary to understand the question.

The environment.yml and the Makefile to create the conda virtual environment from it is included for completeness, but it seems unlikely that this would be necessary in all cases.

**Problem:** Trying to create a choropleth map using this data leads to the error:    

    ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

This seems to be an issue involving how the `color_fill` attribute is interpreted.

Here is the minimal working example of the code:

    import pandas as pd
    import numpy as np
    import folium
    import geopandas as gpd
    import os

    german_cars_price=pd.read_csv('data/german_cars_price.csv',index_col=0)
    federal_state_borders = gpd.read_file(os.path.join('data', 'federal_state_borders.geojson'))

    m = folium.Map(location=[52.5194, 13.4067], tiles='Mapbox Bright', zoom_start=5)
    m.choropleth(geo_data=federal_state_borders, key_on=federal_state_borders.VARNAME_1, data=german_cars_price, columns=['state', 'price'], fill_color='YlGn')

It seems fairly likely of course that I am doing something incorrectly, but it is unclear to me based on the documentation.

**Caveat lector:** this is probably my mistake as an end-user rather than a bug of the software.

[...]

PR #797, opened 2017-12-10
More explicit checks in choropleth

### LINKED ISSUE #790: Bug when trying to set threshold scale in choropleth
I tried to set a threshold scale in choropleth using qcut.
```python
_, threshold_scale = pd.qcut(home_zips["count"], q=8, retbins=True)
m.choropleth(geo_data="./data/zip_code.geojson",
             data=home_zips,
             columns=['ZCTA5CE10',"count"],
             fill_color='YlOrRd',
             key_on='feature.properties.ZCTA5CE10',
             fill_opacity=0.9,
             line_opacity=0.5,
             legend_name='Top US Banks Clients by Home Zip',
             threshold_scale=threshold_scale)
```
#### Problem description
Whatever threshold I am trying to use, this error arises:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-75-0ea58867fab5> in <module>()
      9              line_opacity=0.5,
     10              legend_name='Top US Banks Clients by Home Zip',
---> 11              threshold_scale=threshold_scale)
     12 
     13 m.save(outfile='datamap_TopBanks.html')

/home/mlangberg/venv/local/lib/python2.7/site-packages/folium/folium.pyc in choropleth(self, geo_data, data, columns, key_on, threshold_scale, fill_color, fill_opacity, line_color, line_weight, line_opacity, name, legend_name, topojson, reset, smooth_factor, highlight)
    241 
    242         """
--> 243         if threshold_scale and len(threshold_scale) > 6:
    244             raise ValueError
    245         if data is not None and not color_brewer(fill_color):

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

#### Expected Output
I expect the choropleth map with the correct threshold assigned.

u'0.5.0'

Conengmo:
This sort of fixes #790 and #793. For `key_on` and `threshold_scale` checks for None are now explicit. In general I think that's a good idea. It will also throw more meaningful exceptions if these options are used wrong.

I didn't add a check if `key_on` starts with 'feature'. Technically it's not really necessary, so some users may omit it.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -497,11 +497,16 @@ class Map(MacroElement):
         ...              highlight=True)
 
         """
-        if threshold_scale and len(threshold_scale) > 6:
-            raise ValueError
+        if threshold_scale is not None and len(threshold_scale) > 6:
+            raise ValueError('The length of threshold_scale is {}, but it may '
+                             'not be longer than 6.'.format(len(threshold_scale)))  # noqa
         if data is not None and not color_brewer(fill_color):
             raise ValueError('Please pass a valid color brewer code to '
                              'fill_local. See docstring for valid codes.')
+        if key_on is not None and type(key_on) is not str:
+            raise ValueError('key_on should be a string, indicating the path to'
+                             ' the data key in geo_data. For example: '
+                             '\'features.properties.statename\'.')
 
         # Create color_data dict
         if hasattr(data, 'set_index'):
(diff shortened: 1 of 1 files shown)


Conengmo:
I removed the type check for `key_on`. It doesn't seem necessary since the docstring is quite clear it should be a string. Also, with the explicit None check it will already throw a more meaningful exception when it isn't a string.

ocefpaf:
Makes sense. Thanks!

PR #800, opened 2017-12-18
Link to stable and newest docs

Conengmo:
Again for #788. @AlJohri mentioned the readme only links to the docs of the development version. So I added a link to the docs of the latest release (0.5.0).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,6 +35,9 @@ or
 Documentation
 -------------
 
+http://python-visualization.github.io/folium/docs-v0.5.0/
+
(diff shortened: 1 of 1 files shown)


ocefpaf:
This is fine for now but I need to change the way we publish so "master" is the latest release, past releases are versioned, and `/dev` is latest `master`. I'll take a look at this soon.

Conengmo:
That sounds like a good idea. I don't know how that works, but seems good to learn so I'll follow what you're doing.

Issue #808, opened 2018-01-05
Any way of modifying markers or adding data to ClusterMarker ?


Twansklaf:
#### Code sample

```python

global m
    
m = folium.Map(location=center, tiles='OpenStreetMap', zoom_start=zoom_lvl, prefer_canvas=True)
    
# folium.plugins.FastMarkerCluster(coords).add_to(m)
# m.add_child(folium.plugins.FastMarkerCluster(coords))
# folium.plugins.MarkerCluster(coords).add_to(m)

for pt in coords :
        
     folium.Marker(pt).add_to(m)

```
#### Problem description

I'm making an interactive visualization map inside a Jupyter Notebook, by displaying some points (avg : 1000-2000, max : 25000). I've tried several ways : 
* placing markers one by one
* clustermarker
* fastclustermarker
Each one present advantages and drawbacks. Making markers one by one is way too slow, clustermarker is less slow and allow me to display popup Vincent graphs as I'd like, but it is too long to generate all the charts and put them in the popup, and fastclustermarker doesn't allow me to display popup charts.

I also thought about getting map zoomed coordinates (lat, long) so as to display points in this area only (great zoom when in need of Vincent chart information), that would make 20 points approximately.
In other words, the Vincent chart only needs to be computed and added to the marker when the zoom over an area is great, because the number of points when zoom is normal is too high for selecting one for information.

Is there any way to : 

* Get markers added to a map for modification
* Get map current bounds (topleft and bottom right e.g.), after movement and zoom, to dynamically display markers only in the zoomed area.
* To add info popup to FastMarkerCluster element

Conengmo:
Generating loose points should be faster now #811 is merged. 

Getting callbacks from your browser back to Python is currently not possible with folium, so you can't dynamically insert Vincent charts based on the zoom area. Of course you can build this yourself by working directly with the Leaflet API and doing javascript callbacks to some server than can serve data. But building full-fledged webapps is outside of the scope for folium.

You could hack FastMarkerCluster to enable different popups for each marker. But I'm not sure if it's worth the trouble.

Conengmo:
I'm closing this old issue since it doesn't seem actionable. If you have new information feel free to add it, we can reopen if needed.

Issue #807, opened 2018-01-04
Circle marker creation is slow


psarka:
#### The code sample

```python
import folium

map_ = folium.Map()

for i in range(1000):
    folium.features.Circle([i, i],
                           radius=3,
                           color='blue',
                           fill_color='blue',
                           fill_opacity=1,
                           fill=True).add_to(map_)
```
#### Problem description

Circle creation proceeds roughly at the order of 100 per second, which is quite slow. Creating a map with thousands of circle markers (which is handled easily by the browser) becomes slightly annoying. 

#### Expected Output

I would expect above code to run 10-1000 times faster.

#### Output of ``folium.__version__``
0.5.0

#### Extras
I tried to quickly check what is the bottleneck, and it's the jinja template compilation, which is executed 4133 times.

![circle](https://user-images.githubusercontent.com/1212244/34559260-7993ee10-f141-11e7-90ad-111e0fe01bca.png)

psarka:
OK, this is actually easy to fix. I'll open a PR in the coming days.

Conengmo:
I checked out your branch, interesting idea to move the templates out of `__init__`! It does indeed give more speed, I did a quick timing and saw a factor 2 -- 2.5 increase. But it becomes tricky when an ancestor sets the template in `__init__`. So this would need edits in Branca as well.

I have another idea for you, folium has a `fast_marker_cluster` plugin that can create many markers by doing the work in javascript. I made a test and this way can also be used for other marker types such as circles. I'll link to this issue from my branch.

psarka:
I would guess that your solution will even render faster in browser, as there will be less javascript to parse, right? It seems great! :+1: 

Moving _template from the instance variable to class variable could still make sense, hare are my observations for your consideration:

1. From a "nice" code standpoint class variable seems more appropriate, as these templates never change.

2. The speedup I get is x100 in the case of the thousand circle markers (7.45 s -> 74.8 ms) and x200 in the case of plain `Circle([1, 1], 1)` (9.55 ms -> 58.4 µs) (measured with %timeit from the notebook).

3. In the case of thousand circles, jinja2 compilation is called 6137 times. With templates moved to class attributes 37 times

4. Import time decreases from 2.6 seconds to .8 seconds (measured with `time python -c "import folium"`), although I'm not 100% sure about the representativeness of this. With pip installed version I was getting .4 seconds.

All above were run on bleeding edge branca and folium.

[...]

PR #811, opened 2018-01-09
Moves templates to class attributes (#807)

psarka:
NOTE: Conditional on https://github.com/python-visualization/branca/pull/34.
NOTE: Does not fix two unit tests that are failing at HEAD of master branch
NOTE: Does not fix unit test that fails when HEAD of master of branca is used

This pull request moves the _templates from instance  attributes to class attributes. In this way, the templates are only  compiled once (at import time), rather than every time an object is  created. 

This gives a roughly x100 speedup in cases when a large amount of (on  the order of 1000) objects is created (in my example use case going  from kind of annoying 10s to very reasonable 0.1s)  (see python-visualization/folium#807, as well as here python-visualization/folium#808) 

Having templates as class attributes also makes sense from the  conceptual perspective, as they are identical (and never modified) for  all instances.


Conengmo:
Hi @psarka, because of the great performance increase of this change it would be good to go ahead with merging this PR. Do you want to help out with fixing the conflicts by merging master into your branch?

I'm thinking about what is wise considering publishing. Your PR in branca has already been merged, but not published to PyPI. This PR breaks folium if you don't also have the branca version from master, so we have to be smart. How about:

* Merge this PR. *folium developers must be on branca master to be able to use folium.*
* Publish a new branca version to PyPI.
* Publish a new folium version to PyPI that requires the new branca version.

If breaking folium for developers using the pip version of branca is not acceptable, we should wait with merging this PR until a new version of branca is published to PyPI.

@ocefpaf is publishing a new branca version a possibility?

Conengmo:
Thanks for fixing the conflicts. You're right, the test with branca 0.2 fail because of the templates in init and the tests with branca master fail because some tests in folium haven't been updated to work with changes in branca master.

I think it is best to fix the tests in a separate PR. I tried something in #845, see if the tests for branca master work. 

We'll have to be smart about how to synchronize releases. It seems folium 0.5 works with branca master, so we can first release branca 0.3. Then fix the tests and merge this PR.  Then release folium 0.6 and require branca 0.3.

psarka:
Sounds good to me :+1: 

Let me know if you need my help on any of that.

ocefpaf:
Sorry for the long delay @psarka and thanks for the awesome speed improvement!

PR #815, opened 2018-01-23
fix notebook for hide-overlays

ocefpaf:
I wonder if we should change the default to `overlay=True` for `GeoJson` and alike. What do you think @Conengmo?


Conengmo:
Do you mean changing it from `overlay=True` to `False`? Right now most layer types have `overlay=True`, except `TileLayer` and `Layer`. That seems sensible to me. What would you like to change and why?

ocefpaf:
I guess I overlooked that the default is already `True`. It makes sense 99% of the time, so let's keep it that way. Thanks for warning me and sorry for the noise!

ocefpaf:
Wrong button...

PR #776, opened 2017-11-24
Add 'BeautifyMarker' marker type to the project.

arthuralvim:
Adding https://github.com/marslan390/BeautifyMarker to the project.

PR #819, opened 2018-02-06
Add 'BeautifyMarker' marker type to the project. Rework from Pull ID: #776

JeremyBYU:
Hi,
I'm pretty new to doing a pull request so I hope I'm doing this right. This is basically the same pull request #776 from @arthuralvim, however I tried to finish all the requests/comments you gave before merging.  I needed this work for my own research so I thought I would give it a go.

Comments/Issues from #776 hopefully resolved:
* Remove BeautifyMarker class from `folium/map.py` to its own plugin
* Dynamically load the CSS and JS
* Add Tests (first time writing tests!)
* Add an example of the BeautifyMarker plugin in `examples\Plugins.ipynb`
* Started working on `CHANGES.txt`. Once a pull id is assigned and any fixes you need are done, I will update.

Most of this work is exactly from @arthuralvim, so the credit should mostly go to him. I wasn't sure how to pull from a 'pull request' and then resubmit under the same pull ID. So I have just forked master today and updated these changes. Let me know if there is any other work you need.  

Thanks,

Jeremy


Conengmo:
Hi @JeremyBYU,

I reviewed your PR, it's a good addition. Great that you included tests and some examples. I have some suggestions, let me know what you think. I hope you're still willing to work on this!

* Could you remove the .gitignore file from the branch? There currently is no .gitignore file in folium, if we add one it should be done separately.
* If you update from master the Travis tests should succeed.
* Currently part of the code is shared with `Marker` which would be good to avoid. Since this addition is only about icons, I think it would be best to model this class after the `Icon` class. That would be most similar to what users are currently used to:
```
BeautifyIcon(some_options).add_to(marker_object)
Marker(icon=BeautifyIcon(some_options))
```
* I would suggest to parse the options similarly as in `TileLayer`. Now the options are in two places: they are stored in self (e.g. `self.icon = icon`) and then in the template they are added to an `options` variable. Directly adding them to `self.options` in the `__init__` reduces lines. Furthermore the logic (if-statements) in the template are harder to read than if they were in the `__init__`. But I know this isn't done in folium consistently, and maybe there are also arguments against this, so I understand if you keep it like it is.
* When moving the options to `self.options`, the `has_number` and `is_alpha_numeric_icon` can maybe be removed, since they have only one line and are only used once. 
* Is `_get_self_bounds()` used anywhere?

JeremyBYU:
I'll take a look at it later next week. I think the requests are all doable.

tests/plugins/test_beautify_icon.py
```diff
@@ -0,0 +1,60 @@
...
+    ic2 = plugins.BeautifyIcon(border_color='#00ABDC',
+                               text_color='#00ABDC',
+                               number=10,
+                               inner_icon_style='margin-top:0;')
+
+    # Markers, add icons as keyword argument
+    bm1 = folium.Marker(location=[46, -122],
+                        popup='Portland, OR',
+                        icon=ic1
+                       ).add_to(m)
```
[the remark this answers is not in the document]
JeremyBYU: I can fix that. However I had no idea that I pushed my changes to the python-visualization/folium repository. Are any changes that I make on *my* fork automatically pushed to yours (folium)? Or did I accidentally somehow do a push to folium? I'm guessing the later, but not sure.

JeremyBYU:
Okay @Conengmo, I think this covered what you wanted.  The only problem is that a bunch of commits occurred, and I'm not sure how it happened! I think I just realized that any push that I make to **my** fork will automatically sync up here (specifically because I'm in a pull request)! I had no idea!

Let me know if there's anything else I need to change.

folium/plugins/beautify_icon.py
```diff
@@ -0,0 +1,94 @@
...
+
+from six import iteritems
+
+from jinja2 import Template
+
+
+class BeautifyIcon(MacroElement):
+    """
+    Create a BeautifyIcon that can be added to a Marker
+    Parameters
```
Conengmo: There should be a blank line before each heading. So before `Parameters`, `Returns` and `Examples`.

folium/plugins/beautify_icon.py
```diff
@@ -0,0 +1,94 @@
...
+    ----------
+    icon: string
```
Conengmo: If you want to make this perfect, you could add the default values to each item. Like in the normal `Icon` class.

folium/plugins/beautify_icon.py
```diff
@@ -0,0 +1,94 @@
...
+        the background color of the icon
+    inner_icon_style: string with css styles for the icon.
+        the css styles of the icon
+    spin: boolean
+        allow the icon to be spinning.
+    number: integer
+        the number of the icon.
+    Returns
+    -------
+    Icon names and HTML in obj.template_vars
```
Conengmo: This class doesn't return anything, so the `Returns` part should be removed. (Technically it returns `None` but for a class docstring this is not documented.)

folium/plugins/beautify_icon.py
```diff
@@ -0,0 +1,94 @@
...
+    Examples
+    --------
+    >>> BeautifyIcon(text_color='#000', border_color='transparent',
+                     background_color='#FFF').add_to(marker)
+    >>> number_icon = BeautifyIcon(text_color='#000', border_color='transparent',
+                                    background_color='#FFF', number=10,
+                                    inner_icon_style='font-size:12px;padding-top:-5px;')
+    >>> Marker(location=[45.5, -122.3], popup=folium.Popup('Portland, OR'), icon=number_icon)
+    >>> BeautifyIcon(icon='arrow-down', icon_shape='marker').add_to(marker)
+    """
```
Conengmo: Lastly, can you add a link to the plugin website here? It's this one right?
https://github.com/marslan390/BeautifyMarker

CHANGES.txt
```diff
@@ -10,6 +10,7 @@
 - Added `tooltip` support to all vector layers (ocefpaf #722)
 - Added `TimeSliderChoropleth` plugin (halfdanrump #736)
 - Added `show` parameter to choose which overlays to show on opening (conengmo #772)
+- Added BeautifyIcon Plugin (arthuralvim and jeremybyu #776)
```
Conengmo: The number of this PR is #819, could you add it?

JeremyBYU:
Awesome thanks!

ocefpaf:
Thanks @JeremyBYU for the PR and @Conengmo for the review! The `BeautifyMarker`is an awesome addition to `folium`!

PR #824, opened 2018-02-18
Only set wind markers if wind heading is set

gaqzi:
Previously `None` would be output in the javascript and lead to a
javascript error. Now we instead use the `setHeading` function when we
don't have a wind heading set and sidestep the issue. Also doing what
the documentation hints should be possible, to let the wind markers be
optional.


gaqzi:
I believe the failed Travis build is not related to my changes. :x

ocefpaf:
Thanks @gaqzi!

PR #825, opened 2018-02-18
Add support for loading random Leaflet plugins

gaqzi:
There are two commits for two different concerns in this PR, both done to help me achieve my goal of enabling `Leaflet.RotatedMarker` in a map using `TimestampedGeoJson`.

1) Load any Leaflet plugin from an external JS file
2) Propagate named properties from a GeoJson feature's property to the marker for `TimestampedGeoJson`. This was done so that I could add `rotatedAngle` to the GeoJson and then see it show up on the marker, which would then get run by the RotatedMarker plugin

Adding the property propagation would probably be a nice addition to any marker class, but I started where I needed it. :)

I hope I updated the Jupyter Notebooks correct.


bjorn-spire:
I believe the failed Travis build is not related to my changes. :x

matkoniecz:
@python-visualization Can you consider accepting, rejecting or reviewing this PR?

Ignored PRs from 2018 are kind of discouraging.

Conengmo:
Hi Mat, I'll then close this PR, for three reasons:

- it's probably abandoned by now
- I don't think the empty Leaflet plugin class is particularly interesting
- I don't want to add these changes to timestamped geojson. It's already a quite fragile class with a relatively high burden on maintenanc, so any change that's not super simple is a no-go IMO.

PR #837, opened 2018-03-24
Added options for MarkerCluster and FastMarkerCluster

Conengmo:
I added a way to provide the marker clusters with options. It's quite straight forward. Any options will work, I specifically needed it for `disableClusteringAtZoom`.

The markerClusterGroup declaration in Javascript accepts an object with key/value pairs with options. I made the (Python) MarkerCluster and FastMarkerCluster to accept an `options` dictionary that gets json.dumped and given to the (Javascript) markerClusterGroup as argument.

Since I was there, I also included the `icon_create_function` argument, since it belongs in options. I didn't include any checks on validity of the options, so that's on the user.


Conengmo:
The test failure is fixed in #838

Conengmo:
If there are no comments I'll merge this tomorrow.

folium/plugins/marker_cluster.py
```diff
@@ -26,53 +33,54 @@ class MarkerCluster(Layer):
...
         Override the default behaviour, making possible to customize
         markers colors and sizes.
-
-    locations: list of list or array of shape (n, 2).
-        Data points of the form [[lat, lng]].
-    popups: list of length n.
-        Popup for each marker.
-    icons: list of length n.
-        Icon for each marker.
+    options : dict, default None
```
ocefpaf: We should look into a docstring template functionality to avoid these repetitions.
I believe pandas has a nice example of that, I'll look into it.

ocefpaf:
> If there are no comments I'll merge this tomorrow.

Let's try to avoid self-merges. You can assign me to review whenever you want.
(I filter out most of the PRs here but I'll definitely make time if you explicitly call for my review :wink:)

folium/plugins/marker_cluster.py
Conengmo: Good idea. I don't know how to do that yet, but I'm interested to know.

ocefpaf:
> Alright will do! Thanks for the merge.

Thanks for the PR!

Issue #826, opened 2018-02-25
Popups get hidden below overlay layers by LayerControl


mjgleason:
If you add popups to a map along with an overlay layer, controlled with the LayerControl, the popups will get buried below the when the overlay layer is turned off and back on again. They are still there, but are now behind the overlay and therefore don't respond as long as the overlay layers is visible. 

I believe that simply changing the Template for Popups (https://github.com/python-visualization/folium/blob/master/folium/map.py#L300-L313) to the code below should fix this:

```python
(u"""
            var {{this.get_name()}} = L.popup({maxWidth: '{{this.max_width}}'});

            {% for name, element in this.html._children.items() %}
                var {{name}} = $('{{element.render(**kwargs).replace('\\n',' ')}}')[0];
                {{this.get_name()}}.setContent({{name}});

            {% endfor %}

            {{this._parent.get_name()}}.bindPopup({{this.get_name()}});

            {{this._parent._parent.get_name()}}.on('overlayadd', function(){
                {{this._parent.get_name()}}.bringToFront();
            });

            {% for name, element in this.script._children.items() %}
                {{element.render()}}

            {% endfor %}
            """)
```

Conengmo:
I'm unable to reproduce your problem. Could you provide a code snippit that produces it? Also, which version are you on, 0.5.0 or the Github master?

I tried this, but the popup stays above the overlay:
```
m = folium.Map()
folium.Marker((0, 0), popup=folium.Popup('hello')).add_to(m)
folium.TileLayer(tiles='cartodbpositron', overlay=True).add_to(m)
folium.LayerControl(collapsed=False).add_to(m)
m.save('map.html')
```

mjgleason:
@Conengmo I'm on 0.5.0. Here's a snippet that should reproduce the problem:

```python
import folium
import json

# create polygon geojson
g = {'crs': {u'properties': {u'name': u'urn:ogc:def:crs:OGC:1.3:CRS84'}, 'type': 'name'},
     'features': [{'geometry': { "type": "Polygon", "coordinates": [[[-1,1], [-1,-1], [1,-1], [1,1], [-1,1]]],
                                'properties': {'id': 1}}, 
                                'type': 'Feature'}],
     'type': u'FeatureCollection'}
gj = json.dumps(g)


m = folium.Map()
# add polygon overlay
folium.GeoJson(gj, overlay=True, name="Test Layer").add_to(m)
# add polygon marker popup with the same geometry
[...]

PR #840, opened 2018-04-02
Method to keep objects in front

Conengmo:
This is an approach to tackle #826. I wanted to make something so that a user can select an object that should always stay in front. 

At first I thought of adding this as an argument when creating objects, but that would mean touching many classes and a lot of repeated code and docstrings. This would be most in line with how folium currently works, but I'm not sure if its worth it for a feature that may not be used that much.

Then I looked into adding a method to a base class such as `Element`, so a user can call it on any type of object. But that would mean adding Leaflet code to Branca, which I assume is not the purpose of Branca. Alternatively I could add a method to `Layer` in folium, but then it would not work for objects such as `Marker`.

So finally I went for adding this method to `Map`. This makes sense to me, a user is telling the map which objects it wants to keep in front. It's a separate method, so it shouldn't confuse new users, but its there if someone needs it.

Here's a code snippet to test the functionality:
```python
import folium
m = folium.Map((0, 0), zoom_start=8)
coords = [[[-1, 1], [-1, -1], [1, -1], [1, 1], [-1, 1]]]
fgs = {}
for c in ['blue', 'green', 'red']:
    fgs[c] = folium.FeatureGroup(name=c).add_to(m)
    folium.Polygon(coords, color=c, fill_color=c, popup=c).add_to(fgs[c])
m.keep_in_front(fgs['green'], fgs['blue'])
m.save('_map.html')
```


Conengmo:
The failed test has already been repaired in #838. (Should the fix be a separate pull request?)

ocefpaf:
@Conengmo this is a simple and elegant solution! Thanks!!

Issue #850, opened 2018-04-25
Apostrophes create broken maps without warning or failure


sergiolucero:
#### Please add a code sample or a nbviewer link, copy-pastable if possible

```python
import folium

fm = folium.Map(location=[48.86,2.32], zoom_start=12)
folium.CircleMarker(location=(48.8603634,2.3208891), popup="Hotel D'Orsay").add_to(fm)
fm.save('hotel.html')
```
#### Problem description

When adding popups with an apostrophe in the text, the generated HTML breaks down, so the map is generated but it cannot be displayed. This could possibly lead to failures in other cases, but this is the one I've often encountered. I believe a warning is due.

Conengmo:
You're right about the map breaking down if the popup text contains invalid html. A popup is expected to receive valid html. If not, you should enable html parsing:

`folium.Popup("Hotel D'Orsay", parse_html=True)`

But of course that's not what you're asking. You're saying there should be a warning if invalid html is passed to Popup without `parse_html=True`. Given the number of questions about this (#813, #766, #726) I think you're right. Do you want to help think about how this could be done?

How can we detect invalid html? Without making it too complicated, so no external packages or headless browsing.

- I looked at the html module in the standard library, there's a parsing module there but I can't think of a way that would work. 
- Or maybe use simple heuristics. Like that the number of apostrophes and double quotes should be even, if it is uneven the html is invalid. Or that the number of < and > characters should be the equal. Or that any & character should, with a few characters in between, be either preceded by a ? (query string) or followed by a ; (html special characters).

Another option would be to enable html parsing by default. That way a map will never break silently. It will change behavior for users who are providing html though. But they will see that their html is being parsed, and can figure out they need to set `parse_html=False`. I guess this is a controversial proposal since it seems specifically chosen to do it this way, so that would need approval by others as well.

Conengmo:
This has been fixed in #962.

PR #851, opened 2018-04-25
WIP: method in Popup that validates html

Conengmo:
This is a quick demonstrator of an idea for #850.


Conengmo:
This is no longer needed now we use javascript template literals.

PR #869, opened 2018-05-15
Require branca 0.3.0, raise error otherwise

Conengmo:
Moving the templates to the class attributes breaks folium when using older branca versions which still have the templates in the init's. The failure is silently, so it may confuse users. So apart from listing the version in requirements.txt I also propose raising an exception when the branca version is older than 0.3.0.

### DIFF
--- a/folium/__init__.py
+++ b/folium/__init__.py
@@ -23,6 +23,11 @@ from folium.map import (
 
 from folium.vector_layers import Circle, CircleMarker, PolyLine, Polygon, Rectangle  # noqa
 
+import branca
+if tuple(int(x) for x in branca.__version__.split('.')) < (0, 3, 0):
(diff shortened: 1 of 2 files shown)


Conengmo:
I just witnessed someone being puzzled by an empty map because of his outdated branca, so I'll go ahead and merge this one to prevent others having the same experience.

ocefpaf:
>  so I'll go ahead and merge this one to prevent others having the same experience.

Beat you to it :stuck_out_tongue_winking_eye: 

Sorry I missed this one!

Conengmo:
Haha wow I was just about to press the button :p

ocefpaf:
BTW, this one was perfect for a self-merge, no need to wait for me next time. (Or ping me explicitly if you need a second opinion.)

PR #871, opened 2018-05-19
Updated Tooltip Functionality to reference JSON.

jtbaker:
Updated the Tooltip to take a dictionary referencing GeoJson property fields, alias them, and display the values on separate lines. Use case can be found at the below notebook. Updated DocString as Well.

http://nbviewer.jupyter.org/gist/jtbaker/c472b937fb271726949d51ad340976f0?flush_cache=true

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -325,6 +325,20 @@ class GeoJson(Layer):
         How much to simplify the polyline on each zoom level. More means
         better performance and smoother look, and less means more accurate
         representation. Leaflet defaults to 1.0.
+    tooltip: default None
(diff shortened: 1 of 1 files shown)


ocefpaf:
@jtbaker can you add a test and a changelog entry for this?

Also, can you submit your notebook to our example gallery?

jtbaker:
@ocefpaf I rethought how I wanted to approach this and did some more work to create a separate class. Just committed a separate PR onto the current master to minimize any conflicts that might occur by merging this older one. I'll close this one now.

PR #883, opened 2018-06-15
Tooltip classes (simple and geojson/topojson)

jtbaker:
Added a new Tooltip class to reference GeoJson feature properties, with tests, documentation, and examples. It's flexible, and can take field names, aliases, that can be turned on/off with the 'labels' boolean, a 'sticky' property, and can incorporate Javascript's .toLocaleString() functionality if desired. The same string can be passed for each object with the 'text' variable as well if you'd rather not use the 'fields'.

Also updated the GeoJson JS template to reference this Tooltip object, and added a test to make sure that the values passed exist in the properties.

Use case at [this nbviewer](http://nbviewer.jupyter.org/gist/jtbaker/57a37a14b90feeab7c67a687c398142c?flush_cache=true)


ocefpaf:
@jtbaker god start! Can you please add a test and, if possible, a notebook example?

PS: it seems you have some linter issue there.

jtbaker:
@ocefpaf fixed most of my linting issues, and the one assertion error for the GeoJson class was due to me using an f-string in the check - switching to normal concatenation with .__str__() called for the variable objects fixed the issues.
However, the above Travis CI build is failing on this error:

`The command "if [[ $TEST_TARGET == 'notebooks' ]]; then for file in $(find . -type f -name "*.ipynb"); do jupyter nbconvert --template=tests/strip_markdown.tpl --stdout --to python $file | grep -v '^get_ipython' | flake8 - --ignore=W391,E226,E402,I100 --max-line-length=100 --show-source ; done ; py.test -vv tests/notebooks/test_notebooks.py ; fi" exited with 1.`

I'm not sure if

jtbaker:
Accidentally closed. Reopening.

folium/features.py
```diff
@@ -164,7 +164,7 @@ def render(self, **kwargs):
             """).render(this=self, kwargs=kwargs)), name=self.get_name())
 
         self._parent.script.add_child(Element(Template("""
-            vega_parse({{this.json}},{{this.get_name()}});
+            vega_parse({{this.json}},{{this.get_name()}
```
ocefpaf: Why did you remove the `});`?

folium/features.py
```diff
@@ -652,6 +682,70 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
         self.className = class_name
 
 
+class Tooltip:
+    """
+Creates a Tooltip object for adding to features to display text as a property a Map by executing a javascript function
+when hovering the cursor over each feature.
+
+Parameters
+----------
```
ocefpaf: Please fix the indentation of the docstrings.

folium/features.py
```diff
@@ -652,6 +682,70 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
         self.className = class_name
 
 
+class Tooltip:
```
ocefpaf: My guess is that deriving from `object` is proffered, no? (This is the old style `class` if I'm not mistaken.)

folium/features.py
jtbaker: Whoops! Fixed.

folium/features.py
jtbaker: @ocefpaf think I fixed most of the issues. Thanks for your patience - still new to deploying standard code for public use, but learning a lot with this.

folium/features.py
jtbaker: Is there a reason to use the new style if it does not need class inheritance from another property? Or should I instantiate it with no arguments like `class Tooltip():`?

folium/features.py
```diff
@@ -377,7 +377,31 @@ class GeoJson(Layer):
                         }
                     {% endif %}
                     )
-                    {% if this.tooltip %}.bindTooltip("{{this.tooltip.__str__()}}"){% endif %}
+                    {% if this.tooltip %}
+                    .bindTooltip(function(layer){
+                        {% if this.tooltip.fields %}
+                        let fields = {{ this.tooltip.fields }};
+                            {% if this.tooltip.aliases %}
```
Conengmo: This extra indentation shouldn't be here, please make this consistent.

folium/features.py
```diff
@@ -417,6 +440,13 @@ def __init__(self, data, style_function=None, name=None,
 
         self.highlight_function = highlight_function or (lambda x: {})
 
+        self.tooltip = tooltip
+        if self.tooltip:
+            if self.tooltip.fields:
```
Conengmo: You can combine these two if statements:

`if tooltip and tooltip.fields`

folium/features.py
```diff
@@ -652,6 +682,69 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
         self.className = class_name
 
 
+class Tooltip:
+    """
+    Creates a Tooltip object for adding to features to display text as a property a Map by executing a javascript
```
Conengmo: folium uses a line width of 80 characters, please adhere to it in your code. Are you using Pycharm? I recommend to set the line width in your IDE to warn you when you go over it. Your IDE should also warn for lint errors.

folium/features.py
```diff
@@ -652,6 +682,69 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
...
+    function when hovering the cursor over each feature.
+
+    Parameters
+    ----------
+        fields: list or tuple.
```
Conengmo: The parameter names shouldn't be indentend, but have the same indentation level as the rest of the docstring

folium/features.py
```diff
@@ -652,6 +682,69 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
...
+            Labels of the GeoJson 'properties' or GeoPandas GeodataFrame columns you'd like to display.
+        aliases: list or tuple
+            Optional 'aliases' you'd like to display the each field name as, to describe the data in the tooltip.
+            *Must consist of strings, and be of the same length as fields.
+        text: str
+            Pass the same string as a tooltip for every value in the GeoJson object, I.e. "Click for more info."
+            *May not be passed if fields is not None
+        labels: boolean True or False
```
Conengmo: You don't have to specify that a boolean is True or False. I would suggest to write:

`labels : bool, default True`

folium/features.py
```diff
@@ -652,6 +682,69 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
...
+            True or False value indicating if you'd like the the field names or aliases to display to the left of
+            the value in bold.
+            *Defaults to True.
```
Conengmo: Put the default value on the first line. This goes for all parameters. E.g.:

`text : str, default None`

folium/features.py
```diff
@@ -652,6 +682,69 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
...
+        fields: list or tuple.
+            Labels of the GeoJson 'properties' or GeoPandas GeodataFrame columns you'd like to display.
+        aliases: list or tuple
+            Optional 'aliases' you'd like to display the each field name as, to describe the data in the tooltip.
+            *Must consist of strings, and be of the same length as fields.
+        text: str
+            Pass the same string as a tooltip for every value in the GeoJson object, I.e. "Click for more info."
+            *May not be passed if fields is not None
+        labels: boolean True or False
+            True or False value indicating if you'd like the the field names or aliases to display to the left of
```
Conengmo: You already said it's a bool, so no need to again state it's a True or False value

ocefpaf:
> I want to suggest to make Tooltip a normal folium object, with its own template.

:+1:

folium/features.py
jtbaker: I was, but had switched over to VS Code and committed through the web portal to help out with the JS syntax highlighting and brace closures in the template, as keeping it organized with the Jinja2 curly braces expressions, JS template literal embeds was a little challenging (for me). Running a refactor in PyCharm through the VCS now - thanks for your patience and suggestions!

jtbaker: Thanks! Looking through the repo and other classes, I see several other code >80 character width violations. I've resolved the issues within my python code, but getting the Jinja2 template to comply with all the conditional assertions and embeds within the JS is challenging. Would the template be an acceptable violation, in order to keep it thematically organized by line and indentation so that it is easier to read/modify for future contributors?

jtbaker:
Accidentally closed this again when trying to resolve correct some mistakes on my local git client.

folium/features.py
```diff
@@ -652,6 +685,79 @@ def __init__(self, html=None, icon_size=None, icon_anchor=None,
         self.className = class_name
 
 
+class Tooltip():
```
ocefpaf: Please inherit from `object` here.
https://stackoverflow.com/questions/4015417/python-class-inherits-object

jtbaker:
@Conengmo, I think I've addressed most of your comments in the Tooltip class as of the latest commit. Thanks for taking the time to look at it, and for your suggestions. I learned quite a bit about lower level Pythong and a ton about JS in the process (spent a good amount of time debugging because I didn't understand that `typeof([item1, item2])` returns `"object"` and not `Array`, ha.)

Regarding how it is named and where the Template is located, thinking about it it seemed easiest to me to have a central Tooltip class that could be used for a variety of different folium objects, each with a template built for them - I've just cursorily looked over Topojson, but it seems like with the way the data is organized, it will require a separate tm

ocefpaf:
@Conengmo this looks good to me. Can you take a second look and merge if you approve it too?

PS: let's squash and merge the commits on this one.

Conengmo:
I think I agree with your line of thinking @jtbaker, I'll try to give my take:

The easiest thing to do would be to merge it as is. Don't let 'perfect' stand in the way of good. It works well for the `GeoJson` use case. We can refactor it later when we make a generally usable `Tooltip` class. Some users may need to change their code to work with any changes.

I thing the `Tooltip` class could be a bit better as it is. Let it have it's own template, so it can be added to a `GeoJson` object like `Tooltip().add_to(GeoJson())`, just like other folium objects work.

But now that we're looking at tooltips, I would like to have a `tooltip` class for general folium objects as well. It shouldn't be too hard to specify which arguments are only meant to be used when it's added to a `GeoJson` object. We only need two switches AFAIK, either a fixed text, or getting data from a `GeoJson` thing.

So if we do these two things (make `Tooltip` a proper folium object, let it work with general folium objects) that would be great. But it's also okay to merge it as it is, we can improve later. So I guess it depends on how far you want to go with this now @jtbaker.

jtbaker:
@Conengmo Thanks for the feedback. I did some more work to handle those cases with people's code for the previous version of the object, i.e. fixed string, so this merge wouldn't break their code.

I'm happy to continue to work on this, but if you'd like to go ahead and merge here and open up another PR for the proper object model, I can do that. Could you point me towards another object that might be a good model to use for the `Tooltip(*args).add_to(GeoJson())` structure?

This model could be easily adapted for the Popup object as well (at least for strings). I'll do a little more research and see what it would take to do something like use a country name in the geojson properties to open an iframe with a wikipedia page in the popup, or generate a an Altair->Vega-Lite chart based on the property values.

Conengmo:
I think a good example would be how the `Marker` and `Popup` classes currently work. `Tooltip` should work similarly to `Popup`. I'll add some pointers and suggestions here:

The crux here is to add a line in the `Tooltip` template where the Leaflet tooltip object is added to its parent object, whether that's a geojson thing or a marker or anything:

`{{ this._parent.get_name() }}.bindTooltip({{ this.get_name() }})`

This means that in the template you must first create a Leaflet tooltip object:

`var {{ this.get_name() }} = L.tooltip();`

If you move your template to `Tooltip`, note that you no longer have to use `this.tooltip.whatever` but can simplify it to `this.whatever`.

In `GeoJson` you can still check the validity of the fields in the `Tooltip` object that is passed, but after that you shouldn't store it in `self.tooltip`, but instead add it as a child: `self.add_child(tooltip)`.

`Popup` doesn't have a `text` argument, but instead accepts html, with a flag to allow for the string to be parsed. I'm not entirely happy with this because users may assume to just be able to pass a string without checking that it should be valid html and then the map silently failing. But  that's a discussion for somewhere else. Maybe be consistent here and use it in the same way.

`` \_\_init__(self, html=None, parse_html=False, ...`

You mentioned adapting `Popup` to work with `GeoJson` fields as well, that sounds interesting. More reason to make these classes similar, so if a user understands one they can use the other as well.

folium/folium.py
```diff
@@ -15,7 +15,7 @@
 from branca.element import CssLink, Element, Figure, JavascriptLink, MacroElement
 from branca.utilities import _parse_size, color_brewer
 
-from folium.features import GeoJson, TopoJson
+from folium.features import GeoJson, TopoJson, Tooltip
```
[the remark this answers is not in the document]
Conengmo: You should add this import to `__init__.py` instead.

jtbaker:
Cool. Thanks for the feedback. Looking over the Leaflet documentation ([1](https://leafletjs.com/reference-1.0.0.html#tooltip-l-tooltip), [2](https://www.npmjs.com/package/leaflet-tooltip)) on this, I can't seem to find an example on how a separate `L.tooltip()` object is added to map objects outside of the `.bindTooltip()` method. The structure seems to be `L.tooltip(<Tooltip options> options?, <Layer> source?)`, so I'll need to target my tooltip layer back to the GeoJSON object, but it's not enirely clear to me how that's done - do I add the tooltip to the GeoJSON object, or to the map? Can I pass my function into the `L.tooltip().setContent()` method? I've tried some jsfidlling to try to get it to behave as expected, to no avail.

Conengmo:
`map.py` seems like a good call. If it's next to `Marker` and `Popup`, that would be similar to the location of Tooltip in the Leaflet documentation.

jtbaker:
@Conengmo can you check where I am on the [New-Tooltip branch](https://github.com/jtbaker/folium/blob/New-Tooltip/folium/map.py#L383-L412) of my folium fork? I think I'm most of the way there, but the `{{ this._parent.get_name()}}.bindTooltip({{ this.get_name() }})` doesn't seem to be executing. My `Tooltip` class should inherit from `Element` for `.get_name()`, right?

Conengmo:
It's looking good @jtbaker! I haven't tried it yet, but I'll give some comments right away.

* In the `GeoJson` class, shouldn't there be a check if a tooltip argument is provided? (on line 417)
* The assert error message on line 423 is cleaner IMO if you use a string format, e.g. `'this is {}'.format(int_value)`
* Your docstring for `Tooltip` is very complete, nice. Since the class works for all folium objects, it would help if you alter the text to reflect this. For example for the `text` argument you say 'Pass the same string as a tooltip for every value in the GeoJson object', which is not really generic if you're e.g. adding a tooltip to a marker.
* Make sure the order of the arguments in the docstring is the same as in the actual `__init__()`.
* I'm not sure on the best way to bind it. I suppose your current implementation should work? I'll try it next time, see what's going on.

jtbaker:
@Conengmo Check the latest commit to the PR - progress! We've now got functionality on `GeoJSON`, `TopoJSON`, `Marker` and `Vector` layers - users can pass either `Tooltip` objects or strings(as they did before) to `tooltip` args on those classes.

`Tooltip` now also has a `style` arg in which users can pass inline CSS style strings to style the `div` the tooltip now contains, allowing for custom fonts, colors, etc.

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
                     {{this.location}},
                     {{ this.options }}
                     )
-                    {% if this.tooltip %}.bindTooltip("{{this.tooltip.__str__()}}"){% endif %}
+                    {% if this.tooltip %}.bindTooltip("{{ this.tooltip }}"){% endif %}
```
Conengmo: We don't need these lines in all these templates anymore now that we have a Tooltip class! In all cases add a Tooltip object as a child.

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
...
             """)  # noqa
 
     def __init__(self, locations, popup=None, tooltip=None, **kwargs):
         super(PolyLine, self).__init__(location=locations, popup=popup)
         self._name = 'PolyLine'
-        self.tooltip = tooltip
-
+        if tooltip:
+            if isinstance(tooltip, Tooltip):
+                assert not all((tooltip.text, tooltip.fields)), "Only text " \
```
Conengmo: This check is already done in the init of `Tooltip`, don't repeat it.

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
...
+                                                                "may be " \
+                                                                "passed to a " \
+                                                                "Polyline " \
+                                                                "Tooltip."
+                self.add_child(tooltip, name=tooltip._name)
```
Conengmo: You don't have to pass the name.

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
...
+            elif isinstance(tooltip, str):
+                self.tooltip = tooltip.__str__()
```
Conengmo: Use a Tooltip object here, that's why we have it now.

`self.add_child(Tooltip(tooltip))`

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
...
+            else:
```
Conengmo: It's good manners that you raise an error here. A suggestion to shave a line:

don't use `else`, but use `elif tooltip is not None`. Then you can remove the `if tooltip` at the start of this block.

folium/vector_layers.py
```diff
@@ -159,16 +159,27 @@ class PolyLine(Marker):
...
                     .addTo({{this._parent.get_name()}});
             {% endmacro %}
             """)  # noqa
 
     def __init__(self, locations, popup=None, tooltip=None, **kwargs):
         super(PolyLine, self).__init__(location=locations, popup=popup)
         self._name = 'PolyLine'
-        self.tooltip = tooltip
-
+        if tooltip:
```
Conengmo: This line can be removed if you change the `else` statement below, see the comment there.

folium/map.py
```diff
@@ -335,6 +348,140 @@ def render(self, **kwargs):
...
+                toLocaleString=True,
+                style='background-color: grey; color: white; font-family:
+                courier new; font-size: 24px; padding: 10px;')
+    # Provide fields, with labels off, and sticky True.
+    >>> Tooltip(fields=('CNTY_NM',), labels=False, sticky=True)
+    # Provide only text.
+    >>> Tooltip(text="Click for more info.", sticky=True)
+    """
+    _template = Template(u"""
+            {% macro script(this, kwargs) %}
```
Conengmo: One level of indentation is enough, so you can unindent this one level.

folium/map.py
```diff
@@ -335,6 +348,140 @@ def render(self, **kwargs):
...
+        self.aliases = aliases
+        self.text = text.__str__()
+        self.labels = labels
+        self.toLocaleString = toLocaleString
+        if style:
+            assert isinstance(style, str), \
+                "Pass a valid inline HTML style property string to style."
+            # noqa outside of type checking.
+            self.style = style
+        if self.fields:
```
Conengmo: This can become a oneliner:

`self.result = self.fields if self.fields else self.text`

folium/map.py
```diff
@@ -335,6 +348,140 @@ def render(self, **kwargs):
...
+    Parameters
+    ----------
+    text: str, may not be passed if fields is not None.
+        Pass a string as a tooltip to display on the object.
+        object, I.e. "Click for more info."
+    fields: list or tuple. For GeoJson/TopoJson objects only.
+        Labels of GeoJson/TopoJson 'properties' or GeoPandas GeoDataFrame
+        columns you'd like to display.
+    aliases: list/tuple of strings, same length/order as fields. For
+    GeoJson/TopoJson objects only.
```
Conengmo: This should be indented one level

folium/vector_layers.py
jtbaker: @Conengmo the reason I had preserved them on these was in the case the user passes a string instead of a `Tooltip` to the `tooltip` arg. This won't break code written for previous folium versions on these classes and lets people just write in simple text for the tooltip, if they don't need our additional `Tooltip` functionality exposed, as I imagine may be the case for looping over many `Marker`'s ,etc.

`self.tooltip = tooltip` only if `isinstance(tooltip, str)`
`self.add_child(tooltip)` only if `isinstance(tooltip, Tooltip)`

and in each object's template, the tooltip is only bound if `self.tooltip` exists:

`{% if self.tooltip  %}.bindTooltip('{{ self.tooltip }}'){% endif %}`

Thoughts?

folium/vector_layers.py
jtbaker: Smart, eliminate the redundancy and funnel everything through the `Tooltip` object and template. This makes it way more maintainable. Thanks for this suggestion.

folium/map.py
jtbaker: Looking through it, it seems like `self.result` was a relic from when this was in the GeoJson template. It can go now.

folium/vector_layers.py
jtbaker: All of these checks have now been incorporated in the `Marker.validate_tooltip()` method. Thanks for the suggestions!

folium/vector_layers.py
jtbaker: See above.

folium/vector_layers.py
jtbaker: Duly noted.

folium/vector_layers.py
```diff
@@ -12,7 +12,7 @@
 from branca.element import (CssLink, Element, Figure, JavascriptLink, MacroElement)  # noqa
 from branca.utilities import (_locations_tolist, _parse_size, image_to_url, iter_points, none_max, none_min)  # noqa
 
-from folium.map import Marker
+from folium.map import Marker, Tooltip
```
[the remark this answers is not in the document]
jtbaker: Tooltip class is referenced while type checking in the `Marker.validate_tooltip()` method.

jtbaker:
@Conengmo have you had a chance to look the changes? I just updated the template to reference a 
```
<table>
<tr><th>Label</th><td>Value</td></tr>
</table>
```
as well for a nicer formatting structure.

Sorry for all these commits - this is my first time working on a collaborative project, but I'm learning quite a lot, and want to make as high quality/useful contribution as possible, since folium has proven quite useful to me in my work - being able to generate self contained web maps inline with data processing tools has been invaluable.

I'll definitely have some better Git practices for my next PR - I just have been committing frequently since I'm working across two different machines at different times of day. Thanks again for your patience and suggestions.

Conengmo:
I had a look at your most recent changes, the table looks nice. Instead of doing a review I did some work of my own, it's in a PR in your fork. 

Don't worry about all the commits :) your contribution is welcome. Though good commits help in making reviews easier, so it's a good skill to learn. Note that we're not reviewing on every commit, so give a shout when you want us to have another look.

jtbaker:
@Conengmo get a chance to review my [question](https://github.com/jtbaker/folium/pull/19#discussion_r204835817) regarding the `fields` arg passing in Marker tooltips? I know this isn't likely to be passed on these features, but I would like the templates that use the class to be airtight if possible.

If you don't think that it's worth the time, I'm happy to merge as is. I'd love this feature to be incorporated in the v0.6 release if possible.

ocefpaf:
> If you don't think that it's worth the time, I'm happy to merge as is. I'd love this feature to be incorporated in the v0.6 release if possible.

Let's make that happen!

Conengmo:
I haven't forgotten about this, just been busy with work.

I read an interesting article the other day about writing good code, can't find the link now though. One of the things it wrote about was not having all these if-switches in a function or class. If you find that your function/class is doing multiple things and needs if-statements to separate between these things, it's probably better to split it in multiple functions/classes.

I think that also applies here. We have a single `Tooltip` class that does two different things: render normal text and render geojson/topojson data fields. There are switches in multiple places to distinguish between the two behaviors. Users have to be careful to use the right arguments for their usecase.

So I want to make the case for splitting the class in two: one for simple text tooltips and one for more complicated geojson/topojsons. I'll open a PR on your branch again @jtbaker so you can check it out. I haven't tested it yet so please review.

ocefpaf:
> I read an interesting article the other day about writing good code, can't find the link now though. One of the things it wrote about was not having all these if-switches in a function or class. If you find that your function/class is doing multiple things and needs if-statements to separate between these things, it's probably better to split it in multiple functions/classes.

:+1: 

> So I want to make the case for splitting the class in two: one for simple text tooltips and one for more complicated geojson/topojsons. I'll open a PR on your branch again @jtbaker so you can check it out. I haven't tested it yet so please review.

That is a sound approach. Thanks @Conengmo!

jtbaker:
Thanks for the feedback and PR on this again, @Conengmo. I'm happy with this approach and think it makes for cleaner design, and will be easier to maintain.

I've left a few comments regarding the `kwargs` usage in the classes in your PR on my branch - once we figure out how to resolve those, I'd love to merge.

folium/map.py
```diff
@@ -186,21 +187,29 @@ class Icon(MacroElement):
                     iconColor: '{{this.icon_color}}',
                     markerColor: '{{this.color}}',
                     prefix: '{{this.prefix}}',
-                    extraClasses: 'fa-rotate-{{this.angle}}'
+                    extraClasses: 'fa-rotate-{{this.angle}}',
```
Conengmo: You're adding two options to `Icon`, which is fine in itself but please do this in a separate PR.

folium/map.py
```diff
@@ -186,21 +187,29 @@ class Icon(MacroElement):
...
                 {{this._parent.get_name()}}.setIcon({{this.get_name()}});
             {% endmacro %}
             """)
 
     def __init__(self, color='blue', icon_color='white', icon='info-sign',
-                 angle=0, prefix='glyphicon'):
+                 angle=0, prefix='glyphicon', spin=False):
         super(Icon, self).__init__()
-        self._name = 'Icon'
+        self._name = 'AwesomeMarkers.icon'
```
Conengmo: Changing the name could have unintended consequences so please do this in a separate PR so it doesn't go unnoticed.

folium/map.py
```diff
@@ -186,21 +187,29 @@ class Icon(MacroElement):
...
         self.color = color
         self.icon = icon
         self.icon_color = icon_color
         self.angle = angle
         self.prefix = prefix
+        self.spin = spin
+        self.options = {'icon': icon,
```
Conengmo: Changing the way options work here doesn't belong in this PR.

folium/map.py
```diff
@@ -228,33 +239,32 @@ class Marker(MacroElement):
...
+            [{{this.location[0]}}, {{this.location[1]}}],
+            {
+                icon: new L.Icon.Default()
+                }
+            ).addTo({{this._parent.get_name()}});
+        {% endmacro %}
+        """)
 
-    def __init__(self, location, popup=None, tooltip=None, icon=None):
+    def __init__(self, location=None, popup=None, tooltip=None, icon=None):
```
Conengmo: Changing the default of `location` here doesn't have to do with tooltip, please remove it here.

folium/map.py
```diff
@@ -228,33 +239,32 @@ class Marker(MacroElement):
...
         super(Marker, self).__init__()
-        self._name = 'Marker'
-        self.tooltip = tooltip
-        self.location = _validate_coordinates(location)
+        self._name = 'marker'
+        if location:
```
Conengmo: This check fails with Numpy arrays. Also adding it doesn't belong in this PR.

Conengmo:
I removed the stowaway xml files, @jtbaker maybe add them to your `.gitignore` file so you don't accidentally commit them again?

I think we're good to go!

jtbaker:
@Conengmo Awesome! Thanks for your help on this. This is my first real open source contribution, and I learned quite a bit in the process - have a decent grasp on JS and lower level Python, project best practices and scoping now. Looking forward to future PRs (with less errors).  :-) 

Since the changes have been approved, and I created this PR, am I good to close it out now? Is the merge onto the master complete, or does it need another review from @ocefpaf or someone else?

ocefpaf:
>  @ocefpaf or someone else?

LGTM but I'll leave to @Conengmo to do the honors and merge it.

> This is my first real open source contribution

Awesome! And this is not an easy one!! Thanks for all the effort and for hanging in there while we reviewed it.

Thanks for the PR! Hope to see more :wink:

Conengmo:
Merged! Congrats @jtbaker and glad to hear you learned stuff. It took some time to figure out how to get this right, so I appreciate your patience on this PR. 

Future PRs sounds good, let's do it!

baskoes:
@jtbaker thank you for your efforts. I would really appreciate an example that combines chloropleth and tooltip into one map from the same dataframe / geojson. The example in the doc just seems to call "Tooltip(" and I can't figure it out. If someone could point me in the right direction I would be happy to share a ipynb for the examples directory. thanks again.

jtbaker:
@Conengmo Looks like we forgot to update the naming in the docstring after the class split. Whoops!

@baskoes, Thank you! It was fun, and I learned quite a bit about how the library works under the hood. Those references to `Tooltip` in the `GeoJsonTooltip` docstring should be to `GeoJsonTooltip` itself(we'll get these updated ASAP). `Tooltip` is a separate class for passing text of your choice, through the `text` argument - i.e. 'Click for more info'. `GeoJsonTooltip` allows you to reference the column names/`properties` keys, to display the associated values of each feature in the GeoJson dataset via the `fields` argument, with an option to relabel/describe the data in the visualization. This is completed on the fly for each feature using a JavaScript function.

I'll get a new notebook added to the examples directory to do a demo on how to use both classes.

Conengmo:
> Looks like we forgot to update the naming in the docstring after the class split.

Good catch, didn't notice that.

A notebook with examples of `GeoJsonTooltip` would be great, looking forward to reviewing and merging it. We can change the docstring in the same PR I suppose.

PR #893, opened 2018-06-23
MultiPolygon in doc of TimestampedGeoJson should be changed to Polygon

andy23512:
According to change log,
https://github.com/python-visualization/folium/blob/19bde41edd6346f447bd83523340edb9782dc20a/CHANGES.txt#L120-L121
MultiPolygon was removed and was handled by Polygon, so the document (auto-generated from comment in code, right?) should be updated to prevent misuse.


Conengmo:
I think the changelog references the folium objects, not the Geojson types. For example in vector_layers.py there is a `Polygon` class, but you can also define a polygon type in a geojson definition.

I'm not sure where that line you changed in `TimestampedGeoJson` came from. If you want to figure this out, maybe you could verify if the class works with polygon and multipolygon geojson types.

andy23512:
At first, I used this code ( https://gist.github.com/andy23512/392f01bd7e6f94e82a9996a9922a911d ) with `'type': 'MultiPolygon' `, and the time line wasn't worked (showing "Time is not available" message). After some research, I found that MultiPolygon is deprecated and is handled by Polygon in change log.
So as I changed `'type': 'MultiPolygon'` to `'type': 'Polygon'`, the time line was worked.
I thought that this bug might be directly corresponding to the deprecation of MultiPolygon, since the dict of feature collection is completely transferred to Leaflet.js, and the MultiPolygon class of Leaflet.js have been deprecated.

Conengmo:
I just tried both Polygon and MultiPolygon and both work. MultiPolygon has a different data definition, it needs an extra list. Check out the docs:

http://wiki.geojson.org/GeoJSON_draft_version_6#Polygon
http://wiki.geojson.org/GeoJSON_draft_version_6#MultiPolygon

So if anything the docstring should reflect that using a Polygon is _also_ a possibility.

Here is the code I use to create polygons and multipolygons:

```
def get_lat_lon():
    return random.random() * 2 - 1, (random.random() * 2 - 1) * 2


def get_polygon(nodes):
    return [get_lat_lon() for i in range(nodes)]


def get_multipolygon(nodes_1, nodes_2):
    return [
        [get_lat_lon() for i in range(nodes_1)],
        [get_lat_lon() for i in range(nodes_2)]
    ]

n = 5
times = (1435752000000 + 24 * 3600 * 1000 * np.arange(n)).tolist()
polygons = [get_polygon(3) for j in range(n)]
multipolygons = [get_multipolygon(3, 4) for j in range(n)]

folium.plugins.TimestampedGeoJson({
    'type': 'FeatureCollection',
    'features': [{
        'type': 'Feature',
        'geometry': {'type': 'MultiPolygon', 'coordinates': multipolygons},
        'properties': {'times': times}
    }]
}, duration='PT1H', transition_time=1000).add_to(m)
```

andy23512:
Thanks for the explanation to the data definition, I just update the data with an extra list, and it works.

Conengmo:
Merged! Good to sort this out.

PR #894, opened 2018-06-23
Add duration to timestamped geo json

andy23512:
Add duration option from L.timeDimension.layer.geoJson ( https://github.com/socib/Leaflet.TimeDimension/tree/520cb80f645112e242c5160cb44b7d5f2cae380d#options-3 ), in order to control the shown duration of feature.

Is there anything else I need to do?


folium/plugins/timestamped_geo_json.py
```diff
@@ -41,6 +41,10 @@ class TimestampedGeoJson(MacroElement):
         Used to construct the array of available times starting
         from the first available time. Format: ISO8601 Duration
         ex: 'P1M' -> 1/month, 'P1D' -> 1/day, 'PT1H' -> 1/hour, and'PT1M' -> 1/minute
+    duration: str, default None
+        Period of time which the features will be shown on the map after their time
+        has passed. If null, all previous times will be shown. Format: ISO8601 Duration
```
Conengmo: null -> None

folium/plugins/timestamped_geo_json.py
```diff
@@ -110,14 +114,14 @@ class TimestampedGeoJson(MacroElement):
                 })
 
             var {{this.get_name()}} = L.timeDimension.layer.geoJson(geoJsonLayer,
-                {updateTimeDimension: true,addlastPoint: {{'true' if this.add_last_point else 'false'}}}
+                {updateTimeDimension: true,addlastPoint: {{'true' if this.add_last_point else 'false'}},duration: {{"\""+this.duration+"\"" if this.duration else 'undefined'}}}
```
Conengmo: Can you move the if-statement to the Python code? Where `self.duration` is set.  Makes it easier to read IMO. That way here you can just include it like `duration: "{{ this.duration }}"}`

folium/plugins/timestamped_geo_json.py
```diff
@@ -133,6 +137,7 @@ def __init__(self, data, transition_time=200, loop=True, auto_play=True, add_las
         self.add_last_point = bool(add_last_point)
         self.period = period
         self.date_options = date_options
+        self.duration = duration
```
Conengmo: Then this could be `self.duration = duration if duration is not None else 'undefined'`

andy23512:
Thanks for the suggestions. The code, doc, test suite has been updated, and duration option has been added to existing example.

Conengmo:
Merged. Thanks for your contribution @andy23512!

KirstenCrane8005:
Hello, I am getting the error message "TypeError: __init__() got an unexpected keyword argument 'duration' " when using the duration parameter. Are you able to help with this? I want the markers to only show for the associated timestamp but, at the minute, new markers are just getting added on top of old markers. Kirsten

Conengmo:
Hi @KirstenCrane8005, are you sure you got a recent version? Your error message suggests you don't have a recent version with this PR merged in it. You can check your current version with `print(folium.__version__)`. You should install folium from the git master branch.

KirstenCrane8005:
Hi, I spoke to @andy23512  and he helped me install the most recent development version. I had used pip to install so only had the latest released version 0.5.0. Thanks for your response! If anyone knows whether it is possible to pass a folium marker or folium icon such as BeautifyIcon to the 'icon' parameter under 'properties' when you're creating the geoJSON feature objects that would be a big help. I just get the error message folium marker not JSON serializable. Aim is to display a number on each marker, pulled from my pandas dataframe.

PR #907, opened 2018-07-12
Adding name variable to Map class to create custom names for tiles

mccarthyryanc:
I wanted the ability to define a name when creating Map from custom Leaflet-style URLs. The default was to use the full URL.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -147,6 +147,8 @@ class Map(MacroElement):
         rare environments) even if they're supported.
     zoom_control : bool, default True
         Display zoom controls on the map.
+    name : str, default None
+        Set a custom name when creating a map via tiles.
 
     Returns
     -------
@@ -225,7 +227,8 @@ $(document).ready(objects_in_front);
                  min_lon=-180, max_lon=180, max_bounds=False,
                  detect_retina=False, crs='EPSG3857', control_scale=False,
                  prefer_canvas=False, no_touch=False, disable_3d=False,
-                 subdomains='abc', png_enabled=False, zoom_control=True):
+                 subdomains='abc', png_enabled=False, zoom_control=True,
+                 name=None):
         super(Map, self).__init__()
         self._name = 'Map'
         self._env = ENV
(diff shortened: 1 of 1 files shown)


Conengmo:
I understand what you want to do. To be honest I'm a bit hesitant to expand the arguments of `Map` since it's already quite crowded. Lately we added `zoom_control` because that's an actual map feature. I'm not sure about `name`, it's a generic term and it's meant for the `TileLayer` that's created.

There is an alternative way to do what you want: add a `TileLayer` yourself.
```
m = folium.Map(tiles=None)
m.add_child(folium.TileLayer(tiles=tiles, name=name,
                               min_zoom=min_zoom, max_zoom=max_zoom,
                               max_native_zoom=max_native_zoom,
                               attr=attr, API_key=API_key,
                               detect_retina=detect_retina,
                               subdomains=subdomains,
                               no_wrap=no_wrap)
```

Is that sufficient for you? Or do you want to make the case for adding a `name` argument to `Map`?

mccarthyryanc:
That makes sense. I was adding as `TileLayer` but sometimes found it nicer to have the custom tileset as the "basemap" instead of as an added layer. Perhaps a better option is to display a truncated URL? Something like `http://{s}.yourtiles.com/` instead of the full URL?

Conengmo:
I'm not sure if you see what I'm getting at. You can create a `Map` without a basemap by setting `tiles=None`. Then you can add a `TileLayer` where you can set the `name` argument as you wish.

```
m = folium.Map(tiles=None)
m.add_child(folium.TileLayer(tiles='mycustomtilesurl', attr='attribution', name='fancy name')
```

I do like your idea though to make a better default name when someone passes a tiles URL. Something truncated sounds good. What would be a sensible heuristic to determine the cut-off point? I'm thinking now maybe just do it after a fixed number of characters...

Do you want to alter this PR to implement this? The line you're looking for is in `raster_layers.py`:

```
self.tile_name = (name if name is not None else
                          ''.join(tiles.lower().strip().split()))
```

I would like to keep the functionality for the included tilesets the same so you'd have to find a way to detect whether the given `tiles` are an URL or not.

Conengmo:
I'm closing this PR since we're not going to add a `name` argument to `Map` that is passed on to `TileLayer`. 

I still think the best way to go here is to create a `TileLayer` object yourself and add it to a map which has no tiles: `Map(tiles=None).add_child(TileLayer(name='my name'))`

A useful enhancement of folium would be to have a better default name when passing URLs. You or someone else is welcome to open an issue for this, or even better a PR!

ocefpaf:
@Conengmo and @mccarthyryanc I'm thinking that maybe we can reduce the number of args in the `Map` call and also give users a powerful way to edit the `attribution` and `name` of a custom URL by taking a dictionary rather than separate args.

That way we can also get rid of `folium/folium/templates/tiles/` by providing some short-cuts to the built-in tiles via dicts too.

@Conengmo what do you think?

PR #908, opened 2018-07-13
Warn for pending deprecation of Python 2.7

Conengmo:
I propose we drop support for Python 2.7 at the end of this year. That's one year before official support stops, but other packages are also dropping support before that time. Numpy for example, one of our dependencies, stops development for 2.7 at the end of this year. Check out https://python3statement.org/

I don't know how many of our users are still on 2.7. I suppose not that many, since I don't think folium is a library that's used in heavy production code that would be hard to migrate to Python 3. But still, it's good manners to give some time to warn before dropping support I suppose. I set the actual date on 1-1-2019, but that's debatable.

There is probably not a lot that will change in the code base when we drop support. Still, I would like to not have to think about and test for 2.7 anymore.

@ocefpaf do you want to comment on this proposal?


folium/__init__.py
```diff
@@ -29,6 +32,14 @@
     raise ImportError('branca version 0.3.0 or higher is required. '
                       'Update branca with e.g. `pip install branca --upgrade`.')
 
+if sys.version_info < (3, 0):
+    warnings.warn(
+        ("folium will stop working with Python 2.7 starting Jan. 1, 2019."
+         " Please transition to Python 3 before this time."
+         " Check out https://python3statement.org/ for more info."),
+        PendingDeprecationWarning
```
ocefpaf: I never used `PendingDeprecationWarning` but this seems to not print the warning when importing the library. Is that expected?

ocefpaf: Thanks for the clarification.

Conengmo:
I came across a tool to check Pypi statistics, so I used it to check out what version our users are using.

`pypistats python_minor --last-month folium`

```
| category | percent | downloads |
|----------|--------:|----------:|
|      2.7 |  54.35% |    51,822 |
|      3.5 |  20.42% |    19,469 |
|      3.6 |  15.81% |    15,072 |
|      3.7 |   4.13% |     3,934 |
|      3.4 |   3.29% |     3,138 |
| null     |   2.01% |     1,912 |
|      2.6 |   0.00% |         2 |
| Total    |         |    95,349 |
```

So I guess half our use base will be in trouble if we drop support for Python 2.7?

jtbaker:
Wow, I had no idea the numbers were so high! Cool to know it's getting used by so many people.

Guessing the Python 2.7 majority is coming from the fact that that's the base installation that comes shipped on most people's machines(I know that's the case for Macs at least).

I don't have a lot of experience backporting to support legacy code/languages, but are some of the breaking changes something that could be averted by adding  `__future__` as a dependency, and a check on `sys.version_info.major` at run time?

Something like this in folium's `__init__.py`:
```
py_major_version = sys.version_info.major

if py_major_version is 2:
    from builtins import *
```

https://python-future.org/imports.html

Conengmo:
> Cool to know it's getting used by so many people.

Right! I don't know how trustworthy those numbers are but I'm liking them :)

> Are there cases where these imports are not providing full coverage for the language's features - or changes waiting to be implemented that will be breaking changes to the API?

No, folium is fully compatible with 2.7 right now, and no breaking changes are waiting.

> Are there Python 2.7 - specific bugs getting filed on the issue tracker right know? 

Not that I know of.

> Or is this just kind of a well-mannered phase-out?

Yes, that's it. If think maintaining will be easier and more fun if we can use Python 3.5 and higher features. I wasn't expecting so many users to still be on 2.7 though. If we drop support now half our use base cannot upgrade. 

I'm now thinking we maybe should extend support for half a year, but increase the warning level to a user warning, as is also recommended here:
https://github.com/python3statement/python3statement.github.io/blob/master/_practicalities/intro.md#runtime-warning-on-master

What do you think?

PR #912, opened 2018-07-17
Geo json markers

jtbaker:
Did some work on `GeoJson` and `Marker`/`Icon`/vector layers classes over the weekend to allow plotting of custom markers on the map for point geometries within a folium.GeoJson, as `L.geoJson` has a `pointToLayer` function in its callback that allows for Marker conversion from the default.

In order to do this, some of the structural changes I've made have been:

- to allow `location` to be passes as `None` for each of these classes.
- to update `self._name` in the `Marker`, `Icon` and `Vector` classes to be tightly coupled, case and spacing wise, with it's `L.<counterpart>` in Leaflet. In general, I think folium's naming for classes and arguments should be tightly coupled with Leaflet's, to have less transpilation for each class, to be able to drop all parameters into a template model directly with `json.dumps(**kwargs)`, and to lean on `Leaflet`'s documentation, since that is what we are writing to anyways.

Here's an example [use case](http://nbviewer.jupyter.org/github/jtbaker/NotebookExamples/blob/master/GeoJSONMarkerDemo.ipynb)

This PR does have a dependency on the Tooltip class I'm working on right now, but I wanted to go ahead and handle the tooltipping updates in this one since I was working in those classes anyways.

Right now it has the same marker representing every object, but this could style objects differently with different colors/icons if we passed another style dictionary through to the GeoJson `pointToFeature` function to query against `feature.properties` values. I'm not sure of the best practices to implement that - if it would be best approached by building another class, or having properties values mapped to style values in a dictionary/JSON object.

To me it is more idiomatic to keep data in a central GeoJson structure mapped on the map, and have central JS functions do the styling for each object instead of hardcoding style parameters and adding each point to the map, especially as the scale of the data grows.

Just wanted to get some eyes on it and some feedback/suggestions.


folium/vector_layers.py
```diff
@@ -12,7 +12,7 @@
 from branca.element import (CssLink, Element, Figure, JavascriptLink, MacroElement)  # noqa
 from branca.utilities import (_locations_tolist, _parse_size, image_to_url, iter_points, none_max, none_min)  # noqa
 
-from folium.map import Marker
+from folium.map import Marker, Tooltip
```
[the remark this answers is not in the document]
jtbaker: Tooltip class is referenced in `Marker.validate_tooltip()` function.

folium/map.py
```diff
@@ -11,7 +11,8 @@
 
 from collections import OrderedDict
 
-from branca.element import CssLink, Element, Figure, Html, JavascriptLink, MacroElement  # noqa
+from branca.element import CssLink, Element, Figure, Html, JavascriptLink, \
```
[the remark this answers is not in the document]
jtbaker: I didn't add these imports, but it is throwing a lint error. Should JavascriptLink and CssLink remain?

folium/vector_layers.py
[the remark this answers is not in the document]
jtbaker: Tooltip class is referenced in Marker.validate_tooltip() function.

jtbaker:
@ocefpaf @Conengmo have y'all had a chance to take a look at this? I'll need to wait for #883 to merge and possibly refactor a little bit, but wanted to check if you had any feedback on this idea.

folium/features.py
```diff
@@ -72,14 +72,17 @@ class RegularPolygonMarker(Marker):
...
-                )
+                ){% if this.tooltip %}
+                .bindTooltip('{{ this.tooltip }}')
+                {% endif %}
                 .addTo({{this._parent.get_name()}});
             {% endmacro %}
             """)
 
-    def __init__(self, location, color='black', opacity=1, weight=2,
+    def __init__(self, location=None, color='black', opacity=1, weight=2,
```
ocefpaf: Why making location option?

folium/features.py
```diff
@@ -103,7 +116,9 @@ def render(self, **kwargs):
                                             'if it is not in a Figure.')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),  # noqa
+            JavascriptLink(
+                'https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),
```
ocefpaf: I believe there is a newer version but you don't have to use it here. It can be done in another PR.

folium/features.py
```diff
@@ -182,11 +197,15 @@ def render(self, **kwargs):
             """).render(this=self, **kwargs)), name=self.get_name())
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js'),  # noqa
+            JavascriptLink(
+                'https://cdnjs.cloudflare.com/ajax/libs/d3/3.5.5/d3.min.js'),
+            # noqa
```
ocefpaf: No need for these break lines. Shifting code around makes it hard to review.

folium/features.py
```diff
@@ -588,12 +666,16 @@ def get_bounds(self):
                 ymax = none_max(y, ymax)
         return [
             [
-                self.data['transform']['translate'][1] + self.data['transform']['scale'][1] * ymin,  # noqa
-                self.data['transform']['translate'][0] + self.data['transform']['scale'][0] * xmin  # noqa
+                self.data['transform']['translate'][1] +
+                self.data['transform']['scale'][1] * ymin,  # noqa
+                self.data['transform']['translate'][0] +
+                self.data['transform']['scale'][0] * xmin  # noqa
```
ocefpaf: Hard to see if there is anything new here. If this is only line breaks please revert.

folium/map.py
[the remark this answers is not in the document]
ocefpaf: Safe to ignore. I need to check if other places in `folium` import from here before removing these.

folium/map.py
```diff
@@ -228,33 +239,49 @@ class Marker(MacroElement):
     >>> Marker(location=[45.5, -122.3], popup=folium.Popup('Portland, OR'))
     # If the popup label has characters that need to be escaped in HTML
     >>> Marker(location=[45.5, -122.3],
-               popoup=folium.Popup('Mom & Pop Arrow Shop >>', parse_html=True))
```
ocefpaf: Good catch!

ocefpaf:
@jtbaker I'll take a better look again soon but overall looks OK. My only request is to avoid unnecessary changes, like the line breaks, and send smaller PRs, to make it easier to review. (You can implement that for one class and then apply to all others after the first review.)

ocefpaf:
@Conengmo I'd appreciate if you could take a look at this one too. I'm away for so long and you are probably more familiar with the code base than me at the moment.

folium/features.py
jtbaker: I think for these I had moved them around to try and keep everything in compliance with the <80 character line style guideline.

ocefpaf:
> This is more easily managed with json.dumps(**kwargs) 

Indeed we should use that as much as we can. I started it some time ago but failed to continue the change. Thanks for picking it up!

I don't want to camelCase the python keywords though.

jtbaker:
> I don't want to camelCase the python keywords though.

Yes, I only meant the ones where they pair to a Leaflet value directly - mainly `kwargs` instead of named args.

PR #914, opened 2018-07-19
Change default date_options to 24 hour clock in TimestampedGeoJson

andy23512:
According to the document of moment.js https://momentjs.com/docs/#/displaying/format/ , `hh` shows hours in 01 to 12, while `HH` shows hours in 00 to 23.
The original code use `hh` to display hour, and it causes the time `14:00` shown as `02:00`. It's a pretty weird and misleading default option.
I think displaying 24-hour clock with `HH` is a better choice.
Any suggestion or anything else I need to do?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -19,6 +19,8 @@
 - Added `FeatureGroupSubGroup` plugin (shtrom #875)
 - Added `duration` option to `TimestampedGeoJson` (andy23512 #894)
 - Added `zoom_control` to `Map` to toggle zoom controls as per enhancement (#795) (okomarov #899)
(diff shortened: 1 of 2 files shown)


Conengmo:
That makes sense, we can merge this. Two things:

- Please add a line to the changelog.
- While you're at it, maybe also change the date format to `YYYY-MM-DD`? That makes it even more ISO8601 compatible.

andy23512:
Thanks for the suggestions.
The date format has been refined, and the change information has been added in changelog.

Conengmo:
Merged! Thanks for the PR @andy23512.

PR #916, opened 2018-07-19
Mouse Position plugin

btozer:
Original plugin from: Ardhi Lukianto
https://github.com/ardhi/Leaflet.MousePosition 
MIT Licence
Copyright 2012 Ardhi Lukianto

### DIFF
--- a/folium/plugins/mouse_position.py
+++ b/folium/plugins/mouse_position.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import (absolute_import, division, print_function)
+
+import json
+
+from branca.element import CssLink, Figure, JavascriptLink, MacroElement
+
+from jinja2 import Template
+
+
+class MousePosition(MacroElement):
+    """
+    Adds a mouse position widget on the map.
+
+    Parameters
+    ----------
+    position: location of the widget
+        default is 'bottomright'.
+
+    CONVERTED TO PYTHON FROM:
+
+    https://github.com/ardhi/Leaflet.MousePosition
+
+    With the MIT License as below:
+
+    Copyright 2012 Ardhi Lukianto
+
+    Permission is hereby granted, free of charge, to any person obtaining
+    a copy of this software and associated documentation files (the
+    "Software"), to deal in the Software without restriction, including
+    without limitation the rights to use, copy, modify, merge, publish,
+    distribute, sublicense, and/or sell copies of the Software, and to
+    permit persons to whom the Software is furnished to do so, subject to
+    the following conditions:
+
+    The above copyright notice and this permission notice shall be
+    included in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+    EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+    LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+    OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+    WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+    """
+    _template = Template("""
+        {% macro script(this, kwargs) %}
+            var {{this.get_name()}} = new L.Control.MousePosition(
+            {{ this.options }});
+            {{this._parent.get_name()}}.addControl({{this.get_name()}});
+
+        {% endmacro %}
+        """)  # noqa
+
+    def __init__(self, position='bottomright', separator=' : ', emptyString='Unavailable', lngFirst=False, numDigits=5,
+                 lngFormatter=None, latFormatter=None, prefix=""):
+        """Coordinate, linear, and area measure control"""
+        super(MousePosition, self).__init__()
+        self._name = 'MousePosition'
+
+        options = {
+            'position': position,
+            'separator': separator,
+            'emptyString': emptyString,
+            'lngFirst': lngFirst,
+            'numDigits': numDigits,
+            'lngFormatter': lngFormatter,
+            'latFormatter': latFormatter,
+            'prefix': prefix,
+        }
+        self.options = json.dumps(options)
+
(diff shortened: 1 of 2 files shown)


folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,87 @@
...
+    Parameters
+    ----------
+    position: location of the widget
+        default is 'bottomright'.
+
+    CONVERTED TO PYTHON FROM:
+
+    https://github.com/ardhi/Leaflet.MousePosition
+
+    With the MIT License as below:
```
Conengmo: Good that you include the license. But could you put in in the file header? The docstring is converted to documentation and I don't want that littered with license texts.

folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,87 @@
...
+    _template = Template("""
+        {% macro script(this, kwargs) %}
+            var {{this.get_name()}} = new L.Control.MousePosition(
+            {{ this.options }});
+            {{this._parent.get_name()}}.addControl({{this.get_name()}});
+
+        {% endmacro %}
+        """)  # noqa
+
+    def __init__(self, position='bottomright', separator=' : ', emptyString='Unavailable', lngFirst=False, numDigits=5,
```
Conengmo: Please make this line 80 chars wide.

folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,87 @@
...
+from jinja2 import Template
+
+
+class MousePosition(MacroElement):
+    """
+    Adds a mouse position widget on the map.
+
+    Parameters
+    ----------
+    position: location of the widget
```
Conengmo: Docstring convention is name, type and default value on the first line, description on the second. So something like:

```
position : str, default 'bottomright'
    Location of the widget.
```

folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,87 @@
...
+            var {{this.get_name()}} = new L.Control.MousePosition(
+            {{ this.options }});
+            {{this._parent.get_name()}}.addControl({{this.get_name()}});
+
+        {% endmacro %}
+        """)  # noqa
+
+    def __init__(self, position='bottomright', separator=' : ', emptyString='Unavailable', lngFirst=False, numDigits=5,
+                 lngFormatter=None, latFormatter=None, prefix=""):
+        """Coordinate, linear, and area measure control"""
```
Conengmo: No need for a docstring on a class init when you already have the docstring above.

Conengmo:
Thanks for this PR, it's a useful plugin! Check out my review comments for some details. Here are some longer comments:

**Choose between specifying all arguments or using `kwargs`**
- Current convention in folium is to use Python style variable names throughout. That would mean that for example the arguments `emptyString` and `lngFirst` should become `empty_string` and `lng_first` up until they're added to the options dictionary.
- Any variable that's an explicit argument should also be listed in the docstring parameters. So apart from `position` the other arguments should also be listed.

The fact you only included `position` in the docstring says something about the usefulness of the other arguments. You could decide to leave the other arguments out, and instead accept `**kwargs` and add those to the `options`. Then users can check out the original github project to see what those arguments are. Saves work, but is a bit lazy :) What do you want to do?

**Add an example**

Our current way of feature discovery is through an example gallery in Jupyter Notebooks. Please add an entry which shows how to use this plugin and its effect.

**Optionally add a test**

Not a strict necessity, but it would be nice if you can add a test for this module. You can check out the other plugin tests to see how they verify if the template is rendering correctly.

btozer:
Finally got around to this:

**Choose between specifying all arguments or using kwargs**

_Current convention in folium is to use Python style variable names throughout. That would mean that for example the arguments emptyString and lngFirst should become empty_string and lng_first up until they're added to the options dictionary._
**-> Done**

_Any variable that's an explicit argument should also be listed in the docstring parameters. So apart from position the other arguments should also be listed._
**-> Done**

_The fact you only included position in the docstring says something about the usefulness of the other arguments. You could decide to leave the other arguments out, and instead accept **kwargs and add those to the options. Then users can check out the original github project to see what those arguments are. Saves work, but is a bit lazy :) What do you want to do?_
**-> Added all the parameters to the docstring.**

**Add an example**

_Our current way of feature discovery is through an example gallery in Jupyter Notebooks. Please add an entry which shows how to use this plugin and its effect._
**-> Done**

**Optionally add a test**

_Not a strict necessity, but it would be nice if you can add a test for this module. You can check out the other plugin tests to see how they verify if the template is rendering correctly._
**-> Skipped this**

**Changelog**

_Finally, just before merging, you should add a line in the changelog. Here's an example line from another plugin that was added:_
**-> Done**

Conengmo:
I made two changes:

- The formatter option didn't work, because the functions were passed as string, where they should be Javascript functions. So instead I load them directly into the template. Note that in a Jupyter notebook this still gives an error, but I'm not really interested in solving that :) Instead I added an example in the docstring.
- In the example notebook I put the second example in a new map, because the two MousePosition objects were interfering. I noticed that because the `lng_first` option was not working.

Maybe you have anything to add to this? Otherwise I'll go merge this in a few days.

folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+
+# CONVERTED TO PYTHON FROM: https://github.com/ardhi/Leaflet.MousePosition
```
ocefpaf: You are using the plugin and not really converting it to Python.
I'm not a lawyer but I don't think this is necessary.

folium/plugins/mouse_position.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def render(self, **kwargs):
+        super(MousePosition, self).render()
+
+        figure = self.get_root()
+        assert isinstance(figure, Figure), ('You cannot render this Element '
+                                            'if it is not in a Figure.')
+
+        figure.header.add_child(
+            JavascriptLink('https://cdn.rawgit.com/ardhi/Leaflet.MousePosition/c32f1c84/src/L.Control.MousePosition.js'))  # noqa
```
ocefpaf: We are already using `noqa` so no need for the extra line here.

ocefpaf:
> Maybe you have anything to add to this? Otherwise I'll go merge this in a few days.

I have just one monor comment about the license header added and I'm +1 to merge this.

folium/plugins/mouse_position.py
```diff
@@ -34,7 +10,10 @@
 
 
 class MousePosition(MacroElement):
-    """
+    """Add a field that shows the coordinates of the mouse position.
+
+    Uses the Leaflet plugin by Ardhi Lukianto under MIT license.
+    https://github.com/ardhi/Leaflet.MousePosition
```
ocefpaf: :+1:

Issue #804, opened 2017-12-29
HeatMapWithTime change gradient color


sararl:
I have created a HeatMapWithTime using folium and would like to adjust the gradient/color of the heatmap, but I don't know how to do that. It is possible to change colors working with the simple heatmap (parameter gradient) . Thank you.

Conengmo:
`HeatMapWithTime` currently doesn't expose an option to set the colors. I believe this should be possible with the `gradient` argument, just like in the `HeatMap` plugin. Based on [this documentation](https://www.patrick-wied.at/static/heatmapjs/docs.html) we can add a `gradient` option somewhere in the options.

If somebody wants to work on this, a PR is welcome!

Conengmo:
The `gradient` arguments has been added in PR #925.

PR #925, opened 2018-07-31
Added gradient to HeatMapWithTime.

jtbaker:
Fix for #804, gradient arg is now added to HeatMapWithTime.


Conengmo:
Looks good to me! I was curious what color formats were accepted, so after looking into that I added that info to the docstring. Also did some minor cosmetic changes, and updated the test for this module.

The test for Python 2.7 fails, but it seems unrelated to this PR. It's ready to merge IMO. I'll wait a day though for if you or someone has additional comments.

Conengmo:
Merged. Thanks @jtbaker!

PR #926, opened 2018-08-01
Ignore the failing PNG test on Python 2.7

Conengmo:
Currently Travis fails all PR's because `test_valid_png()` fails in Python 2.7. Something with a temporary file being closed prematurely in the Selenium phantomjs driver.  I don't think it's worth trying to fix this since:

* it seems an issue in Selenium
* Python 2.7 should be deprecated
* phantomjs has already been deprecated

So I think the best solution is to ignore the test for now in Python 2.7.

### DIFF
--- a/tests/test_repr.py
+++ b/tests/test_repr.py
@@ -8,6 +8,7 @@ Folium _repr_*_ Tests
 
 from __future__ import (absolute_import, division, print_function)
 
+import sys
 import io
 
 import PIL.Image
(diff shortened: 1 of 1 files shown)


Conengmo:
That did the trick, Travis check is passing again. I'll wait a day before merging in case someone has a comment on this.

ocefpaf:
Fine with me. We have to move away from selenium and py27 anyway.
Thanks!

PR #933, opened 2018-08-12
Dual map plugin

Conengmo:
I made a plugin using the Leaflet.Sync plugin: https://github.com/jieter/Leaflet.Sync

You can create two maps in the same screen, where position and zoom level are synchronized. This is handy when you want to compare two datasets that don't display well when put on the same map.

I tried to make this as close to the regular `Map` class as possible.

- You can use the same arguments on initialization as `Map` (except the positioning arguments). 
- If you add objects to `DualMap`, they are added to both the submaps. Data is then duplicated.
- The order in which items are added is maintained in both maps.
- The individual maps are accessible by the attributes `DualMap.m1` and `DualMap.m2`.
- Adding `LayerControl` before saving/rendering works.

Here's a minimal working example:

```
m = folium.plugins.DualMap(location=(0, 0), tiles='cartodbpositron',  zoom_start=5)
fg_both = folium.FeatureGroup(name='markers_both').add_to(m)
fg_1 = folium.FeatureGroup(name='markers_1').add_to(m.m1)
fg_2 = folium.FeatureGroup(name='markers_2').add_to(m.m2)
folium.Marker((0, 0), popup='both').add_to(fg_both)
folium.Marker((0, 1), popup='1').add_to(fg_1)
folium.Marker((1, 0), popup='2').add_to(fg_2)
m.save('_map.html')
```

I would like to get some feedback on this before adding tests and an example in the notebooks.


folium/plugins/dual_map.py
```diff
@@ -0,0 +1,114 @@
...
+                children_new[subitem.get_name()] = subitem
+            item._children = children_new
+        return item
+
+    def render(self, **kwargs):
+        figure = self.get_root()
+        assert isinstance(figure, Figure), ('You cannot render this Element '
+                                            'if it is not in a Figure.')
+
+        figure.header.add_child(JavascriptLink('https://rawgit.com/jieter/Leaflet.Sync/master/L.Map.Sync.js'),  # noqa
```
ocefpaf: I believe that rawgit is no longer a viable option, right?

Conengmo: Right, I'll fix that.

Conengmo:
Thanks for reviewing this @ocefpaf, nice that you approve. I'll fix the rawgit thing and look into making tests and an example notebook. But then I'll get it to work in a notebook first :)

ocefpaf:
> Thanks for reviewing this @ocefpaf, nice that you approve. I'll fix the rawgit thing and look into making tests and an example notebook. But then I'll get it to work in a notebook first :)

No rush. Just ping me again when you are done.

Conengmo:
I think it's done. Here is an example notebook (and the entry in the Plugins notebook):
https://nbviewer.jupyter.org/github/conengmo/folium/blob/sync/examples/plugin-DualMap.ipynb
https://nbviewer.jupyter.org/github/conengmo/folium/blob/sync/examples/Plugins.ipynb#DualMap

@ocefpaf do you want to review?

Issue #937, opened 2018-08-22
Folium cannot read VectorTiles


yukexie:
import folium
from folium import plugins

vectorTilesPBF= "http://openapi-dev.aurin.org.au/public/gwc/service/wmts?REQUEST=GetTile&SERVICE=WMTS&VERSION=1.0.0&LAYER=aurin:GCCSA_2011_AUST&STYLE=&TILEMATRIX=EPSG:900913:{z}&TILEMATRIXSET=EPSG:900913&FORMAT=application/x-protobuf;type=mapbox-vector&TILECOL={x}&TILEROW={y}"
    
vectorTilesJSON= "http://openapi-dev.aurin.org.au/public/gwc/service/wmts?REQUEST=GetTile&SERVICE=WMTS&VERSION=1.0.0&LAYER=aurin:GCCSA_2011_AUST&STYLE=&TILEMATRIX=EPSG:900913:{z}&TILEMATRIXSET=EPSG:900913&FORMAT=application/json;type=geojson&TILECOL={x}&TILEROW={y}"
     
rasterTiles= "http://openapi-dev.aurin.org.au/public/gwc/service/wmts?REQUEST=GetTile&SERVICE=WMTS&VERSION=1.0.0&LAYER=aurin:GCCSA_2011_AUST&STYLE=&TILEMATRIX=EPSG:900913:{z}&TILEMATRIXSET=EPSG:900913&FORMAT=image/png&TILECOL={x}&TILEROW={y}"     

m= folium.Map(tiles=None, location=[-37, 140],
           zoom_start=6)

m.add_tile_layer(vectorTilesPBF, attr="test")
m'''

Hi, I tried the code above to read different format of spatial data, it works well with raster tiles but showing nothing with geojson and PBF format, not even the base map. 

I saw leaflet mapboxvectortiles only available in 0.7.x version and I tried previous folium version 1.5 that fit leaflet 0.7.0 as well but still doesn't work. 

Is there any plugins in folium can deal with these vector tiles?

Thanks

Issue #563, opened 2016-10-27
Support for Leaflet VectorGrid


adamkgoldfarb:
Hello! This is more of a question or feature request than an issue, but are there plans to eventually support [VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid)? 

I'm happy to work on an implementation if @BibMartin or @ocefpaf have suggestions for where to start.

Thanks!

ocefpaf:
There are no plans but that would be an awesome addition!

> I'm happy to work on an implementation if @BibMartin or @ocefpaf have suggestions for where to start.

Take a look at the plugins and `GeoJSON` implementations. I will investigate this further next week to give you more than just that :grimacing:

adamkgoldfarb:
Thanks, will do!

adamkgoldfarb:
Circling back to this and hope to take a look this weekend, and just a note-- THANKFUL FOR @ocefpaf @BibMartin and the rest of the dev team! Happy Thanksgiving!

ocefpaf:
> THANKFUL FOR @ocefpaf @BibMartin and the rest of the dev team! Happy Thanksgiving!

We all the devs are thankful for all folium users Issues and PRs. Happy :turkey: day!

ocefpaf:
PR https://github.com/python-visualization/folium/pull/943 took a stab at it and it seems to be working. I would welcomed it as an external plugin but I'm not sure that should be in the core `folium` though.

PR #943, opened 2018-08-28
Support for Vector Grid

mumeraltaf:
Addressing #937 and #563. Added a function to add vector tiles as base layer (namely protobuf) to a map.
Using [VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid) plugin.
Usage Example: examples/VectorGrid.ipynb


ocefpaf:
@mumeraltaf sorry for the long time to review this. Nice stuff and great feature you have here but have you considered creating an external plugin and depend on folium instead?

The reason I say this is b/c we won't be able to maintain the code. I would also recommend you pinning the `folium` version to ensure the plugin won't break with new updates. (Something you cannot do if this is part if the main `folium`.)

Conengmo:
I think this PR is nicely done, code looks good. I'm inclined to include this in the repository because:

- It's not that heavy in code
- The underlying library is in the Leaflet github organization
- and seems to be actively maintained

I would propose to move it to it's own plugin module though, since the appeal is not that broad (I assume) and technically it is a plugin.

Also the two methods in `Map` should be removed.

Here are links to the two notebooks, which aren't working it seems by the way:
http://nbviewer.jupyter.org/github/AURIN/folium/blob/master/examples/VectorGrid.ipynb
http://nbviewer.jupyter.org/github/AURIN/folium/blob/master/examples/VectorChoropleth.ipynb

RenoFilla:
OK, so has this functionality been picked up anywhere yet? I like to display airspace using OpenAIP (https://maps.openaip.net) which is a vector tile layer as far as I can see: 

var OpenAIP = L.tileLayer('http://{s}.tile.maps.openaip.net/geowebcache/service/tms/1.0.0/openaip_basemap@EPSG%3A900913@png/{z}/{x}/{y}.{ext}', {
	attribution: '<a href="https://www.openaip.net/">openAIP Data</a> (<a href="https://creativecommons.org/licenses/by-sa/3.0/">CC-BY-NC-SA</a>)',
	ext: 'png',
	minZoom: 4,
	maxZoom: 14,
	tms: true,
	detectRetina: true,
	subdomains: '12'
});

Conengmo:
This PR is very stale, so I'm going to close it now. This feature is still welcome, so if somebody wants to pick up the work here and address the comments (https://github.com/python-visualization/folium/pull/943#issuecomment-441462510) in a new PR that would be appreciated!

Conengmo:
This was also partially implemented in https://github.com/python-visualization/folium/pull/1576, available in the 0.13.0 release.

PR #948, opened 2018-08-31
TileLayer: Allow passing the tms argument

### LINKED ISSUE #858: Feature request: Expose "tms" argument for TileLayer
The leaflet TileLayer tms option (http://leafletjs.com/reference-1.3.0.html#tilelayer-tms) is not currently exposed in the folium.TileLayer object. This makes it so that certain sources for tiles cannot be displayed in a folium Map (e.g., those created by gdal2tiles.py).

I think this would be a fairly straightforward fix by adding a tms argument to the TileLayer __init__ and then adding it to the _options_ variable. e.g.:

```python
    def __init__(self, tiles='OpenStreetMap', min_zoom=0, max_zoom=18,
                 max_native_zoom=None, attr=None, API_key=None,
                 detect_retina=False, name=None, overlay=False,
                 control=True, show=True, no_wrap=False, subdomains='abc', tms=False):
        self.tile_name = (name if name is not None else
                          ''.join(tiles.lower().strip().split()))
        super(TileLayer, self).__init__(name=self.tile_name, overlay=overlay,
                                        control=control, show=show)
        self._name = 'TileLayer'
        self._env = ENV

        options = {'minZoom': min_zoom,
                   'maxZoom': max_zoom,
                   'maxNativeZoom': max_native_zoom or max_zoom,
                   'noWrap': no_wrap,
                   'attribution': attr,
                   'subdomains': subdomains,
                   'detectRetina': detect_retina,
                   'tms': tms}
```

mpickering:
In particular this is needed for the default output of gdal2tiles.py

Closes #858 

(I couldn't find any instructions how to build this project in the repo so I have not managed to build these changes. I will leave that up to CI.)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -2,6 +2,7 @@
 ~~~~~
 
 - Update leaflet to 1.3.4 (ocefpaf #939)
+- More options (tms, opacity, kwargs) in TileLayer (mpickering #948)
 
 
 0.6.0
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -68,7 +68,8 @@ class TileLayer(Layer):
     tms: bool, default False
         If true, inverses Y axis numbering for tiles (turn this on for TMS
         services).
-
+    opacity: float, default 1
+        Sets the opacity for the layer.
     """
     _template = Template(u"""
 {% macro script(this, kwargs) -%}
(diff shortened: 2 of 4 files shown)


Conengmo:
@ocefpaf the Travis builds are failing because it ocasionally gets an HTTP 500 from conda forge:

`CondaHTTPError: HTTP 500 INTERNAL SERVER ERROR for url <https://conda.anaconda.org/conda-forge/noarch/repodata.json>`

Can we do anything about that?

ocefpaf:
> Can we do anything about that?

Nothing besides wait :-(
Although the servers seems to be more responsive now, let me re-start the CIs and check.

Conengmo:
Luckily the servers did their work in the end, so waiting was indeed the best option :)

Thanks for working on this @mpickering!

PR #955, opened 2018-09-04
Tooltip Template string switch from single quotes to template literal.

jtbaker:
I noticed a bug in the template due to concatenating strings with single quotes which were interrupted with other possessive single quotes. Switching these to template literals should address the issue.
I noticed this with the following string concatenation example from a dataset:

 `'<div>'
            + 'City: Saint John's<br>Country: Antigua and Barbuda<br>Continent: North America' + '</div>'`

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -361,8 +361,8 @@ class Tooltip(MacroElement):
     _template = Template(u"""
         {% macro script(this, kwargs) %}
         {{ this._parent.get_name() }}.bindTooltip(
-            '<div{% if this.style %} style="{{ this.style }}"{% endif %}>'
-            + '{{ this.text }}' + '</div>',
+            `<div{% if this.style %} style="{{ this.style }}"{% endif %}>`
+            + `{{ this.text }}` + `</div>`,
             {{ this.options }}
         );
         {% endmacro %}


ocefpaf:
I'll prepare a bugfix release tomorrow. Thanks for the fix!

Conengmo:
I've been playing around with this and this is great stuff! Using this in `Popup` as well fixes multiple issues reported by people who hadn't figured out to use the `parse_html` argument. I included a list of issues in #850.

Apart from `Tooltip` and `Popup` I don't think this is needed in other places. In `GeoJsonTooltip` you already applied template literals.

ocefpaf:
> Apart from `Tooltip` and `Popup` I don't think this is needed in other places.

Good to know. Thanks for looking into it @Conengmo!

jtbaker:
@Conengmo awesome! one question - does this negate the need for `parse_html` arg in `folium.map.Popup`? Are there other use cases for it other than being able to interpolate string quotes/apostrophes into the template?

Conengmo:
> does this negate the need for `parse_html` arg in `folium.map.Popup`?

I don't think so. A possible use case if you want to show literal html in your text. (But who would want that? :p) Also, special characters seem to work for me, but there may be cases where you want them converted to html entities.

jtbaker:
I think some of the source of the confusion has to do with the naming of the keyword argument. When I read `parse_html` I read it as "display with html formatting", which anything in the template will do by default since its content writes an HTML document. 

I think it would be more clear to end users what it does if it was named something like `insert_as_raw_string` - or have the logic inverted and name the argument `render_as_html` and default to `True` in the constructor.

I'm personally fine with it either way as I don't need it a lot - but I think it might be more clear to people that could need to use it.

ocefpaf:
> I think it would be more clear to end users what it does if it was named something like `insert_as_raw_string` - or have the logic inverted and name the argument `render_as_html `

I like `insert_as_raw_string`.

>  and default to `True` in the constructor.

That was default but it caused other issues so we reverted it.
I need to take a look at the past issues and think about it a little bit.

PR #957, opened 2018-09-05
GeoJson Marker

### LINKED ISSUE #1059: control markers using GeoJson
Hi all,

I can plot markers using `GeoJson` or `Marker`. I like using GeoJson since I can plot them all at once without a loop, and also I love `GeoJson.Tooltip`, which looks great. Please see below for an example.

Unfortunately, I can't figure out how to change the icon, or other properties of the markers when using `GeoJson`. If I switch to looping and using `Marker`, I have more control over the markers but the tooltips don't look as good when specifying a list of values (I get a string representation).

Is there a way to change properties of the markers using the `GeoJson` method?

```python
data={'type': 'FeatureCollection',
 'features': [{'id': '0',
  'type': 'Feature',
  'properties': {'SCHOOL_NAME': 'Ascension Elementary School',
   'SCHOOL_CODE': 'ACSN',
   'ADDRESS': '5205 New Street',
   'CITY': 'Burlington',
   'PROV': 'ON',
   'POST': 'L7L 1V3',
   'COUNTRY': 'Canada',
   'style': {'markerColor': '#00000000', 'color': '#00000000'},
   'highlight': {}},
  'geometry': {'type': 'Point', 'coordinates': [-79.7491672, 43.371468]}},
 {'id': '2',
  'type': 'Feature',
  'properties': {'SCHOOL_NAME': 'Bishop P. F. Reding Secondary School',
   'SCHOOL_CODE': 'BHRD',
   'ADDRESS': '1120 Main Street East',
   'CITY': 'Milton',
   'PROV': 'ON',
   'POST': 'L9T 6H7',
   'COUNTRY': 'Canada',
   'style': {'markerColor': '#00000000', 'color': '#00000000'},
   'highlight': {}},
  'geometry': {'type': 'Point',
   'coordinates': [-79.86306090000001, 43.530966799999995]}}]}

m = folium.Map(location=[43.342094, -79.803047], zoom_start=12)
folium.GeoJson(data, name='Schools', tooltip=folium.features.GeoJsonTooltip(fields=['SCHOOL_NAME', 'ADDRESS'], localize=True)).add_to(m)
m
```
sc

Thanks so much for this incredible package.

Al

jtbaker:
Ok, I've reworked the GeoJsonMarkers PR to have the commits staged more logically so that they are easier to review. The class is works best with is CircleMarker - it works with normal Marker/Icons too, but since they do not have a `.setStyle` method they can't have styles set dynamically using `style_function`.

Closes #1059.


Conengmo:
Here is a link to the Jupyter notebook in nbviewer:
https://nbviewer.jupyter.org/github/jtbaker/folium/blob/geojsonmarker/examples/GeoJsonMarkersandTooltips.ipynb

folium/features.py
```diff
@@ -767,12 +795,20 @@ class DivIcon(MacroElement):
     def __init__(self, html=None, icon_size=None, icon_anchor=None,
                  popup_anchor=None, class_name='empty'):
         super(DivIcon, self).__init__()
-        self._name = 'DivIcon'
+        self._name = 'divIcon'
```
Conengmo: Currently almost all `_name` attributes in folium are UpperCamelCase. Why should it be lower?

folium/features.py
```diff
@@ -767,12 +795,20 @@ class DivIcon(MacroElement):
...
         self.icon_size = icon_size
         self.icon_anchor = icon_anchor
         self.popup_anchor = popup_anchor
         self.html = html
         self.className = class_name
+        self.options = {
```
Conengmo: I like that you're parsing these options in Python instead of the template, but the template edit is missing. Also, this should be a separate PR.

folium/features.py
```diff
@@ -901,6 +937,14 @@ def __init__(self, icon_image, icon_size=None, icon_anchor=None,
         self.shadow_size = shadow_size
         self.shadow_anchor = shadow_anchor
         self.popup_anchor = popup_anchor
+        self.options = {
```
Conengmo: Same as for divIcon. Also, `LatLngPopup` currently doesn't accept options.

folium/map.py
```diff
@@ -186,21 +186,31 @@ class Icon(MacroElement):
...
         super(Icon, self).__init__()
-        self._name = 'Icon'
+        self._name = 'AwesomeMarkers.icon'
         self.color = color
         self.icon = icon
         self.icon_color = icon_color
         self.angle = angle
         self.prefix = prefix
+        self.spin = spin
+        self.options = {
```
Conengmo: Same as for DivIcon

folium/map.py
```diff
@@ -186,21 +186,31 @@ class Icon(MacroElement):
...
+                    extraClasses: 'fa-rotate-{{this.angle}}',
+                    spin: {{this.spin}}
                     });
                 {{this._parent.get_name()}}.setIcon({{this.get_name()}});
             {% endmacro %}
             """)
 
     def __init__(self, color='blue', icon_color='white', icon='info-sign',
-                 angle=0, prefix='glyphicon'):
+                 angle=0, prefix='glyphicon', spin=False):
```
Conengmo: `spin` only works for FontAwesome icons?
https://github.com/lvoogdt/Leaflet.awesome-markers#properties

folium/map.py
```diff
@@ -186,21 +186,31 @@ class Icon(MacroElement):
...
         super(Icon, self).__init__()
-        self._name = 'Icon'
+        self._name = 'AwesomeMarkers.icon'
```
Conengmo: The current convention in folium is that `_name` is the class name, so `'Icon'`. Why should it be `'AwesomeMarkers.icon'`?

folium/map.py
```diff
@@ -243,13 +253,17 @@ class Marker(MacroElement):
...
-        self.tooltip = tooltip
-        self.location = _validate_coordinates(location)
+        self._name = 'marker'
+        if location is not None:
+            self.location = _validate_coordinates(location)
         if icon is not None:
             self.add_child(icon)
+            self.icon_type = icon._name
+            kwargs.update(icon=icon.options)
+        self.options = json.dumps(kwargs)
```
Conengmo: Same as DivIcon

folium/map.py
```diff
@@ -243,13 +253,17 @@ class Marker(MacroElement):
...
         """)
 
-    def __init__(self, location, popup=None, tooltip=None, icon=None):
+    def __init__(self, location=None, popup=None, tooltip=None, icon=None,
+                 **kwargs):
         super(Marker, self).__init__()
-        self._name = 'Marker'
-        self.tooltip = tooltip
-        self.location = _validate_coordinates(location)
+        self._name = 'marker'
```
Conengmo: Again, why change `_name`?

folium/map.py
```diff
@@ -243,13 +253,17 @@ class Marker(MacroElement):
...
+        if location is not None:
```
Conengmo: I get that you want to do this for using it in `GeoJson`. I'm worried though that people forget to provide coordinates, and then don't get an error but a blank map instead, since the error is happening in the browser. What about leaving the `location` argument without default value? So no `location=None`, but still accept that `location` may be `None` when used in `GeoJson`.

folium/vector_layers.py
```diff
@@ -322,8 +322,8 @@ class CircleMarker(Marker):
             {% endmacro %}
             """)
 
-    def __init__(self, location, radius=10, popup=None, tooltip=None, **kwargs):
+    def __init__(self, location=None, radius=10, fill=True, popup=None, tooltip=None, **kwargs):
```
Conengmo: Again, I would argue not to set `location=None` as default.

Conengmo: Why are you adding `fill` here explicitly? It's currently parsed from the `**kwargs` in `_parse_options()`, so no need to add it explicitely. Also, the default is currently not `True` but dependent on the `fill_color`. In `_parse_options()`, if `fill` is not provided it is set to `True` if there is a `fill_color` and `False` otherwise. So it can be used as `kwarg`.

folium/features.py
```diff
@@ -351,40 +354,63 @@ class GeoJson(Layer):
...
-                    {% endif %}
-                    onEachFeature: {{this.get_name()}}_onEachFeature
+                    onEachFeature: {{this.get_name()}}_onEachFeature,
+                {% endif %}
+                {% if this.marker %}
+                    pointToLayer: function (feature, latlng) {
+                        var opts = {{this.marker.options}};
+                        {% if this.marker._name =='marker' and this.marker.icon_type %}
+                        opts.icon=L.{{this.marker.icon_type}}(opts.icon){% endif %}
+                        return L.{{this.marker._name}}(latlng, opts)
```
Conengmo: I guess this is why you're changing the `_name` attributes to lowerCamelCase. Maybe it's better to convert here to the wanted format.

jtbaker:
Thanks for the review and feedback, @Conengmo! I'll work through your comments and address them inline.

jtbaker:
>I'm wondering how necessary it is to pass a folium Marker or CircleMarker object. You're not really using the objects, except for getting the options. But in your notebook it seems your main goal is to update these options for each feature/marker. So the default options get overwritten anyway. How would it be if you let users pass the wanted marker type as a string?

I think my main goal was to follow `folium`'s current convention with existing class structures how people are already familiar with using them. They can optionally pass the same styled marker for every feature point, style the marker based on the GeoJson properties, or a mixture of both: i.e. a base `radius` and `weight`, and then choose `fill_color` or additional properties in the `style_function` logic.

What do you think about this approach?

>How is point geometry currently handled by the way?

Without a `pointToLayer` function in the callback, leaflet renders points with the standard  Font-Awesome blue marker like the one in these examples: https://leafletjs.com/examples/quick-start/

folium/features.py
jtbaker: Goes for all subsequent mentions; so we can interpolate each class name cleanly into its Leaflet counterpart constructor in the Jinja2 template without having to create another redundant variable name or mess with a utility to handle lowercasing the first letter. It's not an actual variable name; just a string value that gets stored. I think keeping the API as clean and straightforward as possible for other people that contribute to or maintain the project is a boon.

folium/features.py
jtbaker: Fair enough. My goal is to have an options dict that gets `json.dump`ed for each class eventually to easily drop into the template callback, in lieu of hard coding each value. This will play nicer with **kwargs too. In a few of these I went ahead and created the dict, but didn't updated the template logic yet.

folium/features.py
jtbaker: Roger.

folium/map.py
jtbaker: Ok, perhaps better to relegate to **kwargs.

folium/map.py
jtbaker: Is there some way to check the type of the parent class the Marker is added to, and validate there? I guess the problem is there are so many different classes that these can be children of - `FeatureGroup`, `MarkerCluster`, directly to `Map`, etc. Would a `render` function, like many other classes have be appropriate? We could call it in any template where a marker is added to a child, and validate there.

It would be nice if the user did not have to explicitly pass `location = None` because it's not really intuitive at face value - it's not `None`, it just has many possible values instead of one. I get we can state it in the `GeoJson` docstring, but I'd like it to be clean and straightforward for users.

folium/vector_layers.py
jtbaker: My thought is that these would be a good opinionated default value for these Vector Marker layers.

I think because when I was first starting to use the API, this tripped me up several times as I wasn't yet familiar with Leaflet, and even when I passed a `color` argument, I could not get my popup/hover methods to display except in the outer radius - and I'm sure that other people have had similar experiences, as there is no documentation of the `fill`, `fill_color`, or `color` arguments in the docstring to help users make sense of how to use the class - they have to go to the Leaflet docs. (I probably could/should have added some annotations in the docstring. Maybe I should make another PR for this).

If a user is adding a marker to the Map, I think it's reasonable to assume they are interested in the interior of the vector and would like click/hover events to fire with the mouse inside of the marker, which will only happen if `fill` or `fill_color` has a value. I don't think we should pass a default `fill_color`, but I think `fill=True` makes sense unless the user really wants to set it `False`, in which case they still have the option. `fill=True` still leaves the vector interior transparent.

To me this is in the same vein higher level ease of use that folium does with many other things, like `tiles="OpenStreetMap"` in `Map`.

Let me know what you think.

ocefpaf:
@jtbaker this LGTM but I'll leave for @Conengmo to take a second look and merge b/c he is more familiar with the code base lately.

Thanks for the PR BTW! It is a nice improvement!

folium/features.py
jtbaker: the problem there isn't a clean string method idiomatic way to handle both styles of `markers`. `AwesomeMarkers` sublcasses their icon to `L.AwesomeMarkers.icon` while the svg vectors map to `L.circleMarker`, `L.circle`. Coupling `self._name` tightly to how the Leaflet constructors seems like the most straightforward way to do it without introducing another variable or conditional template logic.

jtbaker:
Thanks @ocefpaf! For the life of me, I can't figure out how to get the .xml files out of version control - I've tried reverting the commits, rebasing my master, etc, but they're still showing up. I'm not sure what to do - maybe I need to close my fork and start a new one.

folium/vector_layers.py
Conengmo: That makes sense, thanks for explaining! Maybe a tip for making our lives easier: this is really something for its own PR. You show the change (`fill=True`), explain why (which you did nicely here), and then it can be merged quickly without much reviewing.

folium/map.py
Conengmo: > Would a render function, like many other classes have be appropriate?

Sounds good. Add a `render` method to `Marker` that checks if `self.location` is not None. In your change the `marker` in `GeoJson` is not added as a child, so we don't even have to do any logic, since `marker` is not in the rendering tree.

Can you add this? If we do this at `Marker`, we got them all right? Since all other marker classes you set `location=None` inherit from `Marker` and don't have their own `render` method.

Conengmo:
> I think my main goal was to follow folium's current convention with existing class structures how people are already familiar with using them. They can optionally pass the same styled marker for every feature point, style the marker based on the GeoJson properties, or a mixture of both: i.e. a base radius and weight, and then choose fill_color or additional properties in the style_function logic.

Sounds good. I hope your example notebook will learn power users how to do more complicated stuff.

> I can't figure out how to get the .xml files out of version control

They seem already gone? They're not in the files changed list.

Thanks for replying on my comments. Some things have to be resolved:

- making sure the `_name` attributes are consistent throughout folium
- splitting of things that should be a separate PR
  - (I hope I'm not boring you with this, but I think it's important other users can easily find if something changed why that was done.)
- adding a check on `location is not None` in a new `Marker.render()` method

Your changes to the `GeoJson` template look good to me, but I'll have a better look this weekend.

Thanks for your patience @jtbaker! You come with interesting PR's so I hope you appreciate we taking the time to get it right.

FabeG:
Hi @jtbaker, this is a very interesting and useful PR ! Do you plan an update in order to merge it ?

pshassett:
Hi all. Really appreciate this project and would love to use these features...

@jtbaker I can see you put a lot of work into this.  How can we get it to pass test and get merged?

zoesn89:
> Hi all. Really appreciate this project and would love to use these features...
> 
> @jtbaker I can see you put a lot of work into this. How can we get it to pass test and get merged?

Can I help testing and if yes, how? ;)

jtbaker:
I got inspired by the interest in this PR and some of the issues #1059, #1345 and rebased/refactored against the current state of the repo. The decoupling of some of the pieces of the GeoJSON logic since I last worked on it made it much easier. I believe most of the issues that were raised when this was last discussed have been addressed, but I'd be happy to hear any additional feedback seen it has been quite some time.

I've also added a few tests and an example notebook with a few examples of usage. Hopefully if this is able to make it in, it's useful to someone.

Conengmo (approved):
Nice work Jason!

I made two changes: small code style changes and added an example with `Marker` to the notebook and ran the notebook.

I'll merge this when the tests passed.

jtbaker:
> Nice work Jason!
> 
> I made two changes: small code style changes and added an example with `Marker` to the notebook and ran the notebook. Update: also added a docstring.
> 
> I'll merge this when the tests passed.

Thanks for taking the time to review and your updates Frank! Glad this was finally able to get in.

lhoupert:
Hi guys, it seems that the test notebook is not available https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/GeoJSONMarker.ipynb

l-huff:
Hello!
I would also like to view the test notebook...

jtbaker:
@lhoupert @l-huff that [link](https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/GeoJSONMarker.ipynb) works for me. Do any of the other notebooks in the [examples](https://nbviewer.jupyter.org/github/python-visualization/folium/tree/master/examples/) directory work for you?

goanes:
Hi I see the example for the marker which work great.
in my GeoJson data, I have two type of point. 
I would like to customyse the icon depending on the type of the point/marker.

I first try to customyze the icon, but it did not work for me.

icon = folium.features.CustomIcon(icon_image="Flamme.PNG", icon_size =(15,15))
marker = folium.Marker(icon=icon)
folium.GeoJson(gdf, name="point", style_function=style_function, marker=marker ).add_to(m)

I have also try to put that in the style but I did not manage to do that
def style_function(feature):
    if 'Point' == feature.get('geometry').get('type'):
        icon = folium.Icon(icon="star")
        return {"markerColor" : "#FF0000", "markerIcon" : icon}


Any idea ?

PR #959, opened 2018-09-11
Use newer vega versions for altair compatibility

JarnoRFB:
Newer versions of vega, vega-lite and vega-embed are used to allow the 
usage of altair 2.x. Additionally,
the JavaScript for embedding is adapted. jsdelivr is used instead of
cloudflare, since it allows to specify the newest minor version very
nicely and it is recommended by the vega-lite docs
https://vega.github.io/vega-lite/usage/embed.html#cdn

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -232,7 +232,7 @@ class VegaLite(Element):
 
     def __init__(self, data, width=None, height=None,
                  left='0%', top='0%', position='relative'):
-        super(VegaLite, self).__init__()
+        super(self.__class__, self).__init__()
         self._name = 'VegaLite'
         self.data = data.to_json() if hasattr(data, 'to_json') else data
         if isinstance(self.data, text_type) or isinstance(data, binary_type):
@@ -256,13 +256,10 @@ class VegaLite(Element):
             """).render(this=self, kwargs=kwargs)), name=self.get_name())
 
         self._parent.script.add_child(Element(Template("""
-            var embedSpec = {
-                mode: "vega-lite",
-                spec: {{this.json}}
-            };
-            vg.embed(
-                {{this.get_name()}}, embedSpec, function(error, result) {}
-            );
+            const spec = {{this.json}};
+            vegaEmbed({{this.get_name()}}, spec)
+                .then(function(result) {})
+                .catch(console.error);
         """).render(this=self)), name=self.get_name())
 
         figure = self.get_root()
@@ -280,19 +277,19 @@ class VegaLite(Element):
             """).render(this=self, **kwargs)), name=self.get_name())
 
         figure.header.add_child(
-            JavascriptLink('https://d3js.org/d3.v3.min.js'),
+            JavascriptLink('https://d3js.org/d3.v5.min.js'),
             name='d3')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/vega/2.6.5/vega.min.js'),  # noqa
+            JavascriptLink('https://cdn.jsdelivr.net/npm/vega@3'),
             name='vega')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/vega-lite/1.3.1/vega-lite.min.js'),  # noqa
+            JavascriptLink('https://cdn.jsdelivr.net/npm/vega-lite@2'),
             name='vega-lite')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/vega-embed/2.2.0/vega-embed.min.js'),  # noqa
+            JavascriptLink('https://cdn.jsdelivr.net/npm/vega-embed@3'),
             name='vega-embed')
 
 


JarnoRFB:
Hi folium team,
just stumbled upon the old vega version, that caused plots to not work with altair 2.x. If you like the idea of updating, than this feature could probably use some tests.

folium/features.py
```diff
@@ -232,7 +232,7 @@ class VegaLite(Element):
 
     def __init__(self, data, width=None, height=None,
                  left='0%', top='0%', position='relative'):
-        super(VegaLite, self).__init__()
+        super(self.__class__, self).__init__()
```
Conengmo: Please keep this PR clean and don't change this here.

folium/features.py
```diff
@@ -280,19 +277,19 @@ def render(self, **kwargs):
             """).render(this=self, **kwargs)), name=self.get_name())
 
         figure.header.add_child(
-            JavascriptLink('https://d3js.org/d3.v3.min.js'),
+            JavascriptLink('https://d3js.org/d3.v5.min.js'),
             name='d3')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/vega/2.6.5/vega.min.js'),  # noqa
+            JavascriptLink('https://cdn.jsdelivr.net/npm/vega@3'),
```
Conengmo: I like this kind of cdn url

folium/features.py
```diff
@@ -256,13 +256,10 @@ def render(self, **kwargs):
...
         self._parent.script.add_child(Element(Template("""
-            var embedSpec = {
-                mode: "vega-lite",
-                spec: {{this.json}}
-            };
-            vg.embed(
-                {{this.get_name()}}, embedSpec, function(error, result) {}
-            );
+            const spec = {{this.json}};
+            vegaEmbed({{this.get_name()}}, spec)
```
Conengmo: It seems `vegaEmbed` is the function to use in vega embed version 3, while `vg.embed` was used in version 2, is that right?

folium/features.py
```diff
@@ -232,7 +232,7 @@ class VegaLite(Element):
 
     def __init__(self, data, width=None, height=None,
```
Conengmo: You're only updating `VegaLite` it seems, why not `Vega` also?

folium/features.py
JarnoRFB: I somehow got a strange error at some point, when I used `VegaLite` instead of `self.__class__`, but I can revert that.

folium/features.py
JarnoRFB: yes, that seems to be the case.

folium/features.py
JarnoRFB: good point, will do

JarnoRFB:
Thanks for the review! I also thought about leaving support for VegaLite 1, but when I tried it out, it seemed to be backwards compatible at least for a simple examples and I first wanted to find out if this feature is desired at all. I think the easiest way to keep backwards compatible support would be to query the `$schema` field of the vega lite spec. I will add this soon when I find the time.

JarnoRFB:
I finally found time to do this. Let me know what you think of the code, I will add a section describing the changes to the docs then. I now implemented this only for vegalite, as this is relevant for altair. I would prefer to take care of vega in a separate PR.

folium/features.py
```diff
@@ -112,7 +113,8 @@ def render(self, **kwargs):
                                             'if it is not in a Figure.')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),  # noqa
+            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),
+            # noqa
```
ocefpaf: Please keep this all in one line.

folium/features.py
ocefpaf: What was the error?

ocefpaf:
Something is not OK with the background, this is our notebook example with this PR:

screenshot from 2018-12-11 16-18-58

JarnoRFB:
Thanks for the review! I incorporated the comments. Regarding the background issue, I will probably wait until #1040 is merged. Then it might work as intended.

ocefpaf:
> Thanks for the review! I incorporated the comments. Regarding the background issue, I will probably wait until #1040 is merged. Then it might work as intended.

Yep. That is not caused by this PR.

folium/features.py
```diff
@@ -288,21 +282,63 @@ def render(self, **kwargs):
...
+            # Version 2 is assumed as the default, if no version is given in the schema.
+            self._embed_vegalite_v2(figure)
 
-        figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/vega/2.6.5/vega.min.js'),  # noqa
-            name='vega')
+    def _get_vegalite_major_versions(self, spec):
+        try:
+            schema = spec['$schema']
+            version = os.path.splitext(os.path.split(schema)[1])[0].lstrip('v')
```
ocefpaf: Do you need `os` to split `schema` here? I'm just thinking that we can avoid one extra import.

JarnoRFB: I think one could solve it just with `str.split`, but the `os` functions more direct and have somewhat speaking names. I also think importing `os` does not harm as it is in the standard library.

ocefpaf: I'm +0.5 on that. Indeed `os` is standard lib, but it is an extra import that is used for a simple task. I would also argue that .split('/') is more readble than `.path.splitext` which is intended for file system paths.

JarnoRFB: Convinced! Even needs one method call less this way.

folium/features.py
```diff
@@ -295,10 +294,10 @@ def render(self, **kwargs):
     def _get_vegalite_major_versions(self, spec):
         try:
             schema = spec['$schema']
-            version = os.path.splitext(os.path.split(schema)[1])[0].lstrip('v')
-            major_version = version.split('.')[0]
         except KeyError:
             major_version = None
+        else:
+            major_version = schema.split('/')[-1].split('.')[0].lstrip('v')
```
ocefpaf: 1-line and less code :smile: 
Thanks!

folium/features.py
```diff
@@ -288,21 +281,63 @@ def render(self, **kwargs):
...
+    def _embed_vegalite_v2(self, figure):
+        self._vega_embed()
+
+        figure.header.add_child(JavascriptLink('https://cdn.jsdelivr.net/npm/vega@3'), name='vega')
+        figure.header.add_child(JavascriptLink('https://cdn.jsdelivr.net/npm/vega-lite@2'), name='vega-lite')
+        figure.header.add_child(JavascriptLink('https://cdn.jsdelivr.net/npm/vega-embed@3'), name='vega-embed')
+
+    def _vega_embed(self):
+        self._parent.script.add_child(Element(Template("""
+                    const spec = {{this.json}};
```
ocefpaf: The `spec` here will need a unique name or we won't be able to display more than 1 object.

JarnoRFB: I found a simple fix by just avoiding using the `spec` variable. However, we should test against this case for the future. Unfortunately I have no idea how to do this, as the error is happening on the JS side.

ocefpaf: > I found a simple fix by just avoiding using the `spec` variable.

Great!

> However, we should test against this case for the future. Unfortunately I have no idea how to do this, as the error is happening on the JS side. Just executing a test with two VegaLite elements does not throw an error.
> Putting an example in the example notebooks might be a way, but could look a bit awkward and is also not caught by pytest.

Yeah, that is hard. We can create a map and compare an expected output but that may not catch errors like that. We do build all the maps in the example gallery and check them ~~always~~ (sometimes) before merging a PR.

JarnoRFB: Maybe one could start a headless selenium browser and check if the map and the charts are there. But that would be maybe a bit two much for this PR. I can easily provide a notebook example for now and try to tackle more elaborate testing in a separate PR if that is fine for you.

ocefpaf: That kind of testing is in our TODO list for a long time and I agree, it is something for another PR.

This one is good to go, just waiting for @Conengmo last :+1: to merge it.

PR #962, opened 2018-09-17
Use javascript template literal in Popup

Conengmo:
@jtbaker can you check if this works for you as well?


ocefpaf:
> Because the argument became less important with the template literals, only power users will need it. And because when in doubt, don't change the API. But I do agree it is not a very intuitive choice of words.

:+1:

Conengmo:
The Vega and Vega Lite charts were broken because I removed the jQuery code. Worked fine for normal text, but broke the charts. I restored that and now it seems fine. 

This PR is now quite trivial, so I'll wait a bit and then go ahead and merge this.

PR #966, opened 2018-09-18
Pattern

talbertc-usgs:
This is my first attempt to author a plugin.  It's working but I have a few concerns or questions.  

The main one is that I need to pass a JS object with the pattern to the style_function.  I did this by editing features.py to  type check for branca MacroElements and passing their name to the output instead of themselves.  Even still the name was getting passed enclosed in quotes, which I'm manually stripping for now.  Ideas on an alternate or cleaner way to accomplish this business  would be appreciated.

Also I wasn't quite sure how to structure the test.

here's the example:
http://nbviewer.jupyter.org/github/talbertc-usgs/folium/blob/pattern/examples/Pattern.ipynb


tests/plugins/test_pattern.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+"""
+Test BoatMarker
```
ocefpaf: Copy-n-pasta :smile:

folium/plugins/__init__.py
```diff
@@ -29,6 +29,7 @@
 from folium.plugins.timestamped_geo_json import TimestampedGeoJson
 from folium.plugins.timestamped_wmstilelayer import TimestampedWmsTileLayers
 from folium.plugins.search import Search
+from folium.plugins import pattern
```
ocefpaf: The API currently is to expect the plugin classes to be listed here and not another module, so let's change this to

```python
from folium.plugins.pattern import CirclePattern, StripePattern
```

folium/features.py
```diff
@@ -434,9 +434,20 @@ def style_data(self):
...
+                    new_style[item] = "{{'" + feature_style[item].get_name() + "'}}"
+                else:
+                    new_style[item] = feature_style[item]
+
+            feature.setdefault('properties', {}).setdefault('style', {}).update(new_style)  # noqa
             feature.setdefault('properties', {}).setdefault('highlight', {}).update(self.highlight_function(feature))  # noqa
-        return json.dumps(self.data, sort_keys=True)
+
+        # TODO Avoid having to manually remove the quotes from from around the placeholders
+        return json.dumps(self.data, sort_keys=True).replace('"{{', "{{").replace('}}"', "}}")
```
ocefpaf: @Conengmo is more familiar with all the changes in the `GeoJson` class, so I'll defer to him to review this part :grimacing:

Conengmo:
Hi @talbertc-usgs, I was checking out your PR and while doing that I made some changes:

* The pattern classes should always be added to a Map object.
* The pattern objects will be rendered in GeoJson.style_data() if they weren't added to anything. This means you no longer have to call `stripes.add_to(m)`, which I assume many people would forget to do :)
* The test_pattern() test now uses a local data file instead of getting it from a URL.

Since I also made some changes/simplifications to `GeoJson`, I would like for @ocefpaf or @jtbaker to have a look as well.

> Even still the name was getting passed enclosed in quotes, which I'm manually stripping for now. 

It's not perfect but for now this seems a good solution.

folium/utilities.py
```diff
@@ -415,3 +415,11 @@ def _tmp_html(data):
     finally:
         if os.path.isfile(filepath):
             os.remove(filepath)
+
+
+def get_obj_in_upper_tree(element, cls):
```
Conengmo: @jtbaker I reused that function from your Search plugin PR :)

jtbaker: Nice! Good to see a spot to fit for it. :-)

One suggestion - maybe some type checking and a bombproof base case to avoid infinite recursion. Like:

```
from branca.element import Figure

...

def get_obj_in_upper_tree(element, cls):
    """Return the first object in the parent tree of class `cls`."""
    if not hasattr(element, '_parent'):
        raise ValueError('{} does not have a parent object.'.format(element))
    if isinstance(element, Figure) and cls is not Figure:
        raise ValueError('The top of the tree was reached without finding a {}'.format(cls))
    parent = element._parent
    if not isinstance(parent, cls):
        return get_obj_in_upper_tree(parent, cls)
    return parent
```

Conengmo: Good one. I'll combine the two checks you suggested into one: when there's no more `_parent` the top of the tree was reached.

I don't think it will recurse infinitely because at some point there will be no more `_parent`. Maybe unless someone makes a cyclical graph. That would be interesting if someone manages to do that :) But then Pythons maximum recursion depth will kick in so at least an error will be thrown.

Conengmo:
@talbertc-usgs do you want to take a final look at this? I would like to go ahead with merging this :)

Conengmo:
Hope you're not in the shutdown @talbertc-usgs? Anyway, thanks for making this PR! Hope you liked contributing to folium.

PR #974, opened 2018-10-05
Test py37

ocefpaf:
Let's see how this goes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: python
+language: minimal
 
 sudo: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,21 +9,19 @@ env:
 matrix:
   fast_finish: true
   include:
-  - python: 2.7
-    env: TEST_TARGET=default
-  - python: 3.7
-    env: TEST_TARGET=default
-  - python: 3.7
-    env: TEST_TARGET=notebooks
-  - python: 3.7
-    env: TEST_TARGET=latest_branca
-  - python: 3.7
-    env: TEST_TARGET=docs
+  - name: default-python2
+    env: TEST_TARGET=default PY=2.7
+  - name: default
+    env: TEST_TARGET=default PY=3.7
+  - name: notebooks
+    env: TEST_TARGET=notebooks PY=3.7
+  - name: latest_branca
+    env: TEST_TARGET=latest_branca PY=3.7
+  - name: docs
+    env: TEST_TARGET=docs PY=3.7
   allow_failures:
-  - python: 2.7
-    env: TEST_TARGET=latest_branca
-  - python: 3.7
-    env: TEST_TARGET=latest_branca
+  - name: latest_branca
+    env: TEST_TARGET=latest_branca PY=3.7
 
 before_install:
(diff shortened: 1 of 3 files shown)


Conengmo:
Looks good, I like how you made things a bit clearer.

Two questions/comments:

- if branca latest is allowed to fail, why do we test it? Maybe not allow it to fail?
- what is the lowest python 3 version we support, 3.5? Then maybe we should also test default for that version.

ocefpaf:
> if branca latest is allowed to fail, why do we test it? Maybe not allow it to fail?

It is something to guide us to fix upstream but without blocking the development here with the stable version.

> what is the lowest python 3 version we support, 3.5? Then maybe we should also test default for that version. Or when we drop 2.7 we start testing for 3.5, the next lowest supported version.

Good question. Even though it is untested I would guess `3.4`. However, I'm OK just specifying 3.6 as the lower version. The reason we don't test is to avoid straining the CIs and there are virtually no 3.5/3.6 backward incompatible features in `folium`. (While I agree that extensive testing is better I prefer to keep the CI configuration fast and slim.)

TL;DR let's just update our metadata to say `folium is 3.6` (and we can start using f-strings!)

Conengmo:
> It is something to guide us to fix upstream but without blocking the development here with the stable version.

Alright. That test is failing now btw because of the branca version number parsing. I'll make a fix.

Here is the official Python support schedule:
https://devguide.python.org/#status-of-python-branches

They are not deprecating 3.5 for two more years. 3.4 is supported for half a year. I'm a bit hesitant if we can start adopting backwards incompatible features such as f-strings. I don't know that much about our user base. Which version are they using? Can the upgrade to 3.6?

I know in my work I have some machines in the field on Ubuntu LTS which are running Python 3.5. But I guess that's not the kind of machine people run folium on. I imagine it's mainly used by data types that can configure their own Python version. But that's just conjecture, truth is I don't know.

@ocefpaf do you know anything about folium users?

ocefpaf:
> Alright. That test is failing now btw because of the branca version number parsing. I'll make a fix. (Edit: pushed a commit to your branch)

Thanks!

> Here is the official Python support schedule:
> [devguide.python.org/#status-of-python-branches](https://devguide.python.org/#status-of-python-branches)
> 
> They are not deprecating 3.5 for two more years. 3.4 is supported for half a year. I'm a bit hesitant if we can start adopting backwards incompatible features such as f-strings. I don't know that much about our user base. Which version are they using? Can the upgrade to 3.6?
> 
> I know in my work I have some machines in the field on Ubuntu LTS which are running Python 3.5. But I guess that's not the kind of machine people run folium on. I imagine it's mainly used by data types that can configure their own Python version. But that's just conjecture, truth is I don't know.
> 
> @ocefpaf do you know anything about folium users?

Maybe Python 3.6 is forcing it :wink:

Conengmo:
> Maybe Python 3.6 is forcing it

Haha, "but f-strings!" :p 

I guess apart from that it's easy to maintain backwards compatibility with Python 3.4. Also, Pandas and Numpy are dropping active Python 2.7 support this January so we can do that as well if we want.

Conengmo:
So, merge this one?

PR #979, opened 2018-10-11
Remove unused import tempfile

jameslamb:
Hello,

Thank you for this excellent package! Will you please consider this PR to remove an unused import in the `folium` module? As far as I can tell it is unused.


ocefpaf:
Thanks!

PR #980, opened 2018-10-11
update leaflet search plugin

stefanocudini:


ocefpaf:
> Could we refactor this plugin to reference our main GeoJson/TopoJson class object to combine functionality instead of creating a separate one? @Conengmo, @ocefpaf?

That is a good idea but I'm in favor of merging this one as-is for now to get the update out.

PR #988, opened 2018-10-14
GeoJsonTooltip warn for GeometryCollection

### LINKED ISSUE #929: GeoJsonTooltip fails to show pop-up on non-connected areas (e.g. city with islands)
Hi there!
First of all, thank you so much for the excellent work you have done so far with folium!

I played with the folium GeoJson and noticed a problem with the GeoJsonTooltip -function, namely it fails to show a pop-up when moving mouse over an area that is not connected, that is consisting of two (or more) different areas (Polygons). You can see the unwanted behavior in the screen cap I attached with cities having island. The names of the cities consisting of only one connected area (only one Polygon) work as they should. I guess that when generating the pop up the function tries to look for a polygon but fails when it finds a list of them instead.

ezgif-2-599d86b757

jtbaker:
Closes #929. Checks geometry types for `GeometryCollection`s, stores the feature IDs of any in a list, and warns the user of incompatibility, and suggests refactoring to `MultiPolygon` geometry type.


folium/features.py
```diff
@@ -708,6 +709,13 @@ def render(self, **kwargs):
         """Renders the HTML representation of the element."""
         if isinstance(self._parent, GeoJson):
             keys = tuple(self._parent.data['features'][0]['properties'].keys())
+            geom_collections = [feature['id'] for feature in self._parent.data['features'] if feature['geometry'][
```
Conengmo: To make it break up more nicely, you could format it like this:
```
            geom_collections = [
                feature['id'] for feature in self._parent.data['features']
                if feature['geometry']['type'] == 'GeometryCollection'
            ]
```

folium/features.py
```diff
@@ -708,6 +709,13 @@ def render(self, **kwargs):
...
+                'type'] == 'GeometryCollection']
+            if any(geom_collections):
+                warnings.warn(" ".join("""GeoJsonTooltip is not configured to render tooltips for GeoJson
```
Conengmo: This splitting and joining is quite clever. But I'd prefer to keep it simple and just split the string over multiple lines. So we don't have to figure out what's going on with the joins and splits.
```
warnings.warn('this is a '
    '{} long string'
    .format(3.14))
```

Or 

```
warnings.warn(
    'this is a {} '
    'long string'
    .format(3.14)
)
```

jtbaker: That's what I had done initially, but the line breaks in the string were introducing some strange spacing from the tabs and newline characters - so I had gone this route to try and clean up the output.

folium/features.py
jtbaker: This is smart. Much more readable way to write a more complex list comprehension. Thanks Frank!

folium/features.py
Conengmo: I don't really understand what you are saying. The example I showed shouldn't lead to newline characters.

jtbaker: I can't seem to replicate the behavior on the machine I'm on right now but when I was working on it at home last night, having multiple concatenated strings read into the warning prompt was introducing some strange character spacing. I thought I had tried both approaches, having multiple strings which were concatenated, and a multiline-string with triple-quotes, but it may have just been the latter. I'll refactor to this as your suggestion doesn't exhibit the same issues. Thanks!

jtbaker:
>Also a suggestion: what do you think about putting this in a separate method? I like to split of distinct functionality in separate methods when it makes sense. For example this could become def warn_for_geometry_collection() or something, with its own docstring explaining why we're doing this.

Ok, that sounds fine to me. To clarify, do you envision it as a validation function that lives in `utilities.py` to pass `GeoJsonTooltip`'s `self._parent.data` through in the `render` method? Or as a class method only?

Conengmo:
I was thinking simply a class method. The code is only for GeoJsonTooltip, so moving it to some common place such as utilities is not needed.

folium/features.py
```diff
@@ -708,6 +709,15 @@ def render(self, **kwargs):
...
             keys = tuple(self._parent.data['features'][0]['properties'].keys())
+            geom_collections = [
+                feature['id'] for feature in self._parent.data['features']
+                if feature['geometry']['type'] == 'GeometryCollection'
+            ]
+            if any(geom_collections):
+                warnings.warn(
+                    "GeoJsonTooltip is not configured to render tooltips for GeoJson GeometryCollection geometries. "
+                    "Please consider reworking feature IDs {} to MultiPolygon for full functionality.\n"
+                    "https://tools.ietf.org/html/rfc7946#page-9""".format(geom_collections), UserWarning)
```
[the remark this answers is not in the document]
jtbaker: Whoops! Fixed.

Conengmo:
Looks good @jtbaker! This is what I had in mind. Unfortunately the test is not working yet in the Travis build. Here is the error:

```
> assert isinstance(w[-1].category, UserWarning), "GeoJsonTooltip GeometryCollection test failed."
E AssertionError: GeoJsonTooltip GeometryCollection test failed.
E assert False
E +  where False = isinstance(<class 'UserWarning'>, UserWarning)
E +    where <class 'UserWarning'> = <warnings.WarningMessage object at 0x7f328400d7f0>.category
```

It seems the `isinstance` test is not working as intended. Looking at the Pytest docs they give two methods for checking warnings:

The first is to use a `with pytest.warns(UserWarning):` context for the call that's expected to generate a warning.
https://docs.pytest.org/en/latest/warnings.html#warns

The second is to check recorded warnings, like you did. But in their example the don't use `isinstance` but `issubclass`: `assert issubclass(w.category, UserWarning)`
https://docs.pytest.org/en/latest/warnings.html#recwarn

Can you try any of these two methods to get the test working?

Conengmo:
Merged! Thanks Jason. Merging upstream master into your branch went well I see.

PR #990, opened 2018-10-15
Use folium of the same folder instead of installed version when generating the docs

### LINKED ISSUE #983: Creating documentation latex files fails (Python 3.5)

```shell
cd ./docs
make latex
```

#### Problem description

Instead of creating the tex files the following happens:

```shell
sphinx-build -b latex -d _build/doctrees   . _build/latex
Running Sphinx v1.8.1

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/XXXXXXXXX/.local/lib/python3.5/site-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/XXXXXXXXX/.local/lib/python3.5/site-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "/YYYYYYYYY/folium/folium/docs/conf.py", line 52, in <module>
    from folium._version import get_versions
ImportError: No module named 'folium'

Makefile:99: recipe for target 'latex' failed
make: *** [latex] Error 2
```

#### Expected Output

Nice latex files which I could use for creating a PDF.

     0.6.0+30.g9f648a2.dirty'

1kastner:
Closes #983.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,7 +11,9 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os, os.path
+import sys
+import os
+import os.path
 
 # Use the currently checked out folium of this folder
 sys.path.insert(0, os.path.join(os.pardir))
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,10 +13,9 @@
 
 import sys
 import os
-import os.path
 
 # Use the currently checked out folium of this folder
-sys.path.insert(0, os.path.join(os.pardir))
+sys.path.insert(0, os.pardir)
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,7 +15,7 @@ import sys
 import os
 
 # Use the currently checked out folium of this folder
-sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.abspath(os.pardir))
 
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -11,7 +11,12 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os
+import sys
+import os
+import os.path
```
Conengmo: Importing `os.path` is not needed

docs/conf.py
```diff
@@ -11,7 +11,12 @@
...
+
+# Use the currently checked out folium of this folder
+sys.path.insert(0, os.path.join(os.pardir))
```
Conengmo: `os.path.join()` is doing nothing here. Did you mean `os.path.abspath()`?

1kastner: No, I was just sleeping. It can be removed.

Conengmo:
Do we need to make the path explicit with `os.path.abspath` or is the current relative path sufficient?

1kastner:
@Conengmo that's indeed a good question. Maybe we should try it out on a fresh machine without folium installed. Then for the relative path it should fail to import in case it is necessary.

Conengmo:
If you're not sure either I would suggest to put it in. I've had unexpected issues before with relative paths in CI environments. But if you want to test it out that's fine as well.

1kastner:
Let's be on the safe side

Conengmo:
It seems autobuilding the docs went fine. Thanks @1kastner!

PR #1005, opened 2018-10-28
Fixed choropleth

### LINKED ISSUE #905: Threshold scale - limits at five groups /  six values?
folium 0.5.0

By observation of default choropleth maps, the legend seems to be split into six groups (so a list of seven boundary values). 

But if a custom threshold scale is provided, then this seems to limit at five groups (the max length of the boundary value list is six items)?

```#in folium.py
if threshold_scale and len(threshold_scale) > 6
```

Why the difference?

More concerningly (by observation; I'm working with sensitive data and don't have a test example to hand at the moment) I notice that whilst the limits on the legend are updated to values provided via the `threshold_scale` parameter, the colours depicted on the map correspond to default value ranges, not the supplied `threshold_scale` values?

In the following:
```
mymap = folium.Map(zoom_start=10, max_zoom=15)

mymap.choropleth(geo_data='data/geodata.geojson',
                       data=df, 
                       columns=['code', 'data'], 
                       key_on='feature.properties.Name',
                       fill_color='RdYlBu', fill_opacity=0.4, line_opacity=0.9,
                       legend_name='*******', 
                       threshold_scale=[50, 60, 70, 80, 90, 100])
```

image

it doesn't matter what I set the `threshold_scale` to, the map coloring remains the same as used with the default scale, although the legend does change to use the colour boundaries I have set?

UPDATE: it seems that the red coloured group lower band coloring in the above map was being used to colour `None` value items in the data; in the default scale, no colouring is applied to those groups. It would be useful to be able to specify a 'novalue` colour, perhaps set to `None` by default (no colouring) or a specified colour if provided.

----

Another issue with `threshold_scale` appears to be a fencepost error in that if the upper bound of the `threshold_scale` list is less than the upper bound data value, an error occurs (as below), although no error is raised at the other end where the lower bound of the `threshold_scale` may be set above the lower bound value of the data.

```
IndexError                                Traceback (most recent call last)
/usr/local/lib/python3.5/dist-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

/usr/local/lib/python3.5/dist-packages/folium/map.py in _repr_html_(self, **kwargs)
    249             self._parent = None
    250         else:
--> 251             out = self._parent._repr_html_(**kwargs)
    252         return out
    253 

/usr/local/lib/python3.5/dist-packages/branca/element.py in _repr_html_(self, **kwargs)
    317 
    318         """
--> 319         html = self.render(**kwargs)
    320         html = "data:text/html;charset=utf-8;base64," + base64.b64encode(html.encode('utf8')).decode('utf8')  # noqa
    321 

/usr/local/lib/python3.5/dist-packages/branca/element.py in render(self, **kwargs)
    307         """Renders the HTML representation of the element."""
    308         for name, child in self._children.items():
--> 309             child.render(**kwargs)
    310         return self._template.render(this=self, kwargs=kwargs)
    311 

/usr/local/lib/python3.5/dist-packages/folium/map.py in render(self, **kwargs)
    336             '</style>'), name='map_style')
    337 
--> 338         super(LegacyMap, self).render(**kwargs)
    339 
    340 

/usr/local/lib/python3.5/dist-packages/branca/element.py in render(self, **kwargs)
    622 
    623         for name, element in self._children.items():
--> 624             element.render(**kwargs)

/usr/local/lib/python3.5/dist-packages/branca/element.py in render(self, **kwargs)
    618         script = self._template.module.__dict__.get('script', None)
    619         if script is not None:
--> 620             figure.script.add_child(Element(script(self, kwargs)),
    621                                     name=self.get_name())
    622 

/usr/local/lib/python3.5/dist-packages/jinja2/runtime.py in __call__(self, *args, **kwargs)
    573                             (self.name, len(self.arguments)))
    574 
--> 575         return self._invoke(arguments, autoescape)
    576 
    577     def _invoke(self, arguments, autoescape):

/usr/local/lib/python3.5/dist-packages/jinja2/runtime.py in _invoke(self, arguments, autoescape)
    577     def _invoke(self, arguments, autoescape):
    578         """This method is being swapped out by the async implementation."""
--> 579         rv = self._func(*arguments)
    580         if autoescape:
    581             rv = Markup(rv)

<template> in macro(l_1_this, l_1_kwargs)

/usr/local/lib/python3.5/dist-packages/jinja2/runtime.py in call(_Context__self, _Context__obj, *args, **kwargs)
    260                 args = (__self.environment,) + args
    261         try:
--> 262             return __obj(*args, **kwargs)
    263         except StopIteration:
    264             return __self.environment.undefined('value was undefined because '

/usr/local/lib/python3.5/dist-packages/folium/features.py in style_data(self)
    563 
    564         for feature in self.data['features']:
--> 565             feature.setdefault('properties', {}).setdefault('style', {}).update(self.style_function(feature))  # noqa
    566             feature.setdefault('properties', {}).setdefault('highlight', {}).update(self.highlight_function(feature))  # noqa
    567         return json.dumps(self.data, sort_keys=True)

/usr/local/lib/python3.5/dist-packages/folium/folium.py in style_function(x)
    303                 'color': line_color,
    304                 'fillOpacity': fill_opacity,
--> 305                 'fillColor': color_scale_fun(x)
    306             }
    307 

/usr/local/lib/python3.5/dist-packages/folium/folium.py in color_scale_fun(x)
    290             def color_scale_fun(x):
    291                 return color_range[len(
--> 292                     [u for u in color_domain if
    293                      get_by_key(x, key_on) in color_data and
    294                      u <= color_data[get_by_key(x, key_on)]])]

IndexError: list index out of range
```

### LINKED ISSUE #971: Unexpected widening of range in choropleth
#### Code

```python
import pandas as pd
import os
import folium

state_geo = os.path.join('', 'countries.geo.json')

state_unemployment = os.path.join('', 'new.csv')
state_data = pd.read_csv(state_unemployment)

m = folium.Map(location=[48, -102], zoom_start=3)

m.choropleth(
    geo_data=state_geo,
    name='choropleth',
    data=state_data,
    columns=['State', 'Unemployment'],
    key_on='feature.id',
    fill_color='YlGn',
    fill_opacity=0.7,
    line_opacity=0.2,
    legend_name='Unemployment Rate (%)'
)


folium.LayerControl().add_to(m)

m.save('output.html')

```
#### Problem description

Why does Folium show me that I have negative value records? the lowest value record is 0, and the highest value is 500, however the map caption looks like this:
sem titulo

### LINKED ISSUE #996: Better handling of missing values in choropleth
_Might be related to #498_
_Similar issue description on [Stackoverflow](https://stackoverflow.com/questions/51961848/set-color-for-missing-values-in-folium-choropleth)._

Below is a slightly modified version of a cell inside the Quickstart notebook to demonstrate the issue.
```python
import pandas as pd

state_geo = os.path.join('data', 'us-states.json')

state_unemployment = os.path.join('data', 'US_Unemployment_Oct2012.csv')
state_data = pd.read_csv(state_unemployment)
# CHANGE BELOW
state_data = state_data.drop(0)
# END OF CHANGE
m = folium.Map(location=[48, -102], zoom_start=3)

m.choropleth(
    geo_data=state_geo,
    name='choropleth',
    data=state_data,
    columns=['State', 'Unemployment'],
    key_on='feature.id',
    fill_color='BuPu',
    fill_opacity=0.7,
    line_opacity=0.2,
    legend_name='Unemployment Rate (%)'
)


folium.LayerControl().add_to(m)

m
```

screenshot

#### Problem description
As shown above, the Alabama is associated with the _strongest_ color which might not be what we want as the corresponding value is missing.

#### Expected Output
I would suggest a dedicated color parameter to handle this issue:

I would suggest adding a parameter to the function `choropleth` visible [here](https://github.com/python-visualization/folium/blob/584ea34a3f29c84513c5a53115926c58de2479dc/folium/folium.py#L423), maybe `nan_fill='#646464'`.

And modifiying the color_scale_fun function from:
```python
            def color_scale_fun(x):
                idx = len(
                    [
                        u for u in color_domain if
                        get_by_key(x, key_on) in color_data and
                        u <= color_data[get_by_key(x, key_on)]
                    ]
                )
                return color_range[idx-1]
```
to (with explanations):
```python
            def color_scale_fun(x):
                idx = len(
                    [
                        u for u in color_domain if
                        get_by_key(x, key_on) in color_data and
                        u <= color_data[get_by_key(x, key_on)]
                    ]
                )
                # CHANGE HERE
                # idx == 0 when there is no match in the data
                # so idx-1 of color_range would return the last value
                # of color_range, we need to change this
                if idx == 0:
                    return nan_fill
                # END OF CHANGE
                return color_range[idx-1]
```

More investigation is needed for when a value is not missing but is `np.NaN` where the same issue seems to be present (that would need to be checked) and the same modification could be applied. 

'0.6.0'

_PS: sorry for the closing/opening of the issue, I inadvertently created an empty issue first..._

FloChehab:
Closes #971, closes #996 and closes #905.

Other changes:

- I removed the chunk below, as I think this check adds an unwanted constraint (if too many colors are _asked_ then it will fail in `branca.utilities.color_brewer`):

```python
        if threshold_scale is not None and len(threshold_scale) > 6:	
            raise ValueError('The length of threshold_scale is {}, but it may '	
                             'not be longer than 6.'.format(len(threshold_scale)))  # noqa
```

- I have updated the docstring.
- I have made few changes linked to the fact that every time `color_brewer` was used, one extra color was generated.

One thing I am not sure: I have added the `nan_fill_color` parameter at a _logical_ place in the list of parameters but we might want to change that.

All current tests should pass. But new tests might be welcomed for the new _features_.


folium/folium.py
```diff
@@ -550,32 +553,39 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
             key_on = key_on[8:] if key_on.startswith('feature.') else key_on
-            color_range = color_brewer(fill_color, n=len(color_domain))
+            color_range = color_brewer(fill_color, n=nb_bins)
 
             def get_by_key(obj, key):
                 return (obj.get(key, None) if len(key.split('.')) <= 1 else
                         get_by_key(obj.get(key.split('.')[0], None),
                                    '.'.join(key.split('.')[1:])))
 
             def color_scale_fun(x):
```
FloChehab: Here, I decided to mimic the previous behavior described in the comment below in order not to introduce any braking change:

```python
# we consider that values outside the color 
# should be affected to the first (or last) color bin.
```

However, doing so is actually producing misleading maps (for example, if we force the color scale do go from 1 to 10 with step 1 and we have -60 as a value, it would map to the same color as the bin 1 to 2 without any warning).
It might be a good idea to raise a warning in such circumstances or just to forbid this behavior (that's what I would do).
I didn't choose the last option as it was causing unit tests to fails:

Some notebooks use `branca` `split_six` functions which can produce *weird* intervals, for example:
```python
import pandas as pd
from folium import branca
threshold_scale = branca.utilities.split_six(series=pd.Series([1,2,2,3,3,3,3,3,3,4]))
# produces (as expected): [1.0, 3.0, 3.0, 3.0, 3.0]
```
Therefore, the `4` we would like to map to a color would be outside of the `threshold_scale`.

folium/folium.py
```diff
@@ -466,14 +469,17 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
             start with 'feature' and be in JavaScript objection notation.
             Ex: 'feature.id' or 'feature.properties.statename'.
         threshold_scale: list, default None
```
Conengmo: Not your change, but I would like that all arguments that have a default value of `None` because they are optional to have a docstring like `name: type, optional` instead of `name: type, default None`. In this case the default is actually a list with 7 values.

folium/folium.py
Conengmo: Good that you point this out. We have to make a decision here. I'm listing the options:

1. Let values outside of the range use the min or max value (current behavior).
  * This makes sense if you have a few outliers you don't want to stretch out your range for.
  * No time to filter out outliers, get stuff done quickly.
2. Raise a UserWarning when values fall outside of the range.
  * Sort of compromise between the current behavior and raising an error.
  * Numpy does this a lot, which results in people figuring out how to disable/ignore them.
  * Kind of annoying (subjective, I know :p )
3. Raise an exception when values fall outside of the range.
  * Having a value lie outside of the range is a mistake and we don't do mistakes.
  * Force the user to give either a good range or remove outliers.
4. Give the values outside the range the same color as `nan_fill_color`.
  * This does show they are not in the range, but it makes them indistinguishable from nan values.
5. Give the values outside the range a special color and add a new argument for it.
  * Makes it easy to see which values fall outside the range.
  * Adds complexity, another argument to check out and set.

I'm currently leaning towards raising an exception. Many users may not pass a custom range, so they will not encounter this issue. If a user passes a custom range, that means he or she is advanced enough to also be able to properly preprocess the data and deal with outliers.

On the other hand I like how folium is hackable. If someone wants to deal with their outliers by cheating a bit, who are we to judge. But then maybe a UserWarning would be appropriate.

So after writing this comment I think I would lean towards leaving it as is and add a UserWarning. Do you want to make the case for another option? I'd like to hear your opinion. We can change the tests based on what we decide.

folium/folium.py
```diff
@@ -550,32 +553,39 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
-                idx = len(
-                    [
-                        u for u in color_domain if
-                        get_by_key(x, key_on) in color_data and
-                        u <= color_data[get_by_key(x, key_on)]
-                    ]
-                )
-                return color_range[idx-1]
+                key_of_x = get_by_key(x, key_on)
+                if key_of_x in color_data.keys():
```
Conengmo: I was thinking if this could be made a bit simpler. Instead of two if-else statements, make two if-statements that return on something wrong and keep the happy path unindented. For example here return immediately if the value is missing, no comment needed.
```suggestion
                if key_of_x not in color_data.keys():
                    return nan_fill_color
```

Then set the value, do a if-statement whether to return on `np.isnan`, and do the rest.

This is fine as it is, so no problem if you leave it like this. I like to discuss the finishing touch, that's all.

folium/folium.py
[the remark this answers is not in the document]
FloChehab: I was suggesting `bins` following my experience with Pandas [`hist` function](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.hist.html) (similar to Matplotlib).
We could have it set default to `6` and make it accept a number (of bins) or a list of numbers considered as the edges of the bins (just like Pandas).

So if I understand you correctly, we could add a `bins` parameter, and if it is not set would default to `threshold_scale`?

I'll wait for your "go/no go" on this part. (I've noted the need to remove `Data range for D3 threshold scale.`)

FloChehab: I didn't get what you are saying here, could you clarify a bit? Thanks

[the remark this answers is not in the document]
Conengmo: Nice idea to make it behave as in Pandas hist. It's the same as the Numpy hist bins argument isn't it?
https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html

The way to go here IMO is to add `bins` as an argument, removing `threshold_scale` while keeping backwards compatibility. So `threshold_scale` shouldn't be listed in the documentation but still work. One way to do this is to accept `**kwargs` and retrieve `threshold_scale` from it. For example:

```
def choropleth(bins=6, **kwargs):
    if 'threshold_scale' in kwargs:
        bins = threshold_scale
        raise DeprecationWarning('maybe?')
```

Is this what you had in mind?

folium/folium.py
FloChehab: I'll apply the change in the next commit.

folium/folium.py
Conengmo: In many cases the default value is None, but from the users perspective that's not really helpful since None disables the argument. In that case it's better to say the argument is optional.

But here that's not really the case since the default value should actually be a list of 7 values so I guess my point doesn't really work here.

This doesn't matter here so I'll set the conversation as resolved :)

folium/folium.py
Conengmo: Let's raise a `ValueError` then, with a useful message. We'll learn our users some proper preprocessing ;) Better to be explicit than have it fail silently.

Conengmo:
By the way, have you considered working with opacity instead of coloring for missing/nan/out-of-range values?

folium/folium.py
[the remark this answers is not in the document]
FloChehab: > Is this what you had in mind?
Seems better than what I had in mind! (I am not a Python master yet :smile: )

> It's the same as the Numpy hist bins argument isn't it?
I think it is the same spirit.

folium/folium.py
FloChehab: Should be good now

Conengmo: Looks good indeed, thanks.

folium/folium.py
```diff
@@ -614,9 +624,10 @@ def highlight_function(x):
...
         # Create ColorMap.
         if color_domain:
-            brewed = color_brewer(fill_color, n=len(color_domain))
+            nb_bins = len(color_domain) - 1
+            brewed = color_brewer(fill_color, n=nb_bins)
             color_scale = StepColormap(
-                brewed[1:len(color_domain)],
+                brewed,
                 index=color_domain,
                 vmin=color_domain[0],
```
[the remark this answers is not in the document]
Conengmo: Are you sure? It seems they are passed to the init of the parent class, `ColorMap`. That one uses them in its `render` function.

folium/folium.py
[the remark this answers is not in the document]
Conengmo: Oops example should be:
`bins = kwargs['threshold_scale']`

FloChehab:
> By the way, have you considered working with opacity instead of coloring for missing/nan/out-of-range values?

I think those values should standout that's why I went for `'black'`, but if someone wants to play with opacity it should be possible with HEX strings `nan_fill_color='#ffffff00'` (it works, I just tested). We could mention it in the docstring.

folium/folium.py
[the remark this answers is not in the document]
FloChehab: Indeed, I was wrong, I reverted that part.

FloChehab:
I have updated the code accordingly to our discussions, I'll wait for your feedback.

Obviously, CI will fail because of the `DeprecationWarning` and the new check to see if all values fall into the domain defined by bins. So next we will have on how to update the current tests. Then, what new tests to add. And finally, as you said, give new examples.

folium/folium.py
```diff
@@ -508,24 +522,28 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
             raise ValueError('Please pass a valid color brewer code to '
                              'fill_local. See docstring for valid codes.')
 
+        if 'threshold_scale' in kwargs:
+            if kwargs['threshold_scale'] is not None:
+                bins = kwargs['threshold_scale']
+            raise DeprecationWarning('''
+                choropleth `threshold_scale` parameter is now depreciated
+                in favor of the `bins` parameter.''')
+
```
FloChehab: I am not sure that:
```python
            raise DeprecationWarning('''
                choropleth `threshold_scale` parameter is now depreciated
                in favor of the `bins` parameter.''')
```

Is the ideal way to do it as it will stop the script as an error.

I have looked online and tried:
```python
warnings.warn('''
                choropleth `threshold_scale` parameter is now depreciated
                in favor of the `bins` parameter.''', DeprecationWarning)
```

But it doesn't seem to be shown in the Notebook I am using (it works well in the unit test on the other hand).

I am sure you have a suggestion :)

FloChehab:
By the way, while we are working on choropleth, we could also allow the `fill_color` parameter to be a direct mapping of colors for each bin defined. This would enable the creation of more custom maps (eg: with two colors).

folium/folium.py
Conengmo: You're right that the warning shouldn't be raised, instead `warnings.warn(msg, category)` is the way to go. Good find.

Deprecation warnings are ignored by default. They only appear in systems that enable them. We could use a `FutureWarning`, but I think I'm more inclined to use a `DeprecationWarning` and then once we later really remove the feature raise an error.
https://docs.python.org/3/library/warnings.html#warning-categories

folium/folium.py
```diff
@@ -465,15 +467,27 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
-            of quantiles: [0, 0.5, 0.75, 0.85, 0.9], rounded to the nearest
-            order-of-magnitude integer. Ex: 270 rounds to 200, 5600 to 6000.
+        bins: int or sequence of scalars or str, optional
+            If `bins` is an int, it defines the number of equal-width
+            bins between the min and the max of the values (6, by default).
+            If `bins` is a sequence, it defines the bin edges,
+            including the rightmost edge, allowing for non-uniform bin widths.
+            For more information on this parameter, please have a look at
+            numpy.histogram function.
+            Warning: if any (non-nan) values is outside of the inclusive
```
Conengmo: IMO you don't really have to state that here, they will see the error when it happens. Makes the docstring a bit less lenghty also.

folium/folium.py
```diff
@@ -538,55 +556,73 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
-                            if data_max < 0 else 1)
-            data_min, data_max = (1.01*data_min-0.01*data_max,
-                                  1.01*data_max-0.01*data_min)
-            nb_class = 6
-            color_domain = [data_min+i*(data_max-data_min)*1./nb_class
-                            for i in range(1+nb_class)]
-        else:
-            color_domain = None
+        if color_data is not None and key_on is not None:
+            # numpy histogram_bin_edges is only available in the most
```
Conengmo: No need for this comment

folium/folium.py
```diff
@@ -538,55 +556,73 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
+            # recent numpy version, so we will use numpy histogram
+            real_values = np.array(list(color_data.values()))
+            real_values = real_values[~np.isnan(real_values)]
+            _, bins = np.histogram(real_values, bins=bins)
```
Conengmo: I'm not a fan of reusing variable names when their type may have changed. Can we use a different name? In Numpy they name the output `bin_edges`.

folium/folium.py
```diff
@@ -538,55 +556,73 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
+
+            bins_min, bins_max = min(bins), max(bins)
+            if np.any((real_values < bins_min) | (real_values > bins_max)):
+                raise ValueError('''
```
Conengmo: Because of the triple quotes the error includes new lines. Better to break up the string over the lines something like this:

```
raise ValueError(
    'All values are expected to fall into one of the provided '
    'bins (or to be Nan). Please check the `bins` parameter '
    'and/or your data.'
)
```

folium/folium.py
```diff
@@ -538,55 +556,73 @@ def choropleth(self, geo_data, data=None, columns=None, key_on=None,
...
-                        get_by_key(x, key_on) in color_data and
-                        u <= color_data[get_by_key(x, key_on)]
-                    ]
-                )
-                return color_range[idx-1]
+                key_of_x = get_by_key(x, key_on)
+
+                if key_of_x not in color_data.keys():
+                    # 1: opacity is handled within nan_fill_color
+                    return nan_fill_color, 1
```
Conengmo: Having `nan_fill_color` as RGBA values adds some complexity. What if someone uses a regular RGB value and expects the `fill_opacity` to work on these fields as well? Do we have to check if `nan_fill_color` is a RGBA value and only return `opacity = 1` in that case?

I'm skeptical of adding more complexity by parsing color strings. Either return `opacity 1` or go back to the RGB situation we had before. What do you think?

Conengmo:
> we could also allow the fill_color parameter to be a direct mapping of colors for each bin defined

That's not a bad idea. Is that something you would want to do yourself? In any case, I suggest to do that in a new PR after this one, to make this one manageable.

folium/folium.py
Conengmo: Or maybe instead add a `nan_fill_opacity` argument? That would be more in line with the rest of the code. We don't use RGBA in any other place I think.

FloChehab: I think, the easiest way to go would be to add a `nan_fill_opacity` parameter (that would default to the `fill_opacity` value if it's not set) to the function because I am sure at some point in the future someone is going to ask for such possibility.
And going that way we won't have to think about tricky HEX RGBA strings.

What do you think ?

folium/folium.py
FloChehab: I have updated the code, we should be good

folium/folium.py
FloChehab: Better ?

folium/folium.py
FloChehab: I went for `bin_edges`

folium/folium.py
FloChehab: I have made the change. I wasn't aware of this way to split strings in python, thanks !

folium/folium.py
FloChehab: Oh we thought at the same thing... I'll add the parameter.

Conengmo: I agree, that would be best. 

I do like the idea of using RGBA though, so thanks for introducing that. But we would have to handle that more consistently then. Maybe for in another PR.

FloChehab: I have updated the code to have a `nan_fill_opacity` parameter.

Conengmo:
Neat! Nice work. I think this is a well rounded PR, with fixing the binning and missing values.

So, what's next? Updating the examples I would say. Only the notebook tests failed, probably because we deprecated `threshold_scale`.

I know of two notebooks that have choropleth examples, but there may be others.

https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/GeoJSON_and_choropleth.ipynb

https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/Quickstart.ipynb

Another thing would be to look at the tests and see if we can change/add something for the new and updated functionality.

FloChehab:
I have updated the notebooks and added one test (I did my best to come up with something straightforward for testing the new functionalities, it might not be up to your standards).

Also, I am wondering why are the notebooks stored on Github with runned cells? (I feel that it is quite a waste of space, and they are not all up-to-date)

Conengmo:
Link to nbviewer for your notebook changes:
http://nbviewer.jupyter.org/github/FloChehab/folium/blob/fix_choropleth/examples/Quickstart.ipynb
http://nbviewer.jupyter.org/github/FloChehab/folium/blob/fix_choropleth/examples/GeoJSON_and_choropleth.ipynb

> I have updated the notebooks and added one test

Good! Notebooks look good and your test seems decent. Thanks. I think we're done with this PR then! What do you think, good to go?

> I am wondering why are the notebooks stored on Github with runned cells

Good question. It's so they are immediately viewable with something like nbviewer.jupyter.org. I don't really like the diffs being unreadable. We do let Travis test if they are renderable though.

Do you have an idea how the way we deal with the notebooks could be improved?

ocefpaf:
> I would consider the PR as done on my side too.
> 
> I hadn't taken a look at `nbviewer` before and therefore, I get why notebooks are stored with run cells in the repo.
> 
> I am more used to Gitlab CI, and in such context I would convert the notebooks to html in the pipeline and store them as job artefacts (which can be directly accessible using a simple URL). I don't know if this could be possible on Github + Travis.

It is and I used to do that but nbviewer is quite handy, it reduces the maintenance  steps and we don't waste CIs running the notebooks.

ocefpaf:
BTW, locally I use `nbdime` to check notebooks diffs on git. It is quite handy to review notebooks commits.

Conengmo:
Merged! Thanks @FloChehab, I've enjoyed working with you.

FloChehab:
@Conengmo you are welcome ! It was a pleasure working with you too (you took quite some time discussing, commenting and giving advice !)

It might be my only contribution to Folium: we had to use Folium for a course at university when we faced the now fixed bugs related to `choropleth`; I came up with some nasty fix and wanted to have a more durable fix for the whole community :smile: Anyway, it was very nice to collaborate !

ReidWeb:
Has a version with this been published yet?

ocefpaf:
> Has a version with this changeset been published yet? Or is there a way to easily install the pre-release version? Would quite like to utilise the updated version for something.

Not yet. Planning on a new release for next week.

PR #1009, opened 2018-10-29
Add a first guess of timestamped_ImageOverlay as a plugin for Folium.

### LINKED ISSUE #1007: Support TimeDimension plugin for ImageOverlay
It would be great to use the timeDimension plugin also for ImageOverlays. 

Best regards

meteoDaniel:
Closes #1007.

I barely mixed the `raster_layers.ImageOverlay` class with the existing `Plugins.timestamped_wmsTilelAyers-py`. Actually this first guess is not working but i am sure with your help we could add this plugin after debugging. 

As i mentioned in the issue: I am not very familiar with `template`and `javascript`.  I am mainly develop my tools in python.

My Idea to call the plugin as followed:
```
folium.plugins.TimestampedImageOverlay([url1, url2, url3, url4],
                        time_interval='2018-10-28/2018-10-28',
                        valid_time_range='14:55/15:20',
                        overlay=True, name='Test Series of Images',
                        opacity=0.6,
                 bounds=[[lat.min,
                          lon_min],
                         [lat_max,
                          lon:max]]).add_to(base_map)

```


meteoDaniel:
I now understand the example [http://apps.socib.es/Leaflet.TimeDimension/examples/example10.html](url) quite a little bit more. 

The example and my implementation is hard coded and my version still does not work. They implemented a new class for TimeDimension.ImageOverlay. I am not familiar with the timeDimension class. Does the class need a UrlFunction with a timestamp or is it possible to loop over a list of url and a list of timestamp generated by the configurated timerange?

For WMSTileLayers I can't see any timestamps.

Conengmo:
Based on our [plugin acceptance criteria](https://github.com/python-visualization/folium/blob/main/.github/CONTRIBUTING.md#plugin-acceptence-criteria), I don't think we should merge this into folium. The other timestamped plugins have caused some headaches in the past with the amount of custom code and we don't have the capacity to support it.

My suggestion is you publish this work as a package yourself! Then add a link to your package in our documentation so users can find it.

PR #1012, opened 2018-11-06
Fix tests

Conengmo:
It seems the `branca.six` module got deprecated or broken or whatever, so the Travis build fails.

In a recent branca PR it was replaced with the regular `six` package.
https://github.com/python-visualization/branca/pull/49

So I did that here as well, see if it works.


Conengmo:
Thanks for the review @ccclauss! I'll go ahead and merge this, so the other PR's can update from master and have their tests pass.

PR #1013, opened 2018-11-06
Add Python 3.7 to the PyPI trove classifiers in setup.py

cclauss:
These must have been some glitch in the last release to PyPI because these trove classifiers should appear in the leftmost column in https://pypi.org/project/folium/ instead at the very bottom of the README.md txt.


ocefpaf:
The failure here is unrelated to the PR. Merging it. Thanks @cclauss!

Issue #891, opened 2018-06-20
Use Firefox or Chrome for Selenium instead of Phantomjs


Conengmo:
Phantomjs is being deprecated, though it works for the moment, so not a bug yet.

```
C:\Users\micha\Anaconda3\lib\site-packages\selenium\webdriver\phantomjs\webdriver.py:49: 
UserWarning: Selenium support for PhantomJS has been deprecated, please use headless versions of Chrome or Firefox instead
warnings.warn('Selenium support for PhantomJS has been deprecated, please use headless '<
```

We need to use chrome or firefox headless. Take a look at https://github.com/altair-viz/altair/pull/554/files for inspiration.

ocefpaf:
Addressed in https://github.com/python-visualization/folium/pull/1015

PR #1015, opened 2018-11-07
Use firefox headless instead of phantomjs 

ocefpaf:
Closes https://github.com/python-visualization/folium/issues/891

I chose firefox b/c it is easier to test on Travis-CI and we found some rendering issues with chrome/chromium in the past. 

PS: @Conengmo the `_png()` is not a first class citizen yet. Not sure what else we could do to make it more robust. Maybe a driver selection option?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -44,31 +44,37 @@ Whenever possible, please also include a [short, self-contained code example](ht
 
 First of all, thanks for your interest in contributing!
 
-- If you are new to git/Github, please take check a few tutorials
-  on [git](https://git-scm.com/docs/gittutorial) and [GitHub](https://guides.github.com/).
-- The basic workflow for contributing is:
-  1. [Fork](https://help.github.com/articles/fork-a-repo/) the repository
-  2. [Clone](https://help.github.com/articles/cloning-a-repository/) the repository to create a local copy on your computer:
-    ```
-    git clone git@github.com:${user}/folium.git
-    cd folium
-    ```
-  3. Create a branch for your changes
-    ```
-    git checkout -b name-of-your-branch
-    ```
-  4. Make change to your local copy of the folium repository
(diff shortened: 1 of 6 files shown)


Conengmo:
Looks good, and the tests work fine. Only problem is that the temporary file doesn't work on Windows. We had this problem before:
https://docs.python.org/3.6/library/tempfile.html#tempfile.NamedTemporaryFile

I'll see if I can find a solution, otherwise we'll merge it as is.

Conengmo:
I can get the tempfile stuff to work if instead of using `NamedTemporaryFile` we use the underlying `mkstemp` function. We have to close and delete the file ourselves, but it's not that hard.

```
@contextmanager
def _tmp_html(data):
    """Yields the path of a temporary HTML file containing data."""
    filepath = ''
    try:
        f_int, filepath = tempfile.mkstemp(suffix='.html', prefix='folium_')
        os.write(f_int, data.encode('utf8'))
        os.close(f_int)
        yield filepath
    finally:
        if os.path.isfile(filepath):
            os.remove(filepath)
```

Note that this returns the file path, not a file descriptor. What do you think @ocefpaf?

ocefpaf:
@Conengmo I'll take a look at the failures later but it seems that only Python 2.7 is affected and, b/c this feature is not really ready for prime time, I'm OK making it Python 3 only. (We already agreed that the next release is the last one for Python 2.7 anyway.)

ocefpaf:
~~@Conengmo I'll take a look at the failures later but it seems that only Python 2.7 is affected and, b/c this feature is not really ready for prime time, I'm OK making it Python 3 only. (We already agreed that the next release is the last one for Python 2.7 anyway.)~~

Spoke too soon. All the failures were related to new `pytest` version. This one is good to go now.

ocefpaf:
Merging this to b/c of the Travis-CI fixes in the last commit are needed.

Issue #522, opened 2016-10-05
Include AntPath plugin


existeundelta:
https://github.com/rubenspgcavalcante/leaflet-ant-path/issues/19#issuecomment-251670236

PR #1016, opened 2018-11-08
Antpath

ocefpaf:
Closes https://github.com/python-visualization/folium/issues/522


tests/plugins/test_antpath.py
```diff
@@ -0,0 +1,59 @@
...
+          {{this.get_name()}} = L.polyline.antPath(
+                  {{this.location}},
+                  {{ this.options }}
+                )
+                .addTo({{this._parent.get_name()}});
+        """)  # noqa
+
+    expected_rendered = tmpl.render(this=antpath)
+    rendered = antpath._template.module.script(antpath)
+    assert folium.utilities.compare_rendered(expected_rendered, rendered)
```
ocefpaf: I'm planning on a major test re-factor soon.
Probably using this approach if you all agree on it.

ocefpaf:
@Conengmo this is rebased and ready for another round of reviews.

folium/utilities.py
```diff
@@ -389,6 +389,19 @@ def iter_points(x):
...
+    """
+    Return True/False if the normalized rendered version of
+    two folium map objects are the equal or not.
+
+    """
+    return _normalize(obj1) == _normalize(obj2)
+
+
+def _normalize(rendered):
+    return [line.strip() for line in rendered.splitlines() if line.strip()]
```
Conengmo: Add a docstring
```suggestion
    """Return the input string as a list of stripped lines."""
     return [line.strip() for line in rendered.splitlines() if line.strip()]
```

folium/vector_layers.py
```diff
@@ -66,27 +66,15 @@ def path_options(**kwargs):
     http://leafletjs.com/reference-1.2.0.html#path
 
     """
-    valid_options = (
```
Conengmo: I get that it's cleaner to not have to define and maintain this list. But now if someone passes a faulty option it gets ignored silently, which is a problem because a user may not exactly know which options are valid. When no error is raised it's hard to find out. Can we add a check at the bottom of this function that warns when any kwargs have not been popped?

tests/plugins/test_antpath.py
Conengmo: I do like standardizing the comparison function. But I'm not sure this way of testing is that useful. We only test if we correctly copied the template definition here correctly. Suppose the rendered javascript is incorrect, it will be incorrect for both `rendered` and `expected_rendered`. If we really want to do a functional test, we should store the rendered html + js that we know is correct and compare the output of `rendered` with that.

I would suggest discussing a test refactor separately so we can get this PR merged.

folium/vector_layers.py
ocefpaf: I'm coming from a world where `kwargs` are a user responsibility and the valid options are well documented. The latter is not true in `folium` and not all user are like me :grimacing: So please take this with a grain of salt :wink: 

If we choose warning route we must revert the parsing of the `kwargs` in the plugins to avoid triggering warnings for valid plugin options. However, that may override the option with when `path_options` runs. (That was the case for antpath and the reason for this change.)

Maybe the best option is to create a `kwarg` check function that runs after all the options are parsed and run it in the plugin. What do you think? Not sure if that is worth it though...

PS: the reason for removing the default check was to accommodate the extra `kwargs` that are only valid for the antpath and the different defaults from polyline while still relying `path_options`.

tests/plugins/test_antpath.py
ocefpaf: Indeed. We need that kind of testing since the beginning of `folium` :grimacing: 

> I would suggest discussing a test refactor separately so we can get this PR merged.

:+1:

folium/vector_layers.py
Conengmo: Alright I see it's not that easy. We'll leave it as it is now and maybe work on this in a separate PR.

Conengmo:
Nice one @ocefpaf! Loving the hypnotizing moving lines on the map.

ocefpaf:
Thanks for the review and merge!


I'll be back from vacations next week and I'm planning a new release.
What do you think?

PR #1017, opened 2018-11-09
Fix #928 and implements from flake8 fixes that stickler missed

### LINKED ISSUE #928: HeatMap plugin isn't working
```python
import pandas as gpd
import folium

# loading the shapefile and adding it to the folium map
shp = gpd.read_file('./shapefiles/fortal118.shp')
shp.crs = "+proj=utm +zone=24 +ellps=WGS84 +datum=WGS84 +units=m +no_defs +south"
shp = shp.to_crs("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")

Map = folium.Map(location=[-3.731862, -38.526669], zoom_start=12, tiles='cartodbpositron')
folium.GeoJson(shp).add_to(Map)

# loading the data to use in order to make the heatmap
cases2011 = pd.read_csv('cases2011.dat', sep=" ", names=['day', 'month', 'lon', 'lat', 'label'], usecols=['lon', 'lat'])
cases2011['amount'] = np.random.randint(0, 100, size=len(cases2011))

# constructing the heatmap and adding it to the folium map
max_amount = float(cases2011['amount'].max())
hm_wide = HeatMap(zip(cases2011.lat.values, cases2011.lon.values, cases2011.amount.values),
                 min_opacity=0.1,
                 max_val=max_amount,
                 radius=17, blur=15,
                 max_zoom=10)
Map.add_child(hm_wide)
# showing the figure
Map
```
#### Problem description

The issue is that it was supposed to show a figure with the folium map, the shapefile and the heatmap generated. The heatmap is not displayed, just the folium map and the shapefile appears.

#### Expected Output
 
The expected output should be similar to the figure of this post tutorial https://alcidanalytics.com/p/geographic-heatmap-in-python

0.5.0

Could anyone help me with this issue, please?

ocefpaf:
Closes https://github.com/python-visualization/folium/issues/928

I also ported some functions from `branca.`utilities` that should live in `folium.utilities`.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -12,18 +12,18 @@ import warnings
 
 from branca.colormap import LinearColormap, StepColormap
 from branca.element import (Element, Figure, JavascriptLink, MacroElement)
-from branca.utilities import (
+from branca.utilities import color_brewer
+
+from folium.folium import Map
+from folium.map import (FeatureGroup, Icon, Layer, Marker, Tooltip)
+from folium.utilities import (
     _iter_tolist,
     _parse_size,
-    color_brewer,
+    get_bounds,
     image_to_url,
     none_max,
     none_min,
 )
-
-from folium.folium import Map
-from folium.map import (FeatureGroup, Icon, Layer, Marker, Tooltip)
-from folium.utilities import get_bounds
 from folium.vector_layers import PolyLine
 
 from jinja2 import Template
--- a/folium/features.py
+++ b/folium/features.py
@@ -10,21 +10,27 @@ from __future__ import (absolute_import, division, print_function)
 import json
 import warnings
 
-import requests
-import numpy as np
-
 from branca.colormap import LinearColormap, StepColormap
 from branca.element import (Element, Figure, JavascriptLink, MacroElement)
-from branca.utilities import (_locations_tolist, _parse_size, image_to_url,
-                              none_max, none_min, color_brewer)
+from branca.utilities import (
+    _iter_tolist,
+    _parse_size,
+    color_brewer,
+    image_to_url,
+    none_max,
+    none_min,
+)
 
 from folium.folium import Map
 from folium.map import (FeatureGroup, Icon, Layer, Marker, Tooltip)
-
 from folium.utilities import get_bounds
 from folium.vector_layers import PolyLine
 
 from jinja2 import Template
+
+import numpy as np
+
+import requests
 
 from six import binary_type, text_type
 
@@ -84,7 +90,7 @@ class RegularPolygonMarker(Marker):
                  fill_color='blue', fill_opacity=1, number_of_sides=4,
                  rotation=0, radius=15, popup=None, tooltip=None):
         super(RegularPolygonMarker, self).__init__(
-            _locations_tolist(location),
+            _iter_tolist(location),
             popup=popup, tooltip=tooltip
         )
         self._name = 'RegularPolygonMarker'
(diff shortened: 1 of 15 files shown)


folium/utilities.py
```diff
@@ -44,14 +39,14 @@ def _validate_coordinates(coordinates):
     if _isnan(coordinates):
         raise ValueError('Location values cannot contain NaNs, '
                          'got:\n{!r}'.format(coordinates))
-    coordinates = _locations_tolist(coordinates)
+    coordinates = _iter_tolist(coordinates)
     return coordinates
 
 
-def _locations_tolist(x):
+def _iter_tolist(x):
```
ocefpaf: @Conengmo sorry for the amount of noise in this PR. The fix for #928 is actually only this part :grimacing: and its use in the heatmap plugin.

folium/features.py
```diff
@@ -10,22 +10,28 @@
...
+    none_max,
+    none_min,
+)
 from folium.vector_layers import PolyLine
 
 from jinja2 import Template
 
+import numpy as np
+
+import requests
```
Conengmo: What's the import order in folium? First standard library, then branca, then folium, then Jinja, then other packages?
Also, empty lines between other packages?

folium/utilities.py
```diff
@@ -356,3 +351,38 @@ def camelize(key):
     """
     return ''.join(x.capitalize() if i > 0 else x
                    for i, x in enumerate(key.split('_')))
+
+
+def _parse_size(value):
+    try:
+        if isinstance(value, int) or isinstance(value, float):
```
Conengmo:
```suggestion
        if isinstance(value, (int, float)):
```

folium/utilities.py
```diff
@@ -356,3 +351,38 @@ def camelize(key):
...
+        msg = 'Cannot parse value {!r} as {!r}'.format
+        raise ValueError(msg(value, value_type))
+    return value, value_type
+
+
+def iter_points(x):
+    """Iterates over a list representing a feature, and returns a list of points,
+    whatever the shape of the array (Point, MultiPolyline, etc).
+    """
+    if isinstance(x, (list, tuple)):
```
Conengmo: I don't like these convoluted nested if statements. But I guess you just copied this from branca right?

Better IMO would be to exit early. 

```
if not isinstance(x, (list, tuple)):
  raise ValueError
# happy path continues
```

folium/utilities.py
```diff
@@ -356,3 +351,38 @@ def camelize(key):
     """
     return ''.join(x.capitalize() if i > 0 else x
                    for i, x in enumerate(key.split('_')))
+
+
```
Conengmo: These utility functions would be easy to write tests for.

folium/features.py
ocefpaf: Yep. Enforced by `flake8-import-order` (listed in the `requirements-dev.txt`). We used to enforced that in the tests and I turned it off when I added Stickler-CI. (I'm looking into another solution like Stickler but that will allow us to use all the flake8 plugins.)

folium/utilities.py
ocefpaf: That is better. (Yep. I just copied them.)

folium/utilities.py
ocefpaf: Indeed we need to test the functions in `utilities.py`. I would leave that to another PR though. This one drifted from its purpose a lot already :grimacing:

folium/features.py
Conengmo: I haven't used that plugin before, but seems sensible. But seems weird to me that folium comes before jinja, requests and numpy. Since folium is the local package I think it should come last.

> In general stdlib comes first, then 3rd party, then local packages

https://github.com/PyCQA/flake8-import-order#flake8-import-order

folium/utilities.py
Conengmo: That's okay :)

folium/utilities.py
Conengmo: Also something for the next one then.

folium/utilities.py
Conengmo: no problem!

folium/utilities.py
Conengmo: Not that important, so you can set it as resolved if you want.

folium/features.py
ocefpaf: You are correct. My editor does that automatically and, b/c we are not using relative imports, it does not know that `folium ` is local. Let's leave it be for now. I'll see if I can fix my editor to avoid triggering false positives and I'll fix those in another PR. (Sorry for the confusion.)

Conengmo: > I'll fix those in another PR

alright!

PR #1023, opened 2018-11-20
GeoJson Popup

jtbaker:
Added  GeoJsonPopup class that ~combines~ _extends_ functionality of GeoJsonTooltip ~and VegaLite existing functionality by database-style 'joining' key values from GeoJson to properties in the VegaLite chart data section.~


folium/features.py
```diff
@@ -725,93 +885,88 @@ class GeoJsonTooltip(Tooltip):
...
+    """
+
+    _template = Template(u"""
+        {% macro script(this, kwargs) %}
+        {{ this._parent.get_name() }}.bindPopup(""" +
+                         GeoJsonDetail.base_template +
+                         u""",{{ this.popup_options | tojson | safe }})
+                         {% endmacro %}""")
+
+    def __init__(self, fields=None, aliases=None, labels=True, style="margin: auto;", vegalite=None, map_key=None,
```
Conengmo: What do you think about giving this popup the same options as the 'regular' Popup class? I mean `show` and `sticky` options.

Conengmo: The default argument for `style` is a bit confusing. Is it necessary?

folium/features.py
```diff
@@ -725,93 +885,88 @@ class GeoJsonTooltip(Tooltip):
...
-            feature.get('properties') if feature.get('properties') is not None else key
-            for key, feature in enumerate(self._parent.data['features'])
-            if feature['geometry']['type'] == 'GeometryCollection'
-        ]
-        if any(geom_collections):
-            warnings.warn(
-                "GeoJsonTooltip is not configured to render tooltips for GeoJson GeometryCollection geometries. "
-                "Please consider reworking these features: {} to MultiPolygon for full functionality.\n"
-                "https://tools.ietf.org/html/rfc7946#page-9".format(geom_collections), UserWarning)
+    Follows the form of GeoJsonTooltip with `fields`, `aliases` arguments to embed GeoJson properties as a table in
```
Conengmo: Would be good to make the docstring selfcontained. So a user doesn't have to read the one of `GeoJsonTooltip` first.

folium/features.py
```diff
@@ -682,7 +687,156 @@ def get_bounds(self):
...
+            assert len(fields) == len(aliases), 'fields and aliases must have' \
+                                                ' the same length.'
+        assert isinstance(labels, bool), 'labels requires a boolean value.'
+        assert isinstance(localize, bool), 'localize must be bool.'
+        self._name = "GeoJsonDetail"
+        self.fields = fields
+        self.aliases = aliases if aliases is not None else fields
+        self.labels = labels
+        self.localize = localize
+        self.class_name = class_name
```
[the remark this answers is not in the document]
jtbaker: @Conengmo - sorry for the extended delay on the modifications to this PR - I think I've addressed most of your concerns with this revision.

re: the `class_name`: I thought it might be useful for users wanting to add custom css declarations to style the tooltips with a human-readable name. It's also allowed us to take the inline styling out of the html elements and add a top level styling to the class in the `head` of the document.

fullonic:
@Conengmo , @jtbaker  could we merge this feature into master? Or there is still missing something? All test checks are now passed.

In my opinion, having a tooltip as the only way, to add extra information to a marker is a big limitation of folium geojson.

When loading a geojson file from local drive is still possible loop through all points and create a normal marker with a popup, with is ok but doing extra work. However, if the data source is an url, there is no way to generate a popup for each marker. Also tooltips aren't available on mobile devices.

jtbaker:
> @Conengmo , @jtbaker could we merge this feature into master? Or there is still missing something? All test checks are now passed.
> 
> In my opinion, having a tooltip as the only way, to add extra information to a marker is a big limitation of folium geojson.
> 
> When loading a geojson file from local drive is still possible loop through all points and create a normal marker with a popup, with is ok but doing extra work. However, if the data source is an url, there is no way to generate a popup for each marker. Also tooltips aren't available on mobile devices.

Hey @fullonic, I'd like to get it merged too, if possible. When I originally wrote this, my git flow/skills were not very good. I've gotten a little more practice since then, and got all the commits that were in here squashed down into one.

Conengmo:
Nice Jason, I'll look at this PR again.

shape335:
Thanks for the follow up. I was just unsure what the status was as far as merging. Thank you for clarifying!

jtbaker:
> Thanks for the follow up. I was just unsure what the status was as far as merging. Thank you for clarifying!

NP @shape335. @Conengmo, think you might have a chance to give this another review in the near future?

Issue #1020, opened 2018-11-19
Transparent GeoJson


jackemcpherson:
Trying to create a transparent GeoJson, i only want to see the tooltips over a choropleth.

Trying the below, but getting nowhere.

```python
style_function = {'fillColor': '#00FFFFFF',
                 'lineColor': '#00FFFFFF'}

folium.GeoJson(combined,
               tooltip=folium.GeoJsonTooltip(fields=['LGA','MBRS'],
                                             aliases=['Location','Members']),
               style_function=style_function).add_to(m)

folium.LayerControl().add_to(m)

```
Can you tell me how to make the GeoJson layer transparent?

Thanks.

Jack.

jtbaker:
Hey @jackemcpherson, the issue is that your style function is a dictionary instead of a function returning a dictionary.

Rewrite it like this:

```
style = {'fillColor': '#00FFFFFF', 'lineColor': '#00FFFFFF'}

folium.GeoJson(combined,
               tooltip=folium.GeoJsonTooltip(fields=['LGA','MBRS'],
                                             aliases=['Location','Members']),
               style_function=lambda x: style).add_to(m)
```

or:

```
def style_function:
    return {'fillColor': '#00FFFFFF', 'lineColor': '#00FFFFFF'}

folium.GeoJson(combined,
               tooltip=folium.GeoJsonTooltip(fields=['LGA','MBRS'],
                                             aliases=['Location','Members']),
               style_function=style_function).add_to(m)
```

and that should be good. Hope this helps!

jackemcpherson:
Thanks a lot for getting back to me, both options worked!

Just for anybody finding this in future: replace the dict from my code with:

`{'fillColor': '#00000000', 'color': '#00000000'}`

Conengmo:
@jtbaker maybe assert `style_function` is callable, raise a meaningful error otherwise? What do you think, user-friendly or overkill?

jtbaker:
@Conengmo! Good thinking! I like it. I've PR'd a fix in #1024.

paaghi:
I came across a similar problem like yours. Try to set the 'fillOpacity' property in your style function. Because according to [the original Leaflet documentation](https://leafletjs.com/reference-1.5.0.html#path-option), the default value is 0.2. I hope it helps.

mmgamboa:
Thanks @paaghi I came with the same problem and your comment solved it out. Cheers, Martín.

yoondev23:
[...]

PR #1024, opened 2018-11-22
GeoJson StyleCheck

jtbaker:
Added a check on the GeoJson class constructor to ensure that `style_function` and `highlight_function` are functions, and that they return dictionaries.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -419,11 +419,10 @@ class GeoJson(Layer):
 
         callers = {'style_function': style_function, 'highlight_function': highlight_function}
 
-        for caller_key in callers:
-            caller = callers[caller_key]
+        for key, caller in callers.items():
             if caller is not None:
-                assert callable(caller) and isinstance(caller(self.data['features'][0]), dict),\
-                    "Pass a function returning a dictionary to `{}`.".format(caller_key)
+                if not callable(caller) or not isinstance(caller(self.data['features'][0]), dict):
+                    raise ValueError('Pass a function returning a dictionary to `{}`.'.format(key))
 
         self.style_function = style_function or (lambda x: {})
 
--- a/folium/features.py
+++ b/folium/features.py
@@ -438,7 +438,7 @@ class GeoJson(Layer):
         Tests `self.style_function` and `self.highlight_function` to ensure
         they are functions returning dictionaries.
         """
-        test_feature = self.data if self.data.get('features') is None else self.data['features'][0]  #noqa
+        test_feature = self.data if self.data.get('features') is None else self.data['features'][0]  # noqa
         if not callable(func) or not isinstance(func(test_feature), dict):
             raise ValueError('{} should be a function that accepts items from '
                              'data[\'features\'] and returns a dictionary.'
--- a/folium/features.py
+++ b/folium/features.py
@@ -417,10 +417,13 @@ class GeoJson(Layer):
         else:
             raise ValueError('Unhandled object {!r}.'.format(data))
 
-        for caller in (style_function, highlight_function):
+        callers = {'style_function': style_function, 'highlight_function': highlight_function}
+
+        for caller_key in callers:
+            caller = callers[caller_key]
             if caller is not None:
-                assert callable(caller) and isinstance(caller(self.data['features'][0]), dict), \
-                    "Pass a function returning a dictionary to `{}`.".format(caller)
+                assert callable(caller) and isinstance(caller(self.data['features'][0]), dict),\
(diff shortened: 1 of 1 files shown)


folium/features.py
```diff
@@ -416,6 +416,15 @@ def __init__(self, data, style_function=None, name=None,
             self.data = json.loads(json.dumps(data.__geo_interface__))  # noqa
         else:
             raise ValueError('Unhandled object {!r}.'.format(data))
+
+        callers = {'style_function': style_function, 'highlight_function': highlight_function}
+
+        for caller_key in callers:
+            caller = callers[caller_key]
+            if caller is not None:
```
ocefpaf: No need for the `None` and `caller = callers[caller_key]` and we can drop the assert here in lieu of Python Exceptions. You can use `.items()` like:

```python
for key, caller in callers.items():
   if not callable(caller) and not isinstance(caller(self.data['features'][0]), dict):
       raise ValueError(f'{key} must be a callable function returning a dictionary.')
```

jtbaker: Nice, I always forget about `.items()` on dicts. Been spending more time in the JS world lately.

Don't we need the `not None` check though? The default values for `style_function` and `highlight_function` are `None`s right now - so they'll fail this check if unmodified.

ocefpaf: > Don't we need the `not None` check though?

`if not callable(caller)` should take care of that if I'm not mistaken. The `caller` will be `None` and will be `True` in `not callable(caller)`, but I believe the `and` should be an `or`. We want to raise if one or the other is `True`, right?

> Also, will our CI travis build environment accept the f-string usage? Last time I tried to commit one it caused a failure so I've been using `.format()` instead.

I wrote that quickly and I use f-strings everywhere, it was not meant to be a copy-n-paste review, sorry.

With that said, we will drop Python 2.7 support in the next release and we can use a f-string backport package for py35, so I'm OK if we start introducing them now. Let's just wait for @Conengmo opinion on this one :smile:

jtbaker: > The caller will be None and will be True in not callable(caller), but I believe the and should be an or. We want to raise if one or the other is True, right?

But this will pass any other uncallable object, in addition to `None`, I'm pretty sure - like the dictionary in #1020.

IMO there should be only two clear passing cases for the callers - `None` and functions returning dicts.

Calling an uncallable (like `None(self.data['features'][0])`) raises `TypeError`. So we need the `and` operator to take the `False` from `callable` if our second call results in that `TypeError`, and still raise our custom `ValueError`.

So we need:

```
if not isinstance(caller, None):
    if not callable(caller) and isinstance(caller(self.data['features'][0]), dict):
        raise ValueError('Pass a function returning a dictionary to `{}`'.format(key))
```

Or maybe pack all the conditionals into one line? I know it is long, but I want to make it as airtight as possible and give useful feedback to the user. Let me know what you think.

folium/features.py
```diff
@@ -416,6 +416,14 @@ def __init__(self, data, style_function=None, name=None,
             self.data = json.loads(json.dumps(data.__geo_interface__))  # noqa
         else:
             raise ValueError('Unhandled object {!r}.'.format(data))
+
+        callers = {'style_function': style_function, 'highlight_function': highlight_function}
+
+        for key, caller in callers.items():
+            if caller is not None:
+                if not callable(caller) or not isinstance(caller(self.data['features'][0]), dict):
+                    raise ValueError('Pass a function returning a dictionary to `{}`.'.format(key))
```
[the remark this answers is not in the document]
jtbaker: > Can we expand this a bit?

Yes, I like this and it's more clear. Thanks for the feedback.

Conengmo:
> What is the best design pattern here?

I guess if we can't choose, then it doesn't matter what we choose :)

ocefpaf:
> I guess if we can't choose, then it doesn't matter what we choose :)

:smile: 

> What is the best design pattern here?

More inclined to follow the pattern where you don't have to check for `None`.

jtbaker:
>I guess if we can't choose, then it doesn't matter what we choose :)

Haha, I wasn't sure if there was a 'pythonic' or OOP best practice way to manage that kind of stuff.

Sounds good - I'll go ahead and rewrite it to a `_validate_function` structure and call it after assignment.

jtbaker:
> By the way the Travis build is failing on a single notebook related test:
https://travis-ci.org/python-visualization/folium/jobs/459020933#L2934

Looks like I was able to fix the failing test by handling the case where `self.data` is a single GeoJson feature instead of a FeatureCollection.

Conengmo:
Good stuff, thanks @jtbaker! And thanks @ocefpaf for reviewing this.

PR #1026, opened 2018-11-22
Fix publication

ocefpaf:
Mostly boilerplate testing stuff. I'm also working on a better `flake8` test for the notebooks but that will be in another PR.


ocefpaf:
> Seems the only functional change is that we now check PEP8 compliance in the notebooks?

In theory we checked for those before but it was broken. (I still have a minor issue to fix but that is pending due to a PR in nbconvert.)

> I didn't check the notebook contents themselves, but I guess you applied PEP8 fixes.

Some of them, there are two notebooks failing but the CI still return "green" b/c of the error code in the loop.

> Looks good to me! Is this PR ready to merge then or do you want to add anything?

Feel free to merge. I'll send the continuation once https://github.com/jupyter/nbconvert/pull/919 is merged and I can deploy my nblint-like too.

PR #1029, opened 2018-11-25
Fix missing and redundant and imports in __init__

Conengmo:
In #1017 the import of `GeoJsonTooltip` got accidently removed from `__init__.py`. This PR restores it.

I also removed three unused entries in the `__all__` list: `GeoJsonStyle` and `Polyline` don't exist (`PolyLine` does though).  `MarkerCluster` is a plugin.

Once this PR is merged we can continue with #988.


folium/__init__.py
```diff
@@ -88,7 +88,6 @@
     'MacroElement',
     'Map',
     'Marker',
-    'MarkerCluster',
```
ocefpaf: My bad! (Probably a bad rebase :-/)

Do we want to remove `MarkerCluster` from this list?

Conengmo: It's not imported in `folium/__init__.py`, instead it's available in `folium/plugins/__init__.py`. So it shouldn't be listed in the `__all__` of `folium/__init__.py` I think.

ocefpaf: I guess it was left here for historic reasons. I prefer to import from plugins like the rest.

PR #1032, opened 2018-11-29
Added SideBySide plugin basic funcionality

kikocorreoso:
This PR adds basic functionality provided by [this Leaflet plugin](https://github.com/digidem/leaflet-side-by-side) similar to the [ipyleaflet's SplitMap control](https://ipyleaflet.readthedocs.io/en/latest/api_reference/split_map_control.html).

I've added a notebook in the `examples/` directory.

Please, let me know if there is something wrong or if the plugin is not an interesting plugin for folium.

### DIFF
--- a/examples/SideBySide.ipynb
+++ b/examples/SideBySide.ipynb
@@ -1,0 +1,85 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import folium\n",
+    "from folium import TileLayer, WmsTileLayer\n",
+    "from folium.plugins import SideBySide"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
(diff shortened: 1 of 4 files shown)


Conengmo:
Hi, thanks for submitting this PR! I was working on something similar in #933, based on another Leaflet plugin. No reason to have two split map type of plugins, so which Leaflet plugin do you think is better?

kikocorreoso:
Hi @Conengmo 

Sorry, I didn't see your PR.

Well, maybe there is room for both. I think the functionality is not exactly the same.

Conengmo:
Can you tell a bit about what's the difference?

kikocorreoso:
Well, using the `Sync` plugin you can show and sync two different maps with different layers so it is useful to provide more information on the same area and to visually compare Several maps. 

With the `SideBySide` plugin you can have the same but on the same map window.

So, with `Sync` you have to divide the size of the map window in two or more. With the `SideBySide` you can have a complete view of a map and visually check differences on the same place just moving the control. With `Sync` plugin you can have something similar using the `Sync.Cursor` (https://raw.githack.com/jieter/Leaflet.Sync/master/examples/syncCursor_dual.html) but, again, it is not exactly the same and you have a smaller window for both maps.  On small screens I would like to have the `SideBySide` functionality available and, for me, it would be quicker to just move the control to check the differences on the same pixel (or group of pixels) without having to add more code to use the `Sync.Cursor` functionality.

I think both plugins are useful. `SideBySide` is for a very specific use while `Sync` seems more complete.

I don't know if this answer your question.

Conengmo:
This functionality got merged in https://github.com/python-visualization/folium/pull/1292.

Thank you for your effort @kikocorreoso and apologies this PR didn't work out.

kikocorreoso:
Thanks for adding the functionality 🗺️ !!!!!

PR #1035, opened 2018-12-06
add options to draw control

EtsuNDmA:
Add options for [draw control](http://leaflet.github.io/Leaflet.draw/docs/leaflet-draw-latest.html#options). Also add custom filename for geojson file

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 
+
 from __future__ import (absolute_import, division, print_function)
+
+import json
 
 from branca.element import CssLink, Element, Figure, JavascriptLink, MacroElement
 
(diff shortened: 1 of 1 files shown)


folium/plugins/draw.py
```diff
@@ -10,28 +13,36 @@
...
     ----------
     export : bool, default False
         Add a small button that exports the drawn shapes as a geojson file.
-
+    filename : string, default data.geojson
+        Name of geojson file
+    position : string, default 'topleft'
+        Position of control. It can be one of 'topleft', 'toprigth', 'bottomleft', 'bottomright'
+    draw : dict, default None
+        The options used to configure the draw toolbar
```
Conengmo: 1. Can you link to a page where these options are listed? Same for `edit`.
2. I'm not sure about the names `draw` and `edit`, they are too general. The class is named `Draw`, so an object is probably named `draw`. For these arguments, how about `draw_options` and `edit_options`?

folium/plugins/draw.py
```diff
@@ -10,28 +13,36 @@
...
+    edit : dict, default None
+        The options used to configure the edit toolbar
     Examples
     --------
     >>> m = folium.Map()
     >>> Draw(export=True).add_to(m)
-
```
Conengmo: Why did you remove this whitespace?

folium/plugins/draw.py
```diff
@@ -43,37 +54,48 @@ class Draw(MacroElement):
...
+              document.getElementById('export').setAttribute(
+                'download',
+                "{{kwargs.get('filename', 'data.geojson')}}"
+              );
+            }
             {% endmacro %}
             """)
 
-    def __init__(self, export=False):
+    js_url = 'https://cdnjs.cloudflare.com/ajax/libs/leaflet.draw/1.0.2/leaflet.draw.js'
```
Conengmo: Why did you move the URLs here? Throughout folium they are in the `render` method, so I'd prefer to keep them there for consistency.

folium/plugins/draw.py
```diff
@@ -10,28 +13,36 @@
 class Draw(MacroElement):
     """
     Vector drawing and editing plugin for Leaflet.
-
     Parameters
     ----------
     export : bool, default False
         Add a small button that exports the drawn shapes as a geojson file.
-
+    filename : string, default data.geojson
```
Conengmo:
```suggestion
    filename : string, default 'data.geojson'
```

folium/plugins/draw.py
```diff
@@ -10,28 +13,36 @@
...
+        Name of geojson file
+    position : string, default 'topleft'
+        Position of control. It can be one of 'topleft', 'toprigth', 'bottomleft', 'bottomright'
+    draw : dict, default None
```
Conengmo:
```suggestion
    draw : dict, optional
```

folium/plugins/draw.py
```diff
@@ -10,28 +13,36 @@
...
+        The options used to configure the draw toolbar
+    edit : dict, default None
```
Conengmo:
```suggestion
    edit : dict, optional
```

folium/plugins/draw.py
```diff
@@ -43,37 +54,48 @@ class Draw(MacroElement):
...
+        self.filename = filename
+        self.position = position
+        self.draw = draw or {}
+        self.edit = edit or {}
 
     def render(self, **kwargs):
-        super(Draw, self).render()
+        super(Draw, self).render(
+            position=self.position,
+            draw=json.dumps(self.draw),
```
Conengmo: These variables are already available in `self`, you don't have to pass them here. You can do the json dumping in `__init__`.

EtsuNDmA:
Hi! @Conengmo, I took into account your comments. Also add an example to dockstring

Conengmo:
Looks good @EtsuNDmA! I'll merge it after checking it one last time, but I don't think any more changes are necessary.

Conengmo:
Thanks @EtsuNDmA!

PR #1039, opened 2018-12-11
Have GeoJson raise a more descriptive error when passed a GeoDataFrame with missing geometries

### LINKED ISSUE #1037: folium.features.GeoJson could give a more descriptive error when passing GeoDataFrames with missing geometries
I frequently pass GeoDataFrames to folium.features.GeoJson. When GeoDataFrames with missing geometries are passed, it gives `ValueError: Unhandled object` and then prints the table. It might be better if it said something more descriptive about how the GeoDataFrame was missing geometries.

[This](https://github.com/python-visualization/folium/blob/master/folium/features.py#L412) appears to be the relevant line. GeoDataFrames with missing geometries do not have the `__geo_interface__` property. Not sure if this is something worth addressing in Folium or in GeoPandas.

#### Replicate error with
```python
import geopandas as gpd
import folium as f

df = gpd.read_file(gpd.datasets.get_path('nybb'))
df.loc[0, 'geometry'] = None

f.features.GeoJson(df)
```

#### Output
```
ValueError: Unhandled object    BoroCode       BoroName     Shape_Leng    Shape_Area  \
0         5  Staten Island  330470.010332  1.623820e+09   
1         4         Queens  896344.047763  3.045213e+09   
2         3       Brooklyn  741080.523166  1.937479e+09   
3         1      Manhattan  359299.096471  6.364715e+08   
4         2          Bronx  464392.991824  1.186925e+09   

                                            geometry  
0                                               None  
1  (POLYGON ((1029606.076599121 156073.8142089844...  
2  (POLYGON ((1021176.479003906 151374.7969970703...  
3  (POLYGON ((981219.0557861328 188655.3157958984...  
4  (POLYGON ((1012821.805786133 229228.2645874023...  .
```
`0.6.0`

wtadler:
Closes #1037.


folium/features.py
```diff
@@ -409,11 +409,13 @@ def __init__(self, data, style_function=None, name=None,
             else:  # This is a filename
                 with open(data) as f:
                     self.data = json.loads(f.read())
-        elif hasattr(data, '__geo_interface__'):
+        elif hasattr(data, 'geometry'):
```
ocefpaf: `__geo_interface__` takes any object that implements the `__geo_interface__` but `to_json()` is `geopandas` only.

wtadler: GeoDataFrames that have any rows with missing geometries don't implement `__geo_interface__`.

So should this `elif` block be kept the same and have an additional block, `elif hasattr(data, 'geometry') and any(data['geometry'].isnull()):`?

ocefpaf: Interesting, so something is wrong with that logic b/c then you should be directed to the `else` and things should fail as expected, no? If that is the case all you have to do is the improve the error message in the `else`.

wtadler: Right, but the idea is to have an error message specifically for this case, and leave the existing, vague error message for other cases that fail.

wtadler: Ah, okay. So there are no reasons besides missing geometries that you could get to the `else` block in the current code?

ocefpaf: > Ah, okay. So there are no reasons besides missing geometries that you could get to the `else` block in the current code?

That is not what I said, but in a way any "bad" object does not have a `geometry`, so the message are equivalent. One better way to phrase that would be:

`Could not find valid geometries in {!r}`, so the object `__repr__` is printed in the error.

wtadler: I just think that people using GeoDataFrames that have, say, a single row with a missing geometry, might be confused by that, and would appreciate a more descriptive message.

ocefpaf: I believe that the error message above is descriptive enough but feel free to improve on it. However, the current if-clauses would kill all objects that we accept via the geo interface and that is a no-no.

folium/features.py
```diff
@@ -414,8 +414,10 @@ def __init__(self, data, style_function=None, name=None,
             if hasattr(data, 'to_crs'):
                 data = data.to_crs(epsg='4326')
             self.data = json.loads(json.dumps(data.__geo_interface__))  # noqa
+        elif hasattr(data, 'geometry') and any(data['geometry'].isnull()):  # for GeoDataFrames with missing geometries
```
ocefpaf: This needs to be nested inside the `elif` above or it won't be executed.
You can also drop the comment.

wtadler: But GeoDataFrames with missing geometries have no `__geometry_interface__` as I noted in #1037, so it should be executed, no?

ocefpaf: Indeed. I forgot about that. Sorry, my mind is one the generic objects that implement that and should fail if not. That is why the dual message is unnecessary IMO. Anyway, if you are happy with this we can merge it.

wtadler: Okay, I think it's ready.

folium/features.py
```diff
@@ -414,8 +414,10 @@ def __init__(self, data, style_function=None, name=None,
...
+            raise ValueError('Cannot render objects with any missing geometries. {!r}'.format(data))
```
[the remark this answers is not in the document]
wtadler: Yep, I think that's fine.

Conengmo:
Thanks for helping with this @wtadler!

PR #1040, opened 2018-12-11
Popup width

ocefpaf:
@Conengmo this solves problems like https://github.com/python-visualization/folium/pull/959#issuecomment-446306611


Conengmo:
Looks good @ocefpaf. One caveat though is that popups with only text become more narrow than they used to be. For example the 'Fancy HTML popup':

http://nbviewer.jupyter.org/github/ocefpaf/folium/blob/popup_width/examples/Popups.ipynb#Fancy-HTML-popup
http://nbviewer.jupyter.org/github/ocefpaf/folium/blob/master/examples/Popups.ipynb#Fancy-HTML-popup

A solution is that now users of text have to give in a value in pixels. But maybe there's a way to have it work automatically for both graphics and text?

ocefpaf:
> Looks good @ocefpaf. One caveat though is that popups with only text become more narrow than they used to be. For example the 'Fancy HTML popup'

Yep. In a way that is the "default" `100%` width reserved for text from leaflet. We can always explicitly request more like in cell number [3].

> A solution is that now users of text have to give in a value in pixels. But maybe there's a way to have it work automatically for both graphics and text?

We could try to add some logic to add that behavior. I understand that the behavior in this PR is "kind of a breakage" from the previous one, but now things are more closely related to how leaflet works and less "magic."

I'm not against checking for text vs other object to use two defaults, 300px and 100% respective, but I have to confess I'm not in love with the idea of the extra complexity there.

democat3457:
This PR seems to break default popup widths, as Leaflet does not support percentage values for maxWidth - they simply [grab the raw maxWidth value](https://github.com/Leaflet/Leaflet/blob/c511e66c20dce06ac05c43cf287b3ecdba00d924/src/layer/Popup.js#L233) to use in their width calculations, which means that the `width` they set on the content element is `NaN`, resulting in a blank style attribute:
image

Conengmo:
@democat3457 you're right, good find! This change actually works by disabling setting the width of the popup, which makes it stretch to fit the content. In a way that is the desired behavior :) 

I think this deserves its own issue. Are you interested in helping with this issue?

I did a quick look and it seems that we should solve this using css. We should think about how we should solve this. Maybe have a `Popup` add a css element that sets its width to either a percentage or `auto`?

PR #1047, opened 2018-12-31
Update to Leaflet 1.4.0

Conengmo:
Leaflet 1.4.0 has been released:
https://github.com/Leaflet/Leaflet/releases/tag/v1.4.0

I looked through the changelog and it seems there are no breaking changes. Only improvements and bug fixes. So I think it's quite save to adopt this now. Otherwise, we can wait a bit and merge this PR in a week or two, when any critical bugs will probably have been found.


ocefpaf:
Let's go for it! Thanks!!

PR #1049, opened 2019-01-07
Polygons that surround locations

HZALK:
I was asked to create polygons that surround all points from a given list of locations. This is a simple task. However, it required me some research on the how to and it would be much easier if it was a folium build-in function. So I wrote it.

### DIFF
--- a/folium/__init__.py
+++ b/folium/__init__.py
@@ -45,7 +45,7 @@ from folium.map import (
     Tooltip,
 )
 from folium.raster_layers import TileLayer, WmsTileLayer
-from folium.vector_layers import Circle, CircleMarker, PolyLine, Polygon, Rectangle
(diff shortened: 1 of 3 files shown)


ocefpaf:
@HZALK `folium.vector_layers` mimics `leaflet`'s vector layer and `ConvexHull_Polygon_from_Points` and `Envelope_Polygon_from_Points` are not `leaflet` original vector layer. This PR could be a plugin but looks like the extra dependency on scipy and the simplicity to create a `ConvexHull` one could just add an example to our docs instead on a new class. What do you think @Conengmo?

Conengmo:
I think you’re right @ocefpaf. Maybe you want to contribute this as an example notebook @HZALK? Might help someone else with your challenge even better then when we hide it in the source code. You can look in the examples folder and create a new notebook which explains what you want to do and how.

HZALK:
@ocefpaf  and @Conengmo: Thank you for your review and the advice. Wilco (probably next weekend).

PR #1053, opened 2019-01-13
Draggable markers

Conengmo:
This PR adds a new argument to the `Marker` class. By setting `draggable=True` you can drag the markers around the map. I don't know if anybody really needs this behavior, but I thought it was kind of nice :)

If we think this feature is not necessary, we could consider to alter `Marker` to accept `**kwargs` and pass those as options to the Leaflet object.


Conengmo:
If you think this is okay than this is it!

PR #1055, opened 2019-01-19
Create Polygons from List of Points.ipynb

HZALK:
As suggested by reviewers I've created an example notebook that shows the application of convex hull and envelope to create polygons from a list of points in folium maps.


HZALK:
Continuous-integration did not like the sequence of the cells. I've changed that and will open an new pull request.

PR #1056, opened 2019-01-19
Polygon from list of points example

HZALK:
As suggested by reviewers I've created an example notebook that shows the application of convex hull and envelope to create polygons from a list of points in folium maps. Since continuous-integration did not like the sequence of my first request, this is an updated version.


ocefpaf:
Awesome addition to our example gallery. Thanks @HZALK!

Can you please change the last line to `my_map_global` instead of `my_map_global.save('example_map.html')` so the map is displayed in the repr?

HZALK:
Thank you, @ocefpaf :-)
I've changed the respective line.

Conengmo:
Good work @HZALK, this will make a good addition to the example gallery. I have a few comments to possibly improve it, but none of these are critical, so do what you like and let me know which ones you want to skip.

* You use two different variable/function name formats. It's best to stick to the PEP8 one: `variable_name` and `def function_name()`. So `DrawPoints()` would become `draw_points()`. That way we have consistent styling in the folium library.
* Don't make your lines too long. It's best if we don't have to scroll sideways to read what's going on.
* For the example I think it would be better to show a map after each polygon type. So explain a bit about convex hull, then immediately show the code and the map with the convex hull. That way it is clear what the section is about. You can still show a final map at the end combining them.
* The polygon line and the points have the same color, and the line runs over the points. It's hard to see the points under the line. Maybe make the points a different color/shade (and thicker) and have them put on top of the line.
* If you use `LayerControl(collapsed=False)` it shows the controls, which is better I think because not everyone may know you can hover the control button in the top right corner.

HZALK:
Thank you @Conengmo for you great feedback and suggestions. I’ve addressed all points you mentioned.

Conengmo:
Merged! Thanks @HZALK for your contribution! If you want to help out again you're very welcome to do so.

HZALK:
Thanks @Conengmo. It was a pleasure. If have a new idea, I'll be glad to contribute again. :-)

PR #1058, opened 2019-01-21
Geojson styling overhaul

Conengmo:
I've been thinking about how to decouple the GeoJson data and things like styling. This could reduce the html filesize and prevent unexpected behavior where we edit someones data. I spent some time with the `GeoJson` class and made some changes:

* Style and highlight is no longer added to the data, but stored in a separate mapping.
  * This mapping is compact, features sharing parameters are defined once.
  * It needs a field to identify a feature with. I made a heuristic function that tries to find one, otherwise tries to add an `id` field to the data, and throws an error if it fails.
* It is possible to disable embedding the data for files and URLs by setting `embed=False`.
* ~~It is possible to create the object without loading the data (only for files and URLS with `embed=False` and no style or highlight).~~
* Highlighter now works also without `style_function`. It resets the style on `mouseout`.
* I made some things explicit, should as the conversion to `FeatureCollection`.

With better style mappings and an option to disable embedding, html files can become a lot smaller.

Here's a code snippet to try it out with:

```python
highlight_function = lambda x: {'weight': 10 if x['geometry']['type'] == 'MultiPolygon' else 0}

def style_function(feature):
    style = {
        'opacity': 1.0,
        'fillColor': 'green',
        'color': 'black',
        'weight': 2
    }
    if feature['properties']['name'] == 'Alabama':
        style['fillColor'] = 'red'
    if feature['properties']['name'] == 'Utah':
        style['fillColor'] = 'yellow'
    return style

data = 'us-states.json'
# data = 'https://raw.githubusercontent.com/python-visualization/folium/master/tests/us-states.json'
folium.GeoJson(
    data,
    style_function=style_function,
    highlight_function=highlight_function,
    embed=False,
).add_to(m)
```

@jtbaker do you want to review? I'm wondering if it's a bit too much. Maybe we can simplify it by dropping some functionality. On the other hand, this doesn't increase complexity for the users other than the `embed` option. Would love to hear what you think about this.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -518,10 +518,11 @@ class GeoJson(Layer):
         if self.data['type'] == 'FeatureCollection':
             return
         if not self.embed:
-            raise ValueError('Data is not a FeatureCollection, but it should be'
-                             ' to apply style or highlight. Because `embed='
-                             'False` it cannot be converted into one.\n'
-                             'Either set `embed=True` or disable styling.')
+            raise ValueError(
+                'Data is not a FeatureCollection, but it should be to apply '
+                'style or highlight. Because `embed=False` it cannot be '
+                'converted into one.\nEither change your geojson data to a '
+                'FeatureCollection, set `embed=True` or disable styling.')
         # Catch case when GeoJSON is just a single Feature or a geometry.
         if 'geometry' not in self.data.keys():
             # Catch case when GeoJSON is just a geometry.
@@ -556,7 +557,7 @@ class GeoJson(Layer):
         raise ValueError(
             'There is no unique identifier for each feature and because '
             '`embed=False` it cannot be added. Consider adding an `id` '
-            'field or set `embed=True`. '
+            'field to your geojson data or set `embed=True`. '
         )
 
     def _get_self_bounds(self):
--- a/folium/features.py
+++ b/folium/features.py
@@ -9,6 +9,8 @@ from __future__ import (absolute_import, division, print_function)
 
 import json
 import warnings
+import functools
+import operator
 
 from branca.colormap import LinearColormap, StepColormap
 from branca.element import (Element, Figure, JavascriptLink, MacroElement)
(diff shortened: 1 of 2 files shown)


folium/features.py
```diff
@@ -519,10 +570,77 @@ def _get_self_bounds(self):
...
+                        content[key] = "{{'" + value.get_name() + "'}}"
+            key = self._to_key(content)
+            mapping.setdefault(key, []).append(self.get_feature_id(feature))
+        self._set_default_key(mapping)
+        return mapping
+
+    def get_feature_id(self, feature):
+        """Return a value identifying the feature."""
+        fields = self.feature_identifier.split('.')[1:]
+        return functools.reduce(operator.getitem, fields, feature)
```
ocefpaf: Nice!

Conengmo: Thanks :) Found it on stackoverflow, looks a bit like magic. I'll definitely write a test for it.

folium/features.py
```diff
@@ -393,121 +398,167 @@ class GeoJson(Layer):
     """
     _template = Template(u"""
         {% macro script(this, kwargs) %}
+        {%- if this.style %}
+        function {{ this.get_name() }}_styler(feature) {
+            switch({{ this.feature_identifier }}) {
+                {%- for style, ids_list in this.style_map.items() if not style == 'default' %}
+                {% for id_val in ids_list %}case "{{ id_val }}": {% endfor %}
```
jtbaker: Hey @Conengmo, starting to work my way through this now. I'll leave comments as I go. This a library-wide design pattern I suppose, but for interpolating literals into the JS templates, how would you feel about using Jinja2's `tojson` method instead of hardcoding in the string quotes around the interpolaters. This would have the added benefit of raising an error (I think) at the python runtime if an object is not JSON serializable.

So this line might look like:
```
{% for id_val in ids_list %}case {{ id_val | tojson | safe }}: {% endfor %}
```
Docs on this feature:

http://flask.pocoo.org/docs/0.12/templating/#standard-filters

Let me know what you think. I think it could be a good option to standardize as a sort of consistent style guide across different classes - and allow us to flexibly, but consistently embed dtypes including strings as well as others (numbers, lists, dicts) with less template hardcoding.

Conengmo: Hi Jason, nice idea! Let's do this throughout the project. Maybe make our own filters that do the various mappings? Or is `tojson` suitable for all? If it's okay with you I think we should do that in a separate PR. Maybe you want to open an issue for this?

jtbaker: Yes, that sounds great. I'll work in that after we get through getting this merged.

folium/features.py
```diff
@@ -393,121 +398,167 @@ class GeoJson(Layer):
     """
     _template = Template(u"""
         {% macro script(this, kwargs) %}
+        {%- if this.style %}
+        function {{ this.get_name() }}_styler(feature) {
+            switch({{ this.feature_identifier }}) {
```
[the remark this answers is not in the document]
Conengmo: My bad. Don't know what went wrong there, it should be the result of the `find_identifier()` method. Will fix it.

folium/features.py
```diff
@@ -393,121 +398,167 @@ class GeoJson(Layer):
...
+        self.highlight = highlight_function is not None
+
         if isinstance(data, dict):
             self.embed = True
             self.data = data
         elif isinstance(data, text_type) or isinstance(data, binary_type):
-            self.embed = True
             if data.lower().startswith(('http:', 'ftp:', 'https:')):
-                self.data = requests.get(data).json()
+                if self.embed or self.style or self.highlight:
```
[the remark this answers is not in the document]
Conengmo: > `self.data` should still be assigned for other classes to reference for validation

You're right. There was one person who wanted folium to be able to creating html without folium accessing the data, but being able to do proper validation is more important. We cannot know at this point if a user wants to add `GeoJsonTooltip` or not, so we need to set `self.data`.

folium/features.py
```diff
@@ -519,10 +570,77 @@ def _get_self_bounds(self):
 
     def render(self, **kwargs):
         self.parent_map = get_obj_in_upper_tree(self, Map)
-        self.json = self.style_data() if self.embed else json.dumps(self.data)
+        if self.style or self.highlight:
+            mapper = GeoJsonStyleMapper(self.data, self.find_identifier(),
```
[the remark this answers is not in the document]
Conengmo: I moved it to the constructor and pass it to the style mapper as you proposed.

Conengmo:
The tests were failing because the `TimeSliderChoropleth` plugin got broken. It integrates with the `GeoJson` class. I fixed it by making  `TimeSliderChoropleth` no longer inherit from `GeoJson`. Instead, it lightly reuses some of its components. For that I moved the data processing part of the `GeoJson` class to a method instead of having it in `__init__`.

Conengmo:
I think it's ready, @jtbaker what do you think, okay to merge or do you have more comments?

jtbaker:
Hey @Conengmo, I gave it another look over and everything looks good to me. I give it the 👍 for the merge and am pretty excited about it. Thanks for the thought put into implementing this feature, I think it will really make the library more performant and while still retain control of the styling python-side.

Conengmo:
Thanks for the kind words Jason, I appreciate it!

Let’s wait with the merge until 0.8.1 is released, I think this change is big enough to go in 0.9.0.

Issue #918, opened 2018-07-22
Validate `key_on` in choropleth


Conengmo:
There have been issues in the (recent) past where users couldn't figure out why their choropleth wasn't working. Such as #913, #917, #793. These issues where due to a mismatch between the geojson data and the numerical data. The easiest mistake is to have a wrong `key_on` value. In #917 the value in the geojson and the numerical data had a different type leading to a mismatch. 

This can be solved by not making a mismatch happen silently. There are things we can validate before or during rendering that can lead to a meaningful error message:
- whether the `key_on` argument matches the fields in the geojson.
- whether both fields in the `column` argument are present in the data.
-  whether the values in the geojson and the numerical data under the `key_on` and `column[0]` argument match.
- whether those matches lead to valid values using the `column[1]` argument.

If someone wants to pick this up, you're welcome to do so!

evwhiz:
Created #1144. @ocefpaf - please review that PR. You were great today :- ) Many Thanks!

Conengmo:
We haven't seen any questions on this topic recently, so I assume our improvements payed off.

PR #1060, opened 2019-02-05
Adding optional tooltip arg to Choropleth Class

Gordonei:
# Overview
The tooltip arg just gets passed through transparently to the underlying GeoJSON, or TopoJSON objects. I propose it has the same default, `None`, which makes the pass through cleaner.

This isn't earth shattering, but it is the difference between doing this:
```python
choropleth = folium.features.Choropleth(geodata,
                               key_on='feature.properties.index',
                               data=non_geodata,
                               columns=["index","data1"])

choropleth.geojson.add_child(
    folium.features.GeoJsonTooltip(['data1'])
)
```
vs
```python
choropleth = folium.features.Choropleth(geodata,
                               key_on='feature.properties.index',
                               data=non_geodata,
                               columns=["index","data1"],
                               tooltip=folium.features.GeoJsonTooltip(['data1'])
)
```

# Tests
## Linting
Passing
## Testing
Pass, except for the one xfail in `test_time_slider_choropleth.py`:
```python
@pytest.mark.xfail
def test_timedynamic_geo_json():
```

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -896,6 +896,9 @@ class Choropleth(FeatureGroup):
         Whether the Layer will be included in LayerControls.
     show: bool, default True
         Whether the layer will be shown on opening (only for overlays).
+    tooltip: GeoJsonTooltip, Tooltip or str, default None
+        Display text when hovering over the GeoJSON object. Can utilize the data,
+        see folium.GeoJsonTooltip for info on how to do that.
 
     Returns
     -------
@@ -924,7 +927,7 @@ class Choropleth(FeatureGroup):
                  bins=6, fill_color='blue', nan_fill_color='black',
                  fill_opacity=0.6, nan_fill_opacity=None, line_color='black',
                  line_weight=1, line_opacity=1, name=None, legend_name='',
-                 overlay=True, control=True, show=True,
+                 overlay=True, control=True, show=True, tooltip=None,
                  topojson=None, smooth_factor=None, highlight=None,
                  **kwargs):
         super(Choropleth, self).__init__(name=name, overlay=overlay,
@@ -1033,13 +1036,15 @@ class Choropleth(FeatureGroup):
                 geo_data,
                 topojson,
                 style_function=style_function,
-                smooth_factor=smooth_factor)
+                smooth_factor=smooth_factor,
+                tooltip=tooltip)
         else:
             self.geojson = GeoJson(
                 geo_data,
                 style_function=style_function,
                 smooth_factor=smooth_factor,
-                highlight_function=highlight_function if highlight else None)
+                highlight_function=highlight_function if highlight else None,
+                tooltip=tooltip)
 
         self.add_child(self.geojson)
         if self.color_scale:


Conengmo:
Hi @Gordonei, thanks for your PR. It looks good. 

I'm doubting a bit though. The `Choropleth` class is a bit weird, but used in a lot of first examples of folium. So we should not refactor it or delete it. But also I would like to keep it simple for new users, it is already a bit bloated.

I don't mind merging your PR. But I don't want the `Choropleth` class holding us back further developing geojson related features. If everything we do on geojson has to be added to `Choropleth` that may be problematic. But note that of course the `Choropleth` class is not off limits.

Let me know what you think about this.

jtbaker:
@Conengmo @Gordonei the problem with Choropleth class is that the properties or attributes from the pandas DataFrame do not persist into the `geo_data` GeoJson feature `properties` key - which is what the `GeoJsonTooltip` hooks into. And I think the 'data' attributes are probably the pieces of information that are most useful for the tooltip.

It would be great if we could do a GeoPandas GeoDataFrame merge of two separate dataframes to do all the styling, and have access to all data attributes directly from the GeoJson. I think this would involve adding `geopandas` as a dependency to do this data munging - but the Geopandas installation process is pretty fraught with issues already. 

If the `geo_data` object passed was a GeoDataFrame object originally, then the merge could be called as a class method from that. But it allows for static JSON files, text, etc to be passed as well.

I'm not sure of a clear path forward on more advanced functionality for this class without greatly increasing the dependencies and complexity of the codebase.

Conengmo:
I'm afraid I don't follow @jtbaker. Are you talking about having `GeoJsonTooltip` utilize the `data` argument of `Choropleth`? In this PR the `tooltip` argument to `Choropleth` is passed cleanly to `GeoJson`, so it shouldn't cause any trouble.

Or do you mean to alter this PR to have `Choropleth` create a `GeoJsonTooltip` object internally using the values in the `data` argument?

I see now how the docstring might cause confusion: `Can utilize the data` is about the `geo_data`, but might be interpreted as the `data` argument of `Choropleth`.

jtbaker:
> Are you talking about having `GeoJsonTooltip` utilize the `data` argument of `Choropleth`?

In a way yes - what I think would be most useful is to have the data values we are coloring the choropleth on the basis of, included in the `geo_data` GeoJson properties attributes.

here's an example of `Choropleth` usage: [link](https://python-graph-gallery.com/292-choropleth-map-with-folium/#map_2c50156153f44424b381b27fa6af1741)

a snippet of the GeoJson leaflet is working with:

```
{
    "features": [{
        "geometry": {
            "coordinates": [
                [
                    [-87.359296, 35.00118],
                    [-85.606675, 34.984749],
                    ...
                ]
            ],
            "type": "Polygon"
        },
        "id": "AL",
        "properties": {
            "highlight": {},
            "name": "Alabama",
            "style": {
                "color": "black",
                "fillColor": "#41ab5d",
                "fillOpacity": 0.7,
                "opacity": 0.2,
                "weight": 1
            }
        },
        "type": "Feature"
    }]
}
```

The style is determined and hardcoded into the `style` key python-side, but the `Unemployment` value column that the `style` is derived from is not available in the feature properties. But the precise underlying quantitative data for the state's Unemployment might be useful to have in a `GeoJsonTooltip`, and not just as a color on the map. Maybe [as a part of this IMO unnecessarily complex bit](https://github.com/python-visualization/folium/blob/master/folium/features.py#L991-L1009), other `properties` attributes could be set as well could be available for tooltip-ing, in addition to styling.

Gordonei:
Hi @Conengmo and @jtbaker , thanks for taking a look and your comments (and your work on this amazing library, it should go without saying!).

For my two cents, the Choropleth class has been a nice entry point into using folium - you can get something that looks really nice, straight off the bat. However, as soon as you start wanting to do anything more complicated, you end up decomposing it into separate GeoJSON, GeoJSON tooltips, styling functions, etc.

Hence, I agree that development effort is better spent on those more fundamental classes, and as much as possible, the Choropleth should wrap the functionality of those, with sensible defaults (realising it inherited from `FeatureGroup` was a big ah-ha moment for me). I think this change is in line with this school of thought.

Gordonei:
This being said, I am sympathetic to @jtbaker's point - if I understand correctly, you're suggesting we should be careful about how (2) is exposed in the class interface, otherwise, it's easy to get confused between what is in `data` vs the `geojson` attribute that results from (1).

Taking a step back, I think it makes sense to think of the class as providing two functions:
1. Joining data from the `data` structure to a spatial object, `geo_data`.
2. Putting that spatial object on the map, with assorted styling.

I'm tempted to make new, slightly more ambitious change, where (1) is made a bit more generic, and then define all of the styling (i.e. (2)) more strictly with respect to what is being joined in from `data`. The current behaviour could be kept as the default, as I think it makes for sensible defaults, but more of the styling could potentially be overridden.

Does this sound like it would be useful? Or am I going on a bit of tangent?

Conengmo:
I thought a bit more about it and I don't think we should merge this PR.

* It will make using the `Choropleth` class more complicated. 
* It may confuse people thinking it will use `data`. 
* More advanced users can easily add `GeoJsonTooltip` objects to `choropleth.geojson`.

But I'm open to PR's to improve the `Choropleth` class.
* Having easy foolproof tooltips created with the numeric data would be cool.
* @Gordonei I'm not sure what you're proposing exactly, but if you're inspired make a (WIP) PR and I'd be happy to discuss it.
* While we're at it closing #918 would be great :)

Gordonei:
Hokey, I think these are reasonable concerns - let me try organise my thoughts into a coherent proof of concept.

jtbaker:
Hey @Gordonei, I added a few commits to your PR in the `choroplethtooltip` branch on my fork. It uses the render method to test the underlying GeoJson data to see if a properties key is there, and if not, tries to add it from the dataframe. Let me know what you think.
@Conengmo I don't think taking this approach is too complex, and I think would be a pretty useful level of functionality for the end user.

https://github.com/jtbaker/folium/tree/choroplethtooltip

Gordonei:
Odd - let me see if I can fiddle with my permissions to let you do a PR against my fork, so it'll get pulled in here. 

Alternatively, I don't mind if you do a new PR against the main folium package - the commit history is being preserved :)

Gordonei:
@jtbaker, I've allowed maintainer to push commits to this PR, but not sure if you have write permissions or not. Failing that, I can add you as a collaborator on my fork?

jtbaker:
@Gordonei, I'm a contributor, but not a maintainer of the repo - and I guess GitHub will only let me PR to forks that have already had commits merged? (I'm still sort of a git newbie.) Can you merge my `choroplethtooltip` branch into your PR manually?

Also curious to see how @Conengmo feels about this.

PR #1075, opened 2019-02-22
fix pypi rendering

ocefpaf:
@Conengmo this should fix the bad rendering of the [README on PyPI](https://pypi.org/project/folium/).

I'm planning a new release soon. Do you have any PRs you would like to get merged before that?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,9 +43,6 @@ packages = [
     'folium.plugins'
 ]
 
-LICENSE = read('LICENSE.txt')
-long_description = '{}\n{}'.format(read('README.rst'), read('CHANGES.txt'))
-
 # Dependencies.
 with open('requirements.txt') as f:
     tests_require = f.readlines()
@@ -55,7 +52,9 @@ setup(
     name='folium',
     version=versioneer.get_version(),
     description='Make beautiful maps with Leaflet.js & Python',
-    long_description=long_description,
+    license="MIT",
+    long_description='{}'.format(read("README.rst")),
+    long_description_content_type="text/x-rst",
     author='Rob Story',
     author_email='wrobstory@gmail.com',
     url='https://github.com/python-visualization/folium',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -43,9 +43,6 @@ def walk_subpkg(name):
     'folium.plugins'
 ]
 
-LICENSE = read('LICENSE.txt')
-long_description = '{}\n{}'.format(read('README.rst'), read('CHANGES.txt'))
```
Conengmo: Could we still have the changelog following the readme?

Conengmo:
> I'm planning a new release soon. Do you have any PRs you would like to get merged before that?

Good plan! I'll update the changelog now first. I don't have any PR's that we have to wait for. I'm excited about adding the geojson markers and popups and refactoring the geojson styling, but no need to wait for that.

ocefpaf:
> I'm excited about adding the geojson markers and popups and refactoring the geojson styling, but no need to wait for that.

We can issue a new one as soon as those are merged. Just keep me posted if you need help reviewing/testing those.

setup.py
ocefpaf: I can re-add that. I thought it was too verbose and boring for PyPI but I don't have strong opinions about it.

Conengmo: Hmm yeah maybe. If someone wants to read the changelog they know where to find it I guess.

ocefpaf: The CHANGELOG is versioned so I guess that a link to the latest master version would make that easy to find and still keeping it simple. What do you think?

ocefpaf:
> Let's bake a release!

I'll start the processes as soon as this one is merged.

setup.py
Conengmo: That seems good.

ocefpaf:
Self-merging this one b/c it has simple README modifications.

PR #1084, opened 2019-03-08
Relaxing location data type restriction.

dskkato:
Currently, `folium.Map(location)` requires location data to be an instance of _list_ or _tuple_. I felt it to be inconvenient since neither numpy's array nor pandas' DataFrame cannot be used  as a _location_ argument directly. In this PR, I'm trying to relax the restriction so that any iterable object with `len(location)==2` can be accepted.

### DIFF
--- a/tests/test_folium.py
+++ b/tests/test_folium.py
@@ -19,6 +19,7 @@ from folium.features import GeoJson, Choropleth
 import jinja2
 from jinja2 import Environment, PackageLoader
 
+import numpy as np
 import pandas as pd
 
 import pytest
@@ -60,6 +61,17 @@ def test_get_templates():
 
     env = branca.utilities.get_templates()
     assert isinstance(env, jinja2.environment.Environment)
+
+
+def test_location_args():
+    """Test some data types for a location arg."""
+    location = np.array([45.5236, -122.6750])
+    m = folium.Map(location)
+    assert m.location == [45.5236, -122.6750]
+
+    df = pd.DataFrame({"location": [45.5236, -122.6750]})
+    m = folium.Map(df["location"])
+    assert m.location == [45.5236, -122.6750]
 
 
 class TestFolium(object):
@@ -108,9 +120,9 @@ class TestFolium(object):
                     'name': 'TileLayer',
                     'id': '00000000000000000000000000000000',
                     'children': {}
-                    }
                 }
             }
+        }
 
     def test_cloudmade(self):
         """Test cloudmade tiles and the API key."""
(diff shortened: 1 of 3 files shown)


ocefpaf:
You can ignore Appveyor b/c I just merged the PR that implements it. Or rebase against latest master to get the tests going on Windows. Also, the failing notebook test seems to be a network issue. However, your  change is not Python 2.7 compatible and we need a workaround there.

dskkato:
I saw Jinja2 has similar workroud. I will try and test on Python2.7.
```
try:
    from collections import abc
except ImportError:
    import collections as abc
```
https://github.com/pallets/jinja/blob/7e417c5c6ec90709f15e96a977ab59a5397a08fb/jinja2/_compat.py#L102-L105

ocefpaf:
> I saw Jinja2 has similar workroud. I will try and test on Python2.7.
> 
> ```
> try:
>     from collections import abc
> except ImportError:
>     import collections as abc
> ```
> [pallets/jinja:jinja2/_compat.py@`7e417c5`#L102-L105](https://github.com/pallets/jinja/blob/7e417c5c6ec90709f15e96a977ab59a5397a08fb/jinja2/_compat.py#L102-L105)


That is probably OK until we drop support for legacy python.

dskkato:
I would like to use [Collection](https://docs.python.org/3.7/library/collections.abc.html#collections.abc.Collection) for validation, but it's only supported 3.6 or later.  So I added one helper function for the compatibility.

tests/test_folium.py
```diff
@@ -197,7 +209,8 @@ def test_map_build(self):
 
         # Standard map.
         self.setup()
-        rendered = [line.strip() for line in self.m._parent.render().splitlines() if line.strip()]
+        rendered = [line.strip()
```
ocefpaf: Please avoid style changes in the same PR. It makes the reviewing process a little bit harder.
(Style changes are sometimes welcomed though, but preferably in a different PR.)

tests/test_folium.py
```diff
@@ -62,6 +63,17 @@ def test_get_templates():
     assert isinstance(env, jinja2.environment.Environment)
 
 
+def test_location_args():
+    """Test some data types for a location arg."""
+    location = np.array([45.5236, -122.6750])
+    m = folium.Map(location)
+    assert m.location == [45.5236, -122.6750]
+
+    df = pd.DataFrame({"location": [45.5236, -122.6750]})
```
ocefpaf: :+1:

tests/test_folium.py
dskkato: I reverted them. My formatter accidentally modified it...

ocefpaf:
This is good to go IMO but I'll let it here for a few days so @Conengmo can take a second look. Planning on merging it next week.

Thanks @dskkato!

ocefpaf:
Ooops. Wrong button! Sorry! Well... Now it is merge :smile:

Conengmo:
Looks good! Thanks for your contribution @dskkato and Filipe for the review.

ocefpaf:
And sorry again for the premature merge! I had way too many tabs open for a Saturday morning.

Conengmo:
No problem

PR #1086, opened 2019-03-09
Use jquery.com CDN instead of Google

### LINKED ISSUE #1085: jQuery google cdn cannot use in China. Can replace it with other cdn provider
We all know that google api cannot use in China, so the jQuery the folium reference is not available also. When using folium in China, it's so bad, can replace this with other cdn provider, I change the iQuery from google to code.jquery.com, the map can display normal.

Conengmo:
Closes #1085. Use jQuery's own CDN instead of Googles so folium is more easy to use in China.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -25,7 +25,7 @@ _default_js = [
     ('leaflet',
      'https://cdn.jsdelivr.net/npm/leaflet@1.4.0/dist/leaflet.js'),
     ('jquery',
-     'https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js'),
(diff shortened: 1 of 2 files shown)


Conengmo:
@ocefpaf any comments? Do you want to merge this?

ocefpaf:
> @ocefpaf any comments? Do you want to merge this?

LGTM. We just need to fix the failing test and let's merge it.

Conengmo:
Ah right, my bad. Didn't update the URL in the test template. Should be fixed now.

ocefpaf:
I'm planning a minor release next week with this change.

PR #1090, opened 2019-03-10
Explicit coordinate validation

### LINKED ISSUE #1089: unable to use FastMarkerCluster with anything more than lat/lng pairs in data
Hello,

Firstly, thanks for all you work on folium; I'm quite new to Python but have some experience with GIS and am really very impressed at how sophisticated folium is.

I'm having some trouble adapting [this example (at In [6] and [7])](https://nbviewer.jupyter.org/gist/anonymous/33f38b09ad3bfa277e2d9c06e4bb588c) from @JamesGardiner to my situation. Some example code below to illustrate the problem I'm having.

```python
import pandas as pd
import numpy as np
import folium
import folium.plugins as plugins

lats = np.asarray([-54.4, -54.5, -54.6])
longs = np.asarray([-69.5, -69.5, -69.5])
data = np.asarray([70, 90, 5])
color = np.asarray(['green', 'orange', 'red'])

dummy = [list(a) for a in zip(lats, longs)]
d1map = folium.Map(location = [-54.5, -69.5], zoom_start = 7)
plugins.FastMarkerCluster(dummy).add_to(d1map)
d1map

dummy2 = pd.DataFrame(dummy, columns=['lat', 'lng'])
d2map = folium.Map(location = [-54.5, -69.5], zoom_start = 7)
plugins.FastMarkerCluster(dummy2).add_to(d2map)
d2map  ## this gives me 'TypeError: must be real number, not str'

dummy3 = [list(a) for a in zip(lats, longs, data, color)]
dummy3 = pd.DataFrame(dummy3, columns=['lat', 'lng', 'data', 'color'])
d3map = folium.Map(location = [-54.5, -69.5], zoom_start = 7)
plugins.FastMarkerCluster(dummy3).add_to(d3map)
d3map  ## this gives me 'TypeError: must be real number, not str'

```
#### Problem description
I can combine a list of lat and lng and pass that to FastMarkerCluster and plot the points/clusters. However, converting to a dataframe then to JSON, as in the example notebook above, gives me `TypeError: must be real number, not str`. Also, if I combine anything more with the lat and lng, and pass this to FastMarkerCluster, I get a blank map.  Converting to a dataframe then converting to JSON gives me the same error as above.

#### Expected Output
I'd like to produce a map with FastMarkerCluster using callbacks to display popups as in the linked example above, and ideally also show different symbol colours based on the data.  I am unable to give FastMarkerCluster anything more than lat/lng pairs, however, preventing me from using the remaining data in a callback.

FYI I can seemingly do this with no problems using MarkerCluster (see my example 1 in SO question [here](https://stackoverflow.com/questions/55082227/use-customized-markers-in-fastmarkercluster-in-python-folium)), but my data has several thousand rows so that would be infeasible.

0.8.2

Many thanks in advance,
-Rameses

Conengmo:
**Problem**
We have multiple conversion/validation functions in utilities. It's hard to decide which to use, and they are not thorough. 

**Solution**
Have a few validation/conversion functions that are clear and explicit. They raise meaningful errors. They have extensive tests.

So that's what I tried in this PR. I put some things together into two functions: one for single coordinate pairs and one for lists (of lists) of coordinate pairs. Both try to raise meaningful exceptions and convert to a single common format.

I applied the validation functions throughout the code base when possible. In some cases (`HeatMap`, `FastMarkerCluster`) I made custom ones because they have special data formats.

For the Marker class and vector classes I added a new BaseMarker class to make the distinction between single and multiple locations explicit. Before the vector classes just passed a list of coordinate pairs to `Marker`, while `Marker` is documented to only accept a single coordinate pair. 

Possible improvements:
* Use tuple data type for coordinate pairs. That's something for another PR I think, if we want that at all.

@ocefpaf @jtbaker want to review this one?

Closes #1089


folium/map.py
```diff
@@ -205,14 +210,53 @@ def __init__(self, color='blue', icon_color='white', icon='info-sign',
...
+
+    Not for direct consumption.
+
+    Note that self.location can be either a single coordinate pair or
+    a list of coordinate pairs.
+
+    """
+    _template = None
+
+    def __init__(self, location=None, locations=None,
```
ocefpaf: While I don't like the possible confusion that `location` vs `locations`, this is better than the current state of things. (And I believe we are not encouraging users to use the `BaseMarker`.)

folium/utilities.py
```diff
@@ -12,62 +12,105 @@
...
 from six.moves.urllib.parse import urlparse, uses_netloc, uses_params, uses_relative
 
 
 _VALID_URLS = set(uses_relative + uses_netloc + uses_params)
 _VALID_URLS.discard('')
 
 
-def _is_sized_iterable(arg):
-    """Validates the arg is Sized & Iterable"""
-    return isinstance(arg, abc.Sized) & isinstance(arg, abc.Iterable)
```
ocefpaf: Don't you think this was simpler with the `abc`? I'm fine with this approach but I could not see why the change at a first pass.

Conengmo: Hmm I tried to change it into abc.Sequence, because it should also have `__getitem__`, but then it won't work for Numpy arrays. I don't know what other data types have a length but don't inherit from `abc.Sized` for example. I guess I prefer the try/except version because it's more explicit and handles duck types better than abc.

folium/map.py
Conengmo: You're right, the possible confusion is not okay. I'll try something to split it off better.

Conengmo: I made a new `BaseMultiLocation` class for the vector layers. It doesn't inherit from Marker, so no confusion about `self.location` or `self.locations`.

folium/plugins/fast_marker_cluster.py
```diff
@@ -69,15 +70,15 @@ def __init__(self, data, callback=None, options=None,
                                                 control=control, show=show,
                                                 **kwargs)
         self._name = 'FastMarkerCluster'
-        self._data = _validate_coordinates(data)
+        self.data = [[*validate_location(row[:2]), *row[2:]] for row in data]
```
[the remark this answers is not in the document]
Conengmo: We dropped Python 2.7, this works on Python 3.5 and higher.

Conengmo:
Hi @ocefpaf, I think this is pretty much ready. Could you check it out if you have a moment?

ocefpaf:
Awesome changes @Conengmo! Thanks for fixing this!!

Conengmo:
Thanks for the review!

PR #1091, opened 2019-03-12
Adds PolyLineOffset Plugin

FabeG:
See https://github.com/bbecquet/Leaflet.PolylineOffset


folium/plugins/polyline_offset.py
```diff
@@ -0,0 +1,70 @@
...
+    >>> map = folium.Map(location=[58.0, -11.0],
+    ...                  zoom_start=4,
+    ...                  tiles="Mapbox Bright")
+    >>> ll = [ [58.44773, -28.65234], [53, -23.33496]]
+    >>> plugins.PolyLineOffset(ll, color="#f00", opacity=1, offset=-5).add_to(map)
+    >>> plugins.PolyLineOffset(ll, color="#080", opacity=1, offset=10).add_to(map)
+
+    """
+
+    def __init__(self, locations, popup=None, tooltip=None, **kwargs):
```
Conengmo: If people use this class, I'm assuming they want to use the `offset` parameter. Can you add it to the arguments explicitly?

folium/plugins/polyline_offset.py
```diff
@@ -0,0 +1,70 @@
...
+    offset: int, default 0
+        Relative pixel offset to draw a line parallel to an existant one,
+        at a fixed distance.
+    **kwargs:
+        Polyline options. See their Github page for the
+        available parameters.
+
+    See https://github.com/bbecquet/Leaflet.PolylineOffset
+
+    Examples
```
Conengmo: Good that you include example. They can be shorter though I think, no need to include the `folium.Map` stuff.

Maybe:
`>>> PolyLineOffset([[58, -28], [53, -23]], color='#f00', offset=-5).add_to(m)`

(In general we don't use `map` as variable name, because that interferes with Pythons `map` operator. Convention is to use `m` instead.)

folium/plugins/polyline_offset.py
```diff
@@ -0,0 +1,70 @@
...
+    ...                  tiles="Mapbox Bright")
+    >>> ll = [ [58.44773, -28.65234], [53, -23.33496]]
+    >>> plugins.PolyLineOffset(ll, color="#f00", opacity=1, offset=-5).add_to(map)
+    >>> plugins.PolyLineOffset(ll, color="#080", opacity=1, offset=10).add_to(map)
+
+    """
+
+    def __init__(self, locations, popup=None, tooltip=None, **kwargs):
+        super(PolyLineOffset, self).__init__(
+            locations=locations, popup=popup, tooltip=tooltip
```
Conengmo: I'm hesitating whether we should not let the option handling be done by `PolyLine`. And then only add the `offset` argument in this class. I would like this plugin not copy code from `PolyLine`.

But then you would have to load the json, add the offset, and dump it again. Not doing `json.dumps` until render time is something I want to tackle in another PR.

tests/plugins/test_polyline_offset.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+                var {{this.get_name()}} = L.polyline(
+                    {{this.location}},
+                    {{ this.options }}
+                    )
+                    .addTo({{this._parent.get_name()}});
+            """
+    )  # noqa
+
+    expected_rendered = tmpl.render(this=polylineoffset)
```
Conengmo: Cool that you added a test! I'm not sure if it's verifying that an offset was added though? If offset is not in the options, will the test fail?

folium/plugins/polyline_offset.py
```diff
@@ -0,0 +1,70 @@
...
+
+from branca.element import JavascriptLink
+
+from folium.vector_layers import path_options
+from folium.vector_layers import PolyLine
+
+
+class PolyLineOffset(PolyLine):
+    """
+    Class adding offset capabilities to the PolyLine class.
```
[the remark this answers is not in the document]
FabeG: Hi @Conengmo,  thanks for the review !
- I uncommented the commented values (to have the same example as the one shown in the leaflet plugin),
- I renamed the `ll` variable into `coords`,
- I changed the unpythonic code you highlighted
- And corrected the docstring

folium/plugins/polyline_offset.py
FabeG: Ok, added the `offset `parameter.

folium/plugins/polyline_offset.py
FabeG: - Renamed the `map `into `m`
- Reduce the number of lines of the example

folium/plugins/polyline_offset.py
FabeG: Taking advantage of your last PR (#1101, using the `jinja2 tojson `filter) avoiding the `json.dumps`, I reduced a little the code. Let me know if it seems better for you

tests/plugins/test_polyline_offset.py
FabeG: I added a test for this specific case (PolyLineOffset without offset in the options: in such case, offset takes its default value (0) and looks like a normal PolyLine).

FabeG:
The failed CI build seems to come from the `WmsTimeDimension `notebook, not related to this PR then.

folium/plugins/polyline_offset.py
Conengmo: Very nice :)

tests/plugins/test_polyline_offset.py
Conengmo: 👍

Conengmo:
This is good stuff @FabeG! A very neat plugin implementation with a great example notebook. Indeed the Travis build failed on something unrelated, I'll restart it and see if we can get it working. After that this is good to go!

FabeG:
Thanks to your nice comments and your review @Conengmo !
Sorry if my question seems strange (first PR on github), but am I supposed to do something now ?

Conengmo:
No, I was waiting for the tests to finish and then forgot about this :p But the checks have passed so let's merge this thing!

Conengmo:
:rocket:

PR #1094, opened 2019-03-14
First draft of new custom pane functionality.

### LINKED ISSUE #1088: Support for custom map panes
Hi,

Is there currently any way in folium to work with panes as described here?

[https://leafletjs.com/examples/map-panes/](https://leafletjs.com/examples/map-panes/)

I've searched around, but didn't find anything. Same as in the link, my use-case is adding a label-only tile layer above a choropleth, i.e. this

folium_after

instead of this

folium_before

Right now I have a hacky python script that edits the html file after saving, but it would be great to be able to do this within folium.

API could be like

```python
m.create_pane(name="labels", z_index=625, pointer_events=False)

labels = folium.TileLayer(
    name="extra-labels",
    tiles="https://stamen-tiles-{s}.a.ssl.fastly.net/toner-labels/{z}/{x}/{y}{r}.png",
    attr='Map tiles by <a href="http://stamen.com">Stamen Design</a>, <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a> &mdash; Map data &copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
    overlay=True,
    show=True,
    pane="labels",  # This already gets passed through
).add_to(m)
```

If this sounds possible, I'd be happy to take a crack at implementing it.

Cheers,
Liam

lpkirwin:
This is a rough sketch to agree the API before I flesh it out and add docs. Right now I see two options for the API:

1.  Add custom pane object to map, add kwarg to other layers:

```python
CustomPane("name", ...).add_to(m)
OtherLayer(..., pane="name).add_to(m)
```

- Closest to [leaflet API](https://leafletjs.com/examples/map-panes/)
- Would require 'pane' argument or **kwargs for layers (raster layers already pass kwargs, and leaflet layers [all take 'pane' as an option](https://leafletjs.com/reference-1.4.0.html)
- I haven't changed any other class yet, but the example in the notebook works because TileLayer does pass kwargs on to leaflet

2. Add custom pane object to map, add other layers as child of pane:

```python
p = CustomPane(...).add_to(m)
OtherLayer(...).add_to(p)
```

- Don't have to mess around with extra arguments to the other layers
- Further from the leaflet API, but perhaps (?) closer to folium?

What I have so far is the basis for 1. Happy to flip to 2 if that's preferred.

Closes #1088


lpkirwin:
#1088

Conengmo:
Hi @lpkirwin, thanks for taking this up. Good sketch you made. It is clear to me how option 1 would work out.

I have a preference for option 2 because:
* it's more like folium
* we can do more checks on the Python side and raise exceptions there

But I'm not sure if we can implement that in a sensible way. [Leaflet has no methods to move an object to another pane](https://gis.stackexchange.com/questions/231416/change-pane-of-existing-tilelayer-in-leaflet-interactively), it has to be done on creation as far as I can tell. So we would have to rummage about in the templates or options of folium objects anyway.

I'd be interested to hear your ideas about how we could implement option 2.

If that seems not workable, option 1 is also a fine option. Perhaps we can find another way to raise meaningful exception on the Python side when users implement this functionality wrong.

lpkirwin:
Hi @Conengmo, no problem.

Added a new commit with an idea for option 2. It follows the second API above, adding other layers as children to the custom pane. I've added a render argument to the CustomPane class that

- Adds the 'pane' option to each child's options dict
- Switches the parent of each child to the pane's parent (i.e. the map)

and then calls super(). Works fine with TileLayers, e.g. [here](https://github.com/python-visualization/folium/blob/84e4bd2390275e155b9ef9c15c4cd36ff5ccde16/examples/CustomPanes.ipynb). I can see not all folium classes have options, so not sure how well this would extend past TileLayers (and unsure how many use-cases there are outside TileLayers...)

What do you think?

folium/map.py
```diff
@@ -465,3 +465,35 @@ def __init__(self, bounds, padding_top_left=None,
...
+        self._name = 'Pane'
+        self.z_index = str(z_index)
+        self.pointer_events = pointer_events
+
+    def render(self, **kwargs):
+        for name, element in self._children.items():
+            options = json.loads(element.options)
+            options.update({'pane': self.get_name()})
+            element.options = json.dumps(options, sort_keys=True, indent=8)
+            element._parent = self._parent
```
Conengmo: When you move the element up the tree it will not get rendered, because the parent of `CustomPane` has already been rendered...

Conengmo:
Thanks for the quick test. I added a comment about a possible problem with the child elements not being rendered. That's probably fixable by calling render manually. In general I think it best to avoid messing with the object tree. But sometimes it has to be done to get cool features :)

What do you think is the best approach of these two?

folium/map.py
lpkirwin: It rendered OK in my test notebook, see screenshot below. I haven't properly moved the element up the tree -- just swapped out the parent, but the tile layer still exists in the pane's children list.

image

lpkirwin:
I like the API better on the second, but same as you I worry that hacking around in the object tree could have some unintended consequences, e.g. when interacting with other objects or features that might not expect this behavior. After calling render() on the pane, there's some inconsistent state (element has the map as a _parent, but is in the pane's _children and not the map's _children).

But I don't have other concrete ideas at the moment. Maybe it would be possible for the tile layer to check at render time whether its parent was a map or a pane, and then use its parent/grandparent name as appropriate.

folium/map.py
Conengmo: Ah right, it's in the list of children so it will get rendered. Good one.

lpkirwin: Changed the parent just so that, when render is called on the layer, the javascript ends up as
```js
element().add_to(map_name)
```
rather than
```js
element().add_to(pane_name)
```
which leaflet wouldn't understand

Conengmo:
Hmm given your exploration I now lean towards option 1. It seems less fragile than mucking around with the object tree.

I'm currently working on making sure each object passes `**kwargs` as options, so that should work together nicely.

folium/map.py
Conengmo: That makes sense, thanks for explaining.

lpkirwin:
Sounds good, and I agree. I'll revert back to option 1 and flesh out the documentation

lpkirwin:
Hi @Conengmo, have switched back to the earlier implementation and added a docstring and a bit more to the notebook. Couple questions for you:

- Should I do anything about kwargs, e.g. adding 'pane' as a specific kwarg in other classes, or adding **kwargs to classes that don't accept them? You mentioned that this was being developed elsewhere, so if so I can hold off.
- A straightforward test to add would be the equivalent of test_popup_show() in test_map.py. Do you think any other tests would be helpful?

folium/map.py
```diff
@@ -480,3 +480,46 @@ def __init__(self, bounds, padding_top_left=None,
...
+        Name of the custom pane. Other map elements can be added
+        to the pane by specifying the 'pane' kwarg when constructing
+        them.
+    z_index: int or string, default 625
+        The z-index that will be associated with the pane, and will
+        determine which map elements lie over/under it. The default
+        (625) corresponds to between markers and tooltips. Default
+        panes and z-indexes can be found at
+        https://github.com/Leaflet/Leaflet/blob/v1.0.0/dist/leaflet.css#L87
+    pointer_events: bool, default False
```
Conengmo: What are the consequences of setting the default here to `False`?

Maybe good to add a small section in the example notebook on how this option can be used.

folium/map.py
```diff
@@ -480,3 +480,46 @@ def __init__(self, bounds, padding_top_left=None,
...
+    name: string
+        Name of the custom pane. Other map elements can be added
+        to the pane by specifying the 'pane' kwarg when constructing
+        them.
+    z_index: int or string, default 625
+        The z-index that will be associated with the pane, and will
+        determine which map elements lie over/under it. The default
+        (625) corresponds to between markers and tooltips. Default
+        panes and z-indexes can be found at
+        https://github.com/Leaflet/Leaflet/blob/v1.0.0/dist/leaflet.css#L87
```
Conengmo: Could you link here to the Leaflet documentation instead? https://leafletjs.com/reference-1.4.0.html#map-pane
It has the same info, is easier to read, and using this URL makes it easy for us to update all links when a new Leaflet version is released.

folium/map.py
```diff
@@ -480,3 +480,46 @@ def __init__(self, bounds, padding_top_left=None,
...
+    pointer_events: bool, default False
+        Whether or not layers in the pane should interact with the
+        cursor. Setting to False will prevent interfering with
+        pointer events associated with lower layers.
+    """
+    _template = Template(u"""
+        {% macro script(this, kwargs) %}
+            {{this._parent.get_name()}}.createPane('{{this.name}}');
+            {{this._parent.get_name()}}.getPane(
+                '{{this.name}}').style.zIndex = {{this.z_index}};
```
Conengmo: You could use `tojson` here also instead of converting to str in the init... But this also works.

folium/map.py
lpkirwin: For my particular use case (labels on top of choropleth), setting this to False ensures that the new layer won't mask mouse events from lower layers, e.g. if you want a tooltip to pop up when you hover over a choropleth.

Right now I'm struggling to imagine a case where you'd want this to be True! But

Conengmo: Thanks for explaining, sounds like `False` is a good default. I agree on leaving it in.

Conengmo:
Just had a look at your changes and I think this is ready!  The code looks nice, you wrote a test, the example notebook is good.

We fixed the issue with the notebook tests on Travis before, so I took the liberty of updating your branch from master. When the tests are passed I'll go ahead and merge this.

lpkirwin:
> Just had a look at your changes and I think this is ready! The code looks nice, you wrote a test, the example notebook is good.
> 
> We fixed the issue with the notebook tests on Travis before, so I took the liberty of updating your branch from master. When the tests are passed I'll go ahead and merge this.

Thanks Frank!

jmaralc:
Great feature!! thanks a lot 👍

PR #1099, opened 2019-03-16
Icon: warn for wrong color value

Conengmo:
The `Icon` class has two color arguments: one for the marker, one for the icon. The color for the icon can be a color name or any hex color. The color for the marker must be one of a set of values, otherwise the color will default to red.

This PR adds a check on the color argument a user provides. If it is not in the set of allowed colors, a warning is raised. 

I used a warning instead of an exception because even with a wrong color value the map will render normally and it will display without errors. The color will just not be as expected.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -10,6 +10,8 @@ from __future__ import (absolute_import, division, print_function)
 import json
 from collections import OrderedDict
 
+import warnings
+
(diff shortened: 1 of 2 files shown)


Conengmo:
@ocefpaf what do you think, useful or unnecessary?

ocefpaf:
> @ocefpaf what do you think, useful or unnecessary?

It is something nice to have. Let's merge it!

Conengmo:
Thanks!

PR #1100, opened 2019-03-16
Prevent installing 0.9 on dropped Python versions

Conengmo:
* Use pip's `python_requires` argument should work for users with fairly recent pip version.
* Otherwise raise an exception in `setup.py` when it is executed by an old Python version.
* As a final catch raise an exception on importing folium when the Python version is lower than 3.

This is from the guidelines from https://python3statement.org/practicalities/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ env:
 matrix:
   fast_finish: true
   include:
-  - name: legacy-python
-    env: PY=2.7
   - name: default
     env: PY=3.7
   - name: notebooks-conding-standard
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -66,11 +64,6 @@ script:
       pytest /tmp -vv --ignore=tests/notebooks/test_notebooks.py ;
     fi
 
-  - if [[ $TRAVIS_JOB_NAME == 'legacy-python' ]]; then
```
ocefpaf: Glad to see this go away!

ocefpaf:
Thanks @Conengmo! Those safeguards against old `pip` are nice to have.

PR #1101, opened 2019-03-16
Use Jinja2 `tojson` filter

Conengmo:
Don't `json.dumps` options and data, but use Jinja2's `tojson` filter to convert Python variables to JS compatible formats. This works for strings, numbers, bools, lists and dictionaries. I added a test file to verify this behavior.

Normalize the behavior of options. In `__init__`, put options in a `self.options` dictionary using a new `parse_options()` function. Also allow `**kwargs` whenever it makes sense.

This combination: consistent option handling and using `tojson` makes our code and templates more clean and consistent. Allowing `**kwargs` gives users more flexibility.

I also did refactoring of templates whenever I touched them. Have consistent indentation and style.

Fixed the tests for the changes. Consistently use the `normalize` function to normalize rendered output.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -25,9 +25,10 @@ from folium.utilities import (
     image_to_url,
     none_max,
     none_min,
-    get_obj_in_upper_tree
+    get_obj_in_upper_tree,
+    parse_options,
 )
-from folium.vector_layers import PolyLine
+from folium.vector_layers import PolyLine, path_options
 
 from jinja2 import Template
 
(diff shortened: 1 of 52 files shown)


folium/features.py
```diff
@@ -46,65 +47,44 @@ class RegularPolygonMarker(Marker):
...
-                    numberOfSides: {{this.number_of_sides}},
-                    rotation: {{this.rotation}},
-                    radius: {{this.radius}}
-                    }
-                ).addTo({{this._parent.get_name()}});
-            {% endmacro %}
-            """)
+        {% macro script(this, kwargs) %}
+            var {{ this.get_name() }} = new L.RegularPolygonMarker(
+                {{ this.location|tojson }},
```
ocefpaf: Way cleaner! I like this!!

ocefpaf:
@Conengmo this is failing legacy Python and the notebook tests. The former we no longer care, #1100 was already merged BTW, but the latter seems to be related to the changes here, right?

Conengmo:
I’ll have a look and fix the notebooks. After rebasing Python 2 should no longer be a problem :)

This PR touches quite a lot of code. Do you see any problems with this?

ocefpaf:
> I’ll have a look and fix the notebooks. After rebasing Python 2 should no longer be a problem :)
> 
> This PR touches quite a lot of code. Do you see any problems with this?

At a first glance, no. Even though it does touch a lot of the code base it is virtually the same changes everywhere, so let's trust our tests and merge this as soon as you rebase it and the tests are passing.

Conengmo:
> let's trust our tests

Good point! During the work for this PR I looked at the tests more, and so I got thinking about want to look into having better functional tests.

Conengmo:
@ocefpaf seems the Travis build is failing on the installation step. It's hard to debug because it sees the whole block as one command:

```
The command "wget http://bit.ly/miniconda -O miniconda.sh
  bash miniconda.sh -b -p $HOME/miniconda
  export PATH="$HOME/miniconda/bin:$PATH"
  conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
  conda update conda --quiet
  conda config --add channels conda-forge --force
  conda install pycryptosat
  conda config --set channel_priority strict
  conda config --set safety_checks disabled
  conda create --name TEST python=$PY --file requirements.txt --file requirements-dev.txt
  source activate TEST
  " failed and exited with 1 during .
```

ocefpaf:
> @ocefpaf seems the Travis build is failing on the installation step. It started from the last merged PR. It's hard to debug because it sees the whole block as one command:

investigating it. Hold on...

ocefpaf:
@Conengmo the CIs are back online and the failure there is real ;-p

Conengmo:
Thanks for bringing the CI back! I see the test is indeed failing. It's due to the Webmapservice that's used in the WMS example notebook. This has happened before, now it seems to give a HTTP 503 instead of a timeout. I'll see if we can use another URL...

ocefpaf:
> Thanks for bringing the CI back! I see the test is indeed failing. It's due to the Webmapservice that's used in the WMS example notebook. This has happened before, now it seems to give a HTTP 503 instead of a timeout. I'll see if we can use another URL...

Or we can just add an `xfail` (expected failure) there. I'll see if I can do that later today. It is unrelated to this PRs anyway so let's get this merged!

PR #1111, opened 2019-03-23
Remove future imports and Six

Conengmo:
Remove lines with `from __future__ import`.

Remove the use of the `six` library.

I'm assuming nobody will pass `byte` objects instead of `str`.


ocefpaf:
Sweet!

PR #1113, opened 2019-03-27
Fix typo in Py2 import error

Conengmo:
Fix a tiny typo in `__init__.py`.

### DIFF
--- a/folium/__init__.py
+++ b/folium/__init__.py
@@ -67,7 +67,7 @@ if sys.version_info < (3, 0):
     
     - Install an older version of folium:
     
-     $ pip install 'folium<9.0'
+     $ pip install 'folium<0.9.0'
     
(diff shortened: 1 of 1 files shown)


folium/__init__.py
```diff
@@ -67,7 +67,7 @@
     
     - Install an older version of folium:
     
-     $ pip install 'folium<9.0'
+     $ pip install 'folium<0.9.0'
```
ocefpaf: We actually should be at version `9.0` by now :smile: 

Maybe we can make the Python 3 only our 1.0!

Conengmo: I was thinking about what we’d need for a 1.0 version the other day, let’s talk about that some time.

ocefpaf:
Safe to merge without waiting for the notebook-coding-standards test.

PR #1120, opened 2019-04-07
Remove extra ( " ) on Draw plugin template

fullonic:
On master branch there is an extra ' " ' on line 79 of Draw plugin.
https://github.com/python-visualization/folium/blob/913cf298241c227c71c9a8827d17f919c1abdb65/folium/plugins/draw.py#L79

When the plugin is used, the map isn't displayed and produces a JS syntax error.
`"SyntaxError: "" string literal contains an unescaped line break"`

This PR only removes that extra ( " ).


Conengmo:
The Travis build failure is unrelated, so I'll go ahead and merge this.

Conengmo:
Thanks for fixing this bug @fullonic!

PR #1121, opened 2019-04-07
Use Xenial on Travis to fix nbval on Py3.7

Conengmo:
Nbval is a package that tests our notebooks. To make it work on Travis we should either use a newer Ubuntu version than the default, or test on Py 3.6 instead of 3.7.

In this change I'm trying running all Travis tests on Ubuntu Xenial instead of the default Trusty.

https://docs.travis-ci.com/user/reference/overview/#linux


Conengmo:
This doesn't seem to fix the issue, and breaks other tests...

PR #1122, opened 2019-04-07
Require Jinja2 version with tojson

Conengmo:
The `tojson` filter was introduced in version 2.9. folium won't work with lower versions.


Conengmo:
The build fails on unrelated notebook code tests. Given that this is a very minor change I'll go ahead and merge this.

PR #1123, opened 2019-04-07
Use Py 3.6 for Travis notebook tests

Conengmo:


Conengmo:
This also doesn't fix the problem...

PR #1127, opened 2019-04-07
Clean-up Map

Conengmo:
This PR contains changes to simplify the `Map` class.

I removed the `add_tile_layer()` method, it seems like a relic from when folium didn't work with classes yet. The `TileLayer` should be used directly instead.

`Map` currently contains a lot of arguments, some of which are duplicate with `TileLayer`. I removed some of the more advanced ones, such as `API_key`, `max_native_zoom`, `detect_retina`, `subdomains` and `nowrap`. Note that these are all still available in the `TileLayer` class, so if someone needs them they can use the `TileLayer` class directly.

```
m = folium.Map(tiles=None)  # I want to do complicated tiles stuff
folium.TileLayer(**super_complex_options).add_to(m)
```

I also made the options parsing a bit simpler. Use the `parse_options` method to create a dictionary suitable for unpacking in the template. Don't define arrays in the template. Less code, less trouble.


ocefpaf:
I'm happy with the changes here but we will need to re-work the tests to get that to pass.

Conengmo:
> I'm happy with the changes here but we will need to re-work the tests to get that to pass.

Good to hear! I'm working on getting the tests working, but my environment got F-ed up so I need to fix that first :p

Conengmo:
I removed the `test_map_build` test that uses the `fol_template.html` file. That test has caused me more trouble than that it has been helpful. Copy-pasting the template from a class to a test is not a useful way of writing tests.

ocefpaf:
> I removed the `test_map_build` test that uses the `fol_template.html` file. That test has caused me more trouble than that it has been helpful. Copy-pasting the template from a class to a test is not a useful way of writing tests.

I agree. That outlived its usefulness a long time ago. Thanks!

Conengmo:
Thanks for the review and the merge Filipe, I appreciate it. 

Reading back my last comment I must say I was a bit frustrated after an hour of trying to fix DLL problems with Anaconda on Windows :p I don’t want to be disrespectful of previous work. Instead it’s really cool to see what the work of multiple different people over multiple years has led to.

ocefpaf:
> Reading back my last comment I must say I was a bit frustrated after an hour of trying to fix DLL problems with Anaconda on Windows :p I don’t want to be disrespectful of previous work. Instead it’s really cool to see what the work of multiple different people over multiple years has led to.

Sorry about that. The only reason to use conda via minconda to install packages (not Anaconda b/c that is the full distribution and kind of heavy-weight for us) is to install geopandas (b/c of its dependency on `gdal`).

The situation on Windows got better now that you can use `strict` channel. That will be default in the next conda release (4.7) BTW. [These docs does not mention Windows](https://conda-forge.org/docs/user/tipsandtricks.html) exactly but it is probably the same problem. TL;DR feel free to bug me with question before wasting your time with this. I have some experience with conda package, specially those from conda-forge.

PR #1128, opened 2019-04-08
Fix broken attribution for built-in tiles.

### LINKED ISSUE #1117: Map attribution not showing up for built in tiles
In case the default tile is chosen (OpenStreetMap) during the creation of the map (or other built-in tiles, the ones present in the template/tiles folder), no attribution is showing up on the bottom left of maps.

```python

import folium
m = folium.Map(location=[45.5236, -122.6750])
```

Looking into the code, it seems that the file `attr.txt` for each built-in tile is well read, but nothing is done with it, the template does't refer to attr and the options is not updated:

Here is the template:
```python
_template = Template(u"""
            var {{ this.get_name() }} = L.tileLayer(
                {{ this.tiles|tojson }},
                {{ this.options|tojson }}
            ).addTo({{ this._parent.get_name() }});
        {% endmacro %}
        """)
```
and an extract of the code (`raster_layers.py`) that is supposed to set the `self.attr` attribute:

 ```python
attr_template = 'tiles/' + tiles_flat + '/attr.txt'
if tile_template in templates and attr_template in templates:
        self.tiles = self._env.get_template(tile_template).render(API_key=API_key)  # noqa
        self.attr = self._env.get_template(attr_template).render()
    else:
        self.tiles = tiles
        if not attr:
            raise ValueError('Custom tiles must have an attribution.')
self.attr = attr
```

Am I missing something ?

### PR DESCRIPTION
**Pull request by FabeG**
Small correction to fix broken attribution for built-in tiles (Closes #1117).

- `self.attr` (from `Map`) has been removed (`attribution` now set with `options`)
- updated tests

Maybe the attribution files (`attr.txt`) should be reviewed and updated too ?

### DIFF
--- a/tests/test_folium.py
+++ b/tests/test_folium.py
@@ -133,10 +133,10 @@ class TestFolium(object):
         for tiles in default_tiles:
             m = folium.Map(location=[45.5236, -122.6750], tiles=tiles)
             tiles = "".join(tiles.lower().strip().split())
-            url = f"tiles/{tiles}/tiles.txt"
-            attr = f"tiles/{tiles}/attr.txt"
-            url = m._env.get_template(url).render()
-            attr = m._env.get_template(attr).render()
+            url = "tiles/{}/tiles.txt".format
+            attr = "tiles/{}/attr.txt".format
+            url = m._env.get_template(url(tiles)).render()
+            attr = m._env.get_template(attr(tiles)).render()
 
             assert m._children[tiles].tiles == url
             assert htmlsafe_json_dumps(attr) in m._parent.render()
--- a/tests/test_folium.py
+++ b/tests/test_folium.py
@@ -13,10 +13,10 @@ import branca.element
 
 import folium
 from folium.features import GeoJson, Choropleth
-from folium.utilities import normalize
 
 import jinja2
 from jinja2 import Environment, PackageLoader
+from jinja2.utils import htmlsafe_json_dumps
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: tests/test_folium.py [hunk 1]
```diff
@@ -114,40 +114,48 @@ def test_init(self):
...
+        ]
         for tiles in default_tiles:
             m = folium.Map(location=[45.5236, -122.6750], tiles=tiles)
-            tiles = ''.join(tiles.lower().strip().split())
-            url = 'tiles/{}/tiles.txt'.format
-            attr = 'tiles/{}/attr.txt'.format
-            url = m._env.get_template(url(tiles)).render()
-            attr = m._env.get_template(attr(tiles)).render()
+            tiles = "".join(tiles.lower().strip().split())
+            url = f"tiles/{tiles}/tiles.txt"
```
Conengmo: f-strings are great, but folium still supports Python 3.5. Right now we don't run tests for Python 3.5, but that's something we should probably do. Can you use normal `format()` instead?

FabeG: Ok @Conengmo , sorry for that, it is corrected now. 
One question: I wanted to add also a test for tiles needing an API key (cloudmade and mapbox), but I noticed that you suppressed one in your previous commit (`test_cloudmade`). Any reason for that ?

FabeG: @Conengmo Ah, I see, you removed the test `test_cloudmade` following GH-1127.

FabeG:
It seems the previous commit failed, but not sure it comes from the modifcations I made.

Conengmo:
Thanks for fixing this @FabeG! It's appreciated.

PR #1133, opened 2019-04-25
Fix broken prefer_canvas option

### LINKED ISSUE #1131: Performance issue when trying to draw lots of markers: prefer_canvas option inefficient
Trying to plot thousands of circleMarkers, it appeared that rendering the map was very slow.

I tried to use the `prefer_canvas` option during Map creation:
```python

carte = folium.Map(
        location=[48, 8.33],
        tiles='Mapbox Bright',
        prefer_canvas=True,
    )

```
but didn't see any performance improvement. Looking into the generated html, it seems that the  `prefer_canvas` option is rendered as a Global Switch:

```javascript
<script>
            L_PREFER_CANVAS = true;
            L_NO_TOUCH = false;
            L_DISABLE_3D = false;
</script>
```

Unfortunately, it seems that the `L_PREFER_CANVAS` switch isn't supported anymore since Leaflet 1.0, the prefered way of using Canvas is now during the map creation (https://leafletjs.com/reference-1.4.0.html#map-prefercanvas).

Adding directly the ```preferCanvas: true``` option in the html, I saw a huge speed improvement in the drawing of the markers.

FabeG:
Small correction to fix broken prefer_canvas option (Close #1131):
since leaflet 1.0.0, the preferCanvas option is not part of the GlobalSwitch
options, but has to be passed as an option to L.Map.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -55,16 +55,14 @@ class GlobalSwitches(Element):
 
     _template = Template("""
         <script>
-            L_PREFER_CANVAS = {{ this.prefer_canvas|tojson }};
             L_NO_TOUCH = {{ this.no_touch |tojson}};
             L_DISABLE_3D = {{ this.disable_3d|tojson }};
         </script>
     """)
 
-    def __init__(self, prefer_canvas=False, no_touch=False, disable_3d=False):
+    def __init__(self, no_touch=False, disable_3d=False):
         super(GlobalSwitches, self).__init__()
         self._name = 'GlobalSwitches'
-        self.prefer_canvas = prefer_canvas
         self.no_touch = no_touch
         self.disable_3d = disable_3d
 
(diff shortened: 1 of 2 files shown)


folium/folium.py
```diff
@@ -124,11 +122,6 @@ class Map(MacroElement):
         (going from bottom to top).
     control_scale : bool, default False
         Whether to add a control scale on the map.
-    prefer_canvas : bool, default False
```
Conengmo: Why do you remove this?

Conengmo:
This looks good @fabeg, I’ll merge it later this week.

Can I ask, maybe you know, why wouldn’t we always use prefer_canvas? If it increases performance we could set the default to true. Or are there downsides or side effects?

FabeG:
Hi @Conengmo, I let the default to `false` to be in line with `leaflet` default. It seems there are still some drawbacks using Canvas as a default. See https://leafletjs.com/reference-1.0.0.html#canvas : 

> Canvas is not available in all web browsers, notably IE8, and overlapping geometries might not display properly in some edge cases

Conengmo:
Thanks for looking into this. We'll leave it at default False then.

This is ready IMO, I'll merge it later this week. Thanks for your contribution!

Issue #1129, opened 2019-04-11
Unable to cluster a geojson layer


fullonic:
#### Please add a code sample or a nbviewer link, copy-pastable if possible

# Your code here

```
import folium
from folium.plugins import MarkerCluster

map = folium.Map()
geo_data = folium.GeoJson("filename.geojson")

# Create cluster layer
cluster = MarkerCluster()
cluster.add_child(geo_data)

# Add cluster to map
map.add_child(cluster)

map.save("cluster.html")
```
#### Problem description
The problem only happen when using folium installed from the master branch. Folium version 8.3 works as expected, all points are clustered. However, with the master version any point is displayed.

#### Expected Output
Map with clustered markers.
#### Output of ``folium.__version__``
`'0.8.3+41.g0e527a3.dirty'`

#### Possible solution
After reading and comparing the GeoJson class from version 8.3 and master I saw that the possible problem is the new way of adding the data to the geojson layer.

Adding a couple of extra lines to the GeoJson template fixes the problem:
````        
(...)
{% if this._parent.get_name()|truncate(14, True, "") == "marker_cluster" %}
    {{ this.get_name() }}.addTo({{ this._parent.get_name() }})
{% endif %}
{%-endmacro %}
````
I can't figure out other workaround, but this one works.

Conengmo:
This has been fixed in #1190. Thanks for looking into this and collaborating on it @fullonic!

PR #1136, opened 2019-05-03
Fix1129: Makes possible cluster geojson data.

fullonic:
Solves issue #1129.


folium/features.py
```diff
@@ -91,7 +91,8 @@ def render(self, **kwargs):
                                             'if it is not in a Figure.')
 
         figure.header.add_child(
-            JavascriptLink('https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),
+            JavascriptLink(
+                'https://cdnjs.cloudflare.com/ajax/libs/leaflet-dvf/0.3.0/leaflet-dvf.markers.min.js'),
```
ocefpaf: Please do not change the current layout of the code. It makes it harder to review.
We already have a `noqa` to keep this in a single line.

fullonic: Hi @ocefpaf, sorry about that! I just realized after I made the commit. It was automatic code format by the text editor because of the length of the line.
Shall I do a new commit undoing that or amend my local copy and push again?

ocefpaf: > Hi @ocefpaf, sorry about that! I just realized after I made the commit. It was an automatic code format by the text editor because of the length of the line.

No problem. I do that all the time.

> Shall I do a new commit undoing that or amend my local copy and push it again?

Whatever is easier for you. We can squash it later before merging.

fullonic: I pushed again with only my  additions.
Cheers!

fullonic:
@Conengmo , @ocefpaf  may you review this PR or issue #1129 , please? Version 9.1 was released however this problem keeps happening. This PR fixes the issue but I'm not sure if is the best way. 
Thanks!

Conengmo:
Hi @fullonic, sorry for the delay. I'll try to review soon, but I'm in the middle of moving so I don't have a desk with a computer yet to take the time for it ;) I would like to check what the issue is precisely, and if there's a more clean way to do it instead of adding an if-condition, that's duplicated in two places.

fullonic:
hi @Conengmo . Thanks for your fast reply. I also don't like of repeating the same if statement twice but was the faster way I found to solve the problem without changing any of the GeoJson current code. I think the break was made, if I'm not wrong, with PR #1058.
All the best in your moving :)

fullonic:
@Conengmo , I have another solution for this issue. Should I close this PR and create a new one? Or just create a new PR and point to it in this one?

Cheers!

Conengmo:
If it's different than feel free to open up a new PR :)

fullonic:
@Conengmo , I created a new PR with the new approach #1188 . May you checked and let me know your opinion. Sorry about the delay hehe
Thanks :)

PR #1140, opened 2019-05-04
added simple example using flask

penguindustin:
# Added

- example on how to use folium with flask

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,7 +32,7 @@ Contents
    :maxdepth: 3
 
    installing
-   flask
    quickstart.ipynb
+   flask
    modules
    plugins
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,6 +32,7 @@ Contents
    :maxdepth: 3
 
    installing
+   flask
    quickstart.ipynb
    modules
    plugins
--- a/examples/flask_example.py
+++ b/examples/flask_example.py
@@ -1,3 +1,18 @@
+""" flask_example.py
+
+    Required packages:
+    - flask
+    - folium
+    
+    Usage:
+    
+    Start the flask server by running:
+
+        $ python flask_example.py
+    
+    And then head to http://127.0.0.1:5000/ in your browser to see the map displayed
+
+"""
 from flask import Flask
 import folium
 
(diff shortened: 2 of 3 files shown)


examples/flask_example.py
```diff
@@ -0,0 +1,13 @@
+from flask import Flask, render_template
```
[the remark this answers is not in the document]
penguindustin: fixed

ocefpaf:
Awesome! Thanks @penguindustin!! This question pops up so frequently that it is a great addition to our docs. I'll leave it here for a while so @Conengmo can take a quick look at but I'm planning to merge it during the rest of the sprints.

penguindustin:
@ocefpaf thanks for the mentoring! I enjoyed working on the project 😊

docs/index.rst
```diff
@@ -32,6 +32,7 @@ Contents
    :maxdepth: 3
 
    installing
+   flask
```
Conengmo: If we are to include `flask.rst` in the documentation, I would suggest to move it below `quickstart.ipynb`. That file contains some more basic examples.

examples/flask_example.py
```diff
@@ -0,0 +1,15 @@
+from flask import Flask
```
Conengmo: Maybe add a docstring at the top of the file explaining how to use it. It's very basic but might be useful to people. Something about running this and then going to `http://127.0.0.1:5000/` in a browser.

docs/index.rst
penguindustin: Whoops, I meant to put it under `quickstart` thanks for the catch

examples/flask_example.py
penguindustin: Can do!

penguindustin: Added :D

penguindustin:
@Conengmo I reordered the index and added some extra info to the example file. Should I replicate it in the `.rst`? I think there's a way to reference the actual python file instead of copy pasta into the `.rst` but I'm no expert in Sphinx docs.

ocefpaf:
@Conengmo I think it is nice to have it in the docs and the best course of action, to avoid the duplication of code and changes of it going stale, is to reference the example file in the `rst` as @penguindustin suggests. I'm looking into how to do that.

ocefpaf:
Thanks @penguindustin!

PR #1144, opened 2019-05-04
added message when key_on not found in geojson

evwhiz:
This address #918 and raises a TypeError "key_on value not found in geojson" when key_on is given, but not found in the geojson file.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1110,8 +1110,7 @@ class Choropleth(FeatureGroup):
             def color_scale_fun(x):
                 key_of_x = get_by_key(x, key_on)
                 if not key_of_x:
-                    if len(key_on) > 0:
-                        raise("key_on value not found in geojson")
+                    raise ValueError("key_on `{!r}` not found in GeoJSON.".format(key_on))
 
                 if key_of_x not in color_data.keys():
                     return nan_fill_color, nan_fill_opacity
--- a/folium/features.py
+++ b/folium/features.py
@@ -1109,6 +1109,9 @@ class Choropleth(FeatureGroup):
 
             def color_scale_fun(x):
                 key_of_x = get_by_key(x, key_on)
+                if not key_of_x:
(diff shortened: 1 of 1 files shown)


ocefpaf:
@evwhiz this looks good. Thanks! I'll leave this for @Conengmo to take a look at but I'm planning on merging it during the rest of the sprints.

folium/features.py
```diff
@@ -1095,6 +1095,9 @@ def get_by_key(obj, key):
 
             def color_scale_fun(x):
                 key_of_x = get_by_key(x, key_on)
+                if not key_of_x:
```
Conengmo: You can combine these two lines into one line: `if not a and b > 0:`

folium/features.py
```diff
@@ -1095,6 +1095,9 @@ def get_by_key(obj, key):
...
+                    if len(key_on) > 0:
```
Conengmo: In what case will `key_on` be an empty string? And why don't we want to raise an error in that case? If `key_on` is indeed an empty string it will still fail, so good to raise an error I would say.

folium/features.py
```diff
@@ -1095,6 +1095,9 @@ def get_by_key(obj, key):
...
+                        raise("key_on value not found in geojson")
```
Conengmo: I don't know if this is the right syntax.  I would suggest to raise it with `raise ValueError('my message')`. A value error, because the type of `key_on` is right (it's a string), but its value is wrong.

Conengmo: Would be really helpful if we could include some information in the error message on what's wrong with the data. You could for example show the `key_on` value in the error message. Something like `'wrong value: {}'.format(value)`.

ocefpaf:
@Conengmo can you take a second look at this?

ocefpaf:
Thanks @evwhiz! Sorry it took me a while to get back to this.

PR #1151, opened 2019-05-14
Minicharts example using matplotlib

OneAdder:
This is an example of using `matplotlib` to draw small charts and binding them to markers as `html`.


ocefpaf:
Thanks @OneAdder!

PR #1152, opened 2019-05-16
make all example data remote

ocefpaf:
@Conengmo some high school teachers are using the `folium` gallery in "bare bones" binder and/or Google Colab to teach students how to code. This will make it easier for them b/c it won't requite a full github clone and/or download of the data used in  the examples.


ocefpaf:
The failure is due to some data that will be available once we merge this PR :smile: 

Self-merging this one b/c it is mostly documentation stuff and harmless.

PR #1153, opened 2019-05-17
Improve our quickstat docs

### LINKED ISSUE #1145: Feedback for Tutorial
Things to be Added to `folium` quickstart/tutorial:

- Seems like you need to get the location data from somewhere other than the map that is on the screen zoomed in.  Is this true?  If so, it might be helpful to give instructions on how to get the data from Google maps.

- It looks like all the iterations could be done within the initial cell to end with a final product.  It seems it could be helpful to explain the process of inserting a cell below to create separate examples that follow along with the tutorial.

- Click Control-Enter (after the m instance) is also something that I needed to know to have my map appear.

- Some exercises that may help a newbie:
  - How to change the image with the different types of maps available - seems there are different kinds available in folium
  - You wrote "The default tiles are set to OpenStreetMap, but Stamen Terrain, Stamen Toner, Mapbox Bright*, and Mapbox Control Room, and many others tiles are built in."
  - No image showed for Mapbox Bright - (turns out not all levels available)
  - No image showed for Mapbox Control Room
  - How to zoom in on the image with the different numbers.  For me, 16 was a better zoom that we had discovered during the Education Summit.  I tried it and my maps are zoomed in appropriately.  This may be obvious to a more experienced user or discovered by simply thinking about the zoom level command.
  - What colors are available for the list of markers and what other icons are available to put within the icon such as cloud and info sign?

ocefpaf:
Fixes #1145


ocefpaf:
Minor docs fixes. Self-merging this one.

PR #1155, opened 2019-05-25
Fix geojson identifier

### LINKED ISSUE #1154: GeoJSON feature ID creator fails when properties contain list/dict
This code throws a `TypeError`:
(**edit**: add geometry type to feature)
```python
import folium
json = {
    "type": "FeatureCollection",
    "features": [
        {
            "bbox": [
                8.546326,
                47.417879,
                8.548813,
                47.420362
            ],
            "type": "Feature",
            "properties": {
                "summary": {
                    "distance": 343.2,
                    "duration": 50.5
                },
                "way_points": [
                    0,
                    15
                ]
            },
            "geometry": {
                "coordinates": [
                    [
                        8.548813,
                        47.420362
                    ],
                    [
                        8.548795,
                        47.420345
                    ],
                    [
                        8.548621,
                        47.420167
                    ]
                ],
                "type": "LineString"
            }
        }
    ]
}

m = folium.Map(location=[47.417879, 8.546326], zoom_start=13)

folium.features.GeoJson(json, style_function=lambda x: {'opacity': 1.0}).add_to(m)
m
```
#### Problem description
If a `style_function` is passed, `folium` tries to find an ID for the GeoJSON features. When no `id` field is present (not required in GeoJSONs AFAIK), it will try to pick a field from properties by checking if the values of the property fields are unique. However, the current implementation with `set` to check uniqueness fails when the property field values are `dict`s or `list`s.

#### Quick Fix
It works with a try/except around the [the appropriate lines](https://github.com/python-visualization/folium/blob/5d14833595718acad49d45e0a30fb4225e3f9c56/folium/features.py#L531):
```python
try:
    if len(set(feat['properties'][key] for feat in features)) == n:
        return 'feature.properties.{}'.format(key)
except TypeError:
    continue
```
I'd be happy with this if was my library, as far as I can see nothing else could raise a `TypeError` here. What do you think?

#### Output of ``0.9.0 ``
```
    530         for key in feature.get('properties', []):
    531             # try:
--> 532             if len(set(feat['properties'][key] for feat in features)) == n:
    533                 return 'feature.properties.{}'.format(key)
    534             # except TypeError:

TypeError: unhashable type: 'dict'
```

Conengmo:
For styling and highlights we need a unique field on each geojson feature. This functionality was broken for geojson data with certain properties layouts. This change makes the functionality work for all properties allowed by the geojson spec.

Add a utility function to do a dynamic key lookup in a dictionary.

Add test cases for all changes.

Closes #1154.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -25,7 +25,6 @@ from folium.utilities import (
     none_min,
     get_obj_in_upper_tree,
     parse_options,
-    dict_get,
 )
 from folium.vector_layers import PolyLine, path_options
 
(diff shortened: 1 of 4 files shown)


Conengmo:
This part of the code is becoming a bit extensive, I'm doubtful if it's worth it.

We could instead decide not to support using properties as feature identifiers. That way, a feature either needs to have an id or data should be embedded.

Alternatively there are two other changes possible:
- don't require a feature identifier if the style function applies to all features equally.
- translate the style function to a Javascript function, so no other feature identifier is needed.

ocefpaf:
> This part of the code is becoming a bit extensive, I'm doubtful if it's worth it.

Probably not but I'll leave the call to you b/c you will be here longer than me :smile: 

> We could instead decide not to support using properties as feature identifiers. That way, a feature either needs to have an id or data should be embedded so we can add an id.

I'm OK with that.

> Alternatively there are two other changes possible:
> 
> * don't require a feature identifier if the style function applies to all features equally.
> * translate the style function to a Javascript function, so no other feature identifier is needed.

I'm not sure I like either of these but again, it is you call.

Conengmo:
I simplified it to the following cases:

- Each feature has a unique id.
- Each feature has a properties dict with a unique string or int value.

I think these two cases are well defined, so they're easy to test. I added multiple tests, mainly to verify we recover from getting faulty data.

PR #1156, opened 2019-05-25
Fix DualMap plugin layer control

Conengmo:
The layer control on the DualMap plugin doesn't get copied properly. To fix this, reset it on the second map at render time.

There's some other weird thing with the base layers in the layer control, but that hasn't to do with our code but with the underlying JS plugin.


ocefpaf:
Do you want to add something more or is this one good to go?

Conengmo:
Thanks for checking. I think this is good to go. It's not perfect but it get's the job done.

PR #1158, opened 2019-05-26
Fix stale bit.ly link

ocefpaf:
We were downloading an old version of `miniconda` and wasting CI time updating packages later.


Conengmo:
Good catch!

PR #1169, opened 2019-06-22
Explicit `key_of_x` check

### LINKED ISSUE #1168: ValueError raised when `key_of_x` = 0
This line:
https://github.com/python-visualization/folium/blob/f05f3ce9b3d71f87a0738f3c8c1236b933e497a9/folium/features.py#L1112
Provokes this `ValueError` to be raised, not only when `key_of_x` is `None` but also when `key_of_x == 0` (which I think is a perfectly valid value for a key):
https://github.com/python-visualization/folium/blob/f05f3ce9b3d71f87a0738f3c8c1236b933e497a9/folium/features.py#L1113
Shouldn't this line be `if key_of_x is None:`?

LeandroOrdonez:
An explicit check for None (`key_of_x == None`) in the `color_scale_fun(x)` method from class `folium.features.Choropleth` has been added, along with the corresponding (provisional?) test.

Closes #1168


tests/test_folium.py
```diff
@@ -230,6 +230,29 @@ def test_choropleth_features(self):
...
+        """
+        with open(os.path.join(rootpath, 'geo_grid.json')) as f:
+            geo_data = json.load(f)
+        data = pd.DataFrame({'idx': {'0': 0, '1': 1, '2': 2, '3': 3, '4': 4, '5': 5},
+                             'value': {'0': 78.0, '1': 39.0, '2': 0.0, '3': 81.0, '4': 42.0, '5': 68.0}})
+        fill_color = 'BuPu'
+        columns = ['idx', 'value']
+        key_on = 'feature.properties.idx'
+
+        try:
```
Conengmo: This test is good and indeed fails without the change `Choropleth` class you made.

I do think it's better to remove the try/except. It doesn't add anything. If you want to make it explicit that we expect the call to `Choropleth` NOT to raise an exception, you could do something like [this](https://stackoverflow.com/a/20275035), but IMO it's not necessary.

LeandroOrdonez: It's done @Conengmo! :) I removed the try/except block since, as you said, it doesn't add anything.

Conengmo:
Thanks @LeandroOrdonez for your contribution!

LeandroOrdonez:
Happy to help @Conengmo! :)

PR #1175, opened 2019-07-12
rendering draw and edit options to json

mccarthyryanc:
I wanted the ability to remove some the drawing methods from the Draw Toolbar. This could only be done if given a `false` value, which requires to options to be converted to json. Here is an example:
```python
import folium
m = folium.Map()
Draw(
    draw_options={
        'polyline':False,
        'polygon':False,
        'circle':False,
        'marker':False,
        'circlemarker':False,
    }
).add_to(m)
m.save('testing.html')
```
image

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -44,8 +44,8 @@ class Draw(MacroElement):
         {% macro script(this, kwargs) %}
             var options = {
               position: {{ this.position|tojson }},
-              draw: {{ this.draw_options }},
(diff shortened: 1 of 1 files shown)


Conengmo:
Does this mean the draw options is currently broken?

mccarthyryanc:
@Conengmo , I think so.

Conengmo:
Thanks for spotting this and making a PR @mccarthyryanc!

PR #1180, opened 2019-07-19
Fix JS error in Draw plugin export option

### LINKED ISSUE #1179: [Draw Plugin] TypeError: document.getElementById(...) is null

```
import folium

m = folium.Map()
Draw().add_to(m)
m.save("map.html")

```
#### Problem description

When we use the Draw plugin with `export=False` it creates a javascript `TypeError: document.getElementById(...) is null`
As consequence of this, all the follow code added after we add the Draw plugin don't run. For instance, if we want to use the follow code:
```
import folium
from folium.plugins import Draw, LocateControl

m = folium.Map()
Draw().add_to(m)
LocateControl().add_to(m)
m.save("map.html")
```
LocateControl plugin will never be available because of the type error.  The issue is produce by this part of the code:

https://github.com/python-visualization/folium/blob/0767eb953a1062140f3db919feeb0bc82d853bd6/folium/plugins/draw.py#L71

As there is any element with `#export `ID, it breaks.

#### Expected Output

No javascript error.

'0.9.1'

fullonic:
This PR fix #1179


Conengmo:
Nice @fullonic! Good find and thanks for the fix.

ocefpaf:
@Conengmo I'm planning on a new release with this fix. Do you have any more PRs you would like to include?

Conengmo:
Good call, let's do a release indeed. I would like to include https://github.com/python-visualization/folium/pull/1169, but that test needs a change. Let's wait a day and see if we get a reply on that, otherwise I can push a change to that branch.

Conengmo:
@ocefpaf it's ready! Let's do a release. I wrote the changelog: https://github.com/python-visualization/folium/commit/c1d76e9defb29b727e315f04fc0c111e8e70603a

Since we include a new feature (the LocateControl) plugin and we bumped Leaflet from 1.4.0 to 1.5.1 I figured this shouldn't be a patch release, but should be 0.10.0. Would you agree? I'm also okay with doing 0.9.2, but then you'd have to change the number in the changelog first before releasing.

ocefpaf:
> but should be 0.10.0. Would you agree

:+1: 

Working on it today. Thanks!

fullonic:
@Conengmo , LocateControl was realesed on version 0.9.0 hehe  :+1: 
https://github.com/python-visualization/folium/blob/bfba30ee34cd9b32ead50e06e37a1416c93f4f15/CHANGES.txt#L43

Conengmo:
Seems you're right. Weird, because the commit showed up in the diff from 0.9.1 to 0.10.0: https://github.com/python-visualization/folium/compare/v0.9.1...v0.10.0 I suppose these commits got snuck in again on another PR that wasn't squashed. @ocefpaf let's squash most PR's so it's easier to make the changelog.

PR #1181, opened 2019-07-19
Div icon typo

fullonic:
This is only to fix a DivIcon python code typo
https://github.com/python-visualization/folium/blob/0767eb953a1062140f3db919feeb0bc82d853bd6/folium/features.py#L1215


ocefpaf:
Good catch!

PR #1182, opened 2019-07-19
add amap tiles

anhenghuang:
This PR add amap tiles. Amap is a reliable and widely used map service in China. Most of users from here are not familiar with the existing built-in tiles in folium now. So there is a gap and we have to set amap tiles attr everytimes. Let's add more built-in tiles.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -78,6 +78,7 @@ class Map(MacroElement):
         - "Mapbox Bright" (Limited levels of zoom for free tiles)
         - "Mapbox Control Room" (Limited levels of zoom for free tiles)
         - "Stamen" (Terrain, Toner, and Watercolor)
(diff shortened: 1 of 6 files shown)


ocefpaf:
While I'm against adding more tiles, b/c we provide a way to users add whatever URL they want, you make a case about built-ins not representing your geographical region. So I'm inclined to merge this one. However, I'll defer to @Conengmo for the final word on this.

Conengmo:
We're aiming to move away from including all possible tile services in folium, instead only including a small core set and for the rest relying on websites like http://leaflet-extras.github.io/leaflet-providers/preview/ and the `xyzservices` package: https://github.com/geopandas/xyzservices. I suggest you ask them to include this tileset, see https://github.com/geopandas/xyzservices#contributors.

PR #1184, opened 2019-07-23
Removes warning about conflicts with Draw plugin

fullonic:
PR #1180 fixed this conflict.

### DIFF
--- a/examples/Plugins.ipynb
+++ b/examples/Plugins.ipynb
@@ -793,9 +793,6 @@
     "\n",
     "plugins.LocateControl().add_to(m)\n",
     "\n",
-    "# If used with Draw plugin, LocateControl needs to be added before it.\n",
-    "# Draw().add_to(m)\n",
-    "\n",
     "m"
    ]
   }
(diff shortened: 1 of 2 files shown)


Conengmo:
It did? Seems unlikely doesn't it, since all we did was add the if statement. If that's true, the situation is still the same.

Do you know what the conflict was, and why it's fixed now?

fullonic:
It fixes the conflict because it was related with the javascript type error caused by the Draw plugin. Fixing the js error fixed this conflict. 
Actually, the conflict wasn't only with the LocateControl plugin, I found it as well using the Fullscreen plugin and Draw together. I tested the LocateControl plugin after fixing the Draw js type error and it works without conflicts.

Conengmo:
Thanks for explaining. I haven't tried this out myself but I trust you to have looked into this. 

Before merging this, could you take out the commit that changes `CONTRIBUTING.md`?

fullonic:
@Conengmo, Thanks.
I removed the CONTRIBUTING.md. It was from updating my master branch with the folium master.
Cheers

Conengmo:
Thanks @fullonic! Nice we got this sorted out.

PR #1186, opened 2019-07-24
Fix heading length

ocefpaf:
Self-merging this one so we can do the new release.


Conengmo:
🤦‍♂

Conengmo:
But 🎉  on the release!

PR #1188, opened 2019-07-27
Fix geojson issue with marker cluster

### LINKED ISSUE #1129: Unable to cluster a geojson layer

fullonic:
This PR fixed #1129 
My understating of the cause of this issue is because we are currently adding a geojson with a null value to cluster marker layer and only later adding the geojson data.
<https://github.com/python-visualization/folium/blob/212262520e722ccdf0bd43497c1c691661d58b8d/folium/features.py#L416>

And because of this, the following code produces a empty map:

    import folium
    folium.__version__
      '0.10.0'
    from folium.plugins import MarkerCluster
    m = folium.Map()
    mc = MarkerCluster()
    geojson = folium.GeoJson("random_points.geojson")
    mc.add_child(geojson)
    mc.add_to(m)
    m.save("cluster.html")

The `cluster.html` html code:

    (....)

      var marker_cluster_e6eb5734160c404c8ba80bee9e023ff6 = L.markerClusterGroup(
          {}
      );
      map_555133c3ff134e328f848608cd78fce6.addLayer(marker_cluster_e6eb5734160c404c8ba80bee9e023ff6);

    function geo_json_93c25184bf6049d2a7b7b92750ddd587_onEachFeature(feature, layer) {
      layer.on({
          click: function(e) {
              map_555133c3ff134e328f848608cd78fce6.fitBounds(e.target.getBounds());
          }
      });
    };
    var geo_json_93c25184bf6049d2a7b7b92750ddd587 = L.geoJson(null, {
          onEachFeature: geo_json_93c25184bf6049d2a7b7b92750ddd587_onEachFeature,

    }).addTo(marker_cluster_e6eb5734160c404c8ba80bee9e023ff6);  # **Here we are adding the empty geojson layer**

    geo_json_93c25184bf6049d2a7b7b92750ddd587.addData({GEOJSON_DATA});
    (....)

However, with the same python code, if we change the html code and add the geojson data at the same time we create the geojson layer, as bellow, the markers are displayed on map.

    var geo_json_93c25184bf6049d2a7b7b92750ddd587 = L.geoJson({**GEOJSON_DATA**}, {
          onEachFeature: geo_json_93c25184bf6049d2a7b7b92750ddd587_onEachFeature,

    }).addTo(marker_cluster_e6eb5734160c404c8ba80bee9e023ff6);

This PR fixed this issue and at the same time makes it possible have data not embed on html file when an url is passed as data source.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -413,22 +413,23 @@ class GeoJson(Layer):
                 }
             });
         };
-        var {{ this.get_name() }} = L.geoJson(null, {
-            {%- if this.smooth_factor is not none  %}
-                smoothFactor: {{ this.smooth_factor|tojson }},
-            {%- endif %}
-                onEachFeature: {{ this.get_name() }}_onEachFeature,
-            {% if this.style %}
-                style: {{ this.get_name() }}_styler,
-            {%- endif %}
-        }).addTo({{ this._parent.get_name() }});
+        var options = {
+        {%- if this.smooth_factor is not none  %}
+            smoothFactor: {{ this.smooth_factor|tojson }},
+        {%- endif %}
+            onEachFeature: {{ this.get_name() }}_onEachFeature,
+        {% if this.style %}
+            style: {{ this.get_name() }}_styler,
+        {%- endif %} };
         {%- if this.embed %}
-            {{ this.get_name() }}.addData({{ this.data|tojson }});
+        var {{ this.get_name() }} = L.geoJson(
+            {{ this.data|tojson }}, options).addTo({{ this._parent.get_name() }});
         {%- else %}
-            $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true,
-                success: function(data) {
-                    {{ this.get_name() }}.addData(data);
-            }});
+        var json_url = $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true })
+        $.when(json_url).done(function() {
+            var geo_json_32b06f4a0e6d419c8f501b4e6c43bc57 = L.geoJson(json_url.responseJSON
+            , options).addTo({{ this._parent.get_name() }});
+            });
         {%- endif %}
         {% endmacro %}
         """)  # noqa


folium/features.py
```diff
@@ -413,22 +413,23 @@ class GeoJson(Layer):
...
+        var {{ this.get_name() }} = L.geoJson(
+            {{ this.data|tojson }}, options).addTo({{ this._parent.get_name() }});
         {%- else %}
-            $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true,
-                success: function(data) {
-                    {{ this.get_name() }}.addData(data);
-            }});
+        var json_url = $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true })
+        $.when(json_url).done(function() {
+            var geo_json_32b06f4a0e6d419c8f501b4e6c43bc57 = L.geoJson(json_url.responseJSON
```
Conengmo: The problem with this is that this breaks `LayerControl` for regular geojson use cases. This variable has to be declared and defined outside of the ajax call.

folium/features.py
```diff
@@ -413,22 +413,23 @@ class GeoJson(Layer):
...
-            {%- if this.smooth_factor is not none  %}
-                smoothFactor: {{ this.smooth_factor|tojson }},
-            {%- endif %}
-                onEachFeature: {{ this.get_name() }}_onEachFeature,
-            {% if this.style %}
-                style: {{ this.get_name() }}_styler,
-            {%- endif %}
-        }).addTo({{ this._parent.get_name() }});
+        var options = {
+        {%- if this.smooth_factor is not none  %}
```
Conengmo: Why do you remove the indentation here?

folium/features.py
```diff
@@ -413,22 +413,23 @@ class GeoJson(Layer):
...
-            {{ this.get_name() }}.addData({{ this.data|tojson }});
+        var {{ this.get_name() }} = L.geoJson(
+            {{ this.data|tojson }}, options).addTo({{ this._parent.get_name() }});
         {%- else %}
-            $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true,
-                success: function(data) {
-                    {{ this.get_name() }}.addData(data);
-            }});
+        var json_url = $.ajax({url: {{ this.embed_link|tojson }}, dataType: 'json', async: true })
+        $.when(json_url).done(function() {
```
Conengmo: You can link this to the ajax call directly: `$.ajax().done(function() {});`

folium/features.py
fullonic: Yes, in fact it breaks. I wasn't aware of it. I just tried your PR and works as expected :+1:

folium/features.py
fullonic: May be automatic linting, I didn't see it. Sorry

folium/features.py
fullonic: Have you tried? Because the problem I had was related with async ajax request and because of that I used a variable to check if the request was finished. I don't have experience with JS. I just searched about it and I found that solution in a blog post.

Conengmo: It seems to work [here](https://github.com/python-visualization/folium/pull/1190/files#diff-188d69782eb76f735a9d6a3c508f3aa3R432), and is also listed in the examples in the jQuery documentation, so assume it should be fine.

Conengmo:
This problem has been fixed in #1190.

PR #1190, opened 2019-07-28
GeoJson: add to parent after data

Conengmo:
When using `GeoJson` with non-embedded data, in JS, don't add the object to the parent object until the data has been added.

This should solve the issue where geojson doesn't work with marker_cluster anymore. While not breaking the interaction between geojson and layercontrol.

An alternative is to make the ajax call synchronously, but that doesn't seem like a better approach to me.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -421,14 +421,16 @@ class GeoJson(Layer):
             {% if this.style %}
                 style: {{ this.get_name() }}_styler,
             {%- endif %}
-        }).addTo({{ this._parent.get_name() }});
+        });
+        function {{ this.get_name() }}_add (data) {
+            {{ this.get_name() }}.addData(data)
+                .addTo({{ this._parent.get_name() }});
+        }
         {%- if this.embed %}
-            {{ this.get_name() }}.addData({{ this.data|tojson }});
+            {{ this.get_name() }}_add({{ this.data|tojson }});
         {%- else %}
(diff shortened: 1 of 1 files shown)


Conengmo:
Here's the code I used to test behavior, using a data file from the folium repo:

```
with open('examples/data/search_bars_rome.json') as f:
    data = json.load(f)
data_url = 'https://raw.githubusercontent.com/python-visualization/folium/master/examples/data/search_bars_rome.json'

m = folium.Map((41.9, 12.5), tiles='cartodbpositron')
marker_cluster = folium.plugins.MarkerCluster(name='mc').add_to(m)
folium.GeoJson(data_url, embed=False).add_to(marker_cluster)
folium.GeoJson(data_url, embed=False, name='geojson').add_to(m)
m.save('_map.html')

```

ocefpaf:
@Conengmo are you done with the PR or do you want to add a test case to avoid regression? (I'm fine merging this as-is.)

Conengmo:
@fullonic will have to check first, and discuss. 

A test case would be on the JS side, so first we’ll have to setup Selenium tests. I looked into that a bit the other day. Maybe discuss that in another place though.

fullonic:
@Conengmo I tried your PR and it work's. For me we are good to go with the merge. May be a test and it's it. :+1:

Conengmo:
Thanks for checking it out @fullonic. If this indeed solves your problem without breaking current behavior let's merge it. Would be good to add Selenium tests to our suite before doing another release.

Issue #1185, opened 2019-07-24
Choropleth doesn't bind if using numeric index


alessioarena:
#### Please add a code sample or a nbviewer link, copy-pastable if possible

```python
gdf = gpd.read_file('SA2 shapefile')
gdf.columns
# output: ['name', 'probability']
gdf.index.dtype
# output: dtype('int64')

m = folium.Map(location=[	-37, 144], zoom_start=7, tiles='cartodbpositron')
folium.Choropleth(
    geo_data=gdf.geometry,
    data= gdf.probability,
    key_on='feature.id',
    fill_color='YlGn',
    fill_opacity=0.7,
    line_opacity=0.2,
    nan_fill_color= 'purple',
    legend_name='Probability',
    highlight=True
).add_to(m)

```
#### Problem description
I'm pretty new to folium, but I did work with geopandas and pandas a lot.
I also cannot provide the data, but that should be reproducible with any shapefile if the problem is what I think is.

My understanding of the code above is that will try to:

- select the `geometry` column and pass it as a GeoSeries for conversion to GeoJSON
- select the `probability` column and pass it as a Series
- leveraging the common index, trying to bind the two on `feature.id` <-> `index`
- do some magic and spit the choropleth out

Unfortunately this is not working as expected, as no error was raised but all polygons have a nice purple color (NaN). The legend was correctly generated so I reckon is purely a merging/binding issue.

Doing some quick troubleshooting I suspect that the bind failed because the feature.id is now a string, while the Series index is numeric, but I may be very wrong.

This instead worked as expected:
```
folium.Choropleth(
    geo_data=gdf,
    data= gdf,
    columns=['name', 'probability'],
    key_on='feature.properties.name',
    fill_color='YlGn',
    fill_opacity=0.7,
    line_opacity=0.2,
    nan_fill_color= 'purple',
    legend_name='Probability',
    highlight=True
).add_to(m)
```

Happy to do additional troubleshooting if needed.

#### Expected Output
A nice colorful choropleth without NaNs (they were dropped at a previous step)

#### Output of ``folium.__version__``
'0.9.1'

Conengmo:
Hi @alessioarena, thanks for your report. I think you're right, if the index keys are integer and the feature ids are string, it won't match. I think here's the line where that goes wrong:

https://github.com/python-visualization/folium/blob/master/folium/features.py#L1115

[...]

PR #1193, opened 2019-08-04
handle numeric key_on values for Choropleth

alessioarena:
This addresses issue #1185 where geometries and data coming from GeoDataFrames
were not joined. This was caused by the index being casted to string
during the conversion to GeoJSON.

The new behaviour is the following:
- try the join as it is
- if index contained numeric value, cast to str and try again

This commit includes 3 new test cases where:
- 1st and 2nd were failing under old behaviour
- 3rd passes under both behaviours


folium/features.py
```diff
@@ -1060,12 +1060,27 @@ def __init__(self, geo_data, data=None, columns=None, key_on=None,  # noqa
...
+                                   for e in future_keys])
+
+            if hasattr(data, 'set_index'):
             # This is a pd.DataFrame
-            color_data = data.set_index(columns[0])[columns[1]].to_dict()
-        elif hasattr(data, 'to_dict'):
-            # This is a pd.Series
+                data = data.set_index(columns[0])[columns[1]]
+
+            # convert the data to dictionary
```
Conengmo: Nit: this comment is a bit superfluous

folium/features.py
```diff
@@ -1060,12 +1060,27 @@ def __init__(self, geo_data, data=None, columns=None, key_on=None,  # noqa
                 'in favor of the `bins` parameter.', DeprecationWarning)
 
         # Create color_data dict
-        if hasattr(data, 'set_index'):
+        str_color_data = None
+        if hasattr(data, 'to_dict'):
+            # This is a pd.Dataframe or pd.Series
+            # check for numbers in future color_data keys
+            future_keys = data[columns[0]] if columns is not None else data.index  # noqa
+            any_numeric_key = any([isinstance(e,
```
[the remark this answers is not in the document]
alessioarena: The reason why I included the "any" case is that with the previous code any feature having a numeric index will silently fail to be match with the GeoJSON dataset. On the other hand it is hard to capture those failed matches, as you may have a lot of features that in fact don't have a match.
On all tests I did, the GeoJSON conversion of the spatial dataset ends up with indices becoming strings, so I could not see a case for numeric values being preserved in the color dataset.

Really the problem is in the GeoJSON conversion and the fact that numeric values are not preserved, but I could not find a direct solution for that.

folium/features.py
alessioarena: Noted. I can push the change, but I'm keen to wait for the other discussion to be agreed upon and resolved

Conengmo (commented):
The original logic was actually not working for my test case, in which the geojson feature id is an integer and the data has a string index. I couldn't really work out what was going on, so I opted to simplify the logic instead: if the original key doesn't fit, try again with a str or int, depending on the original type. I hope this should be a bit simpler to grok.

PR #1197, opened 2019-08-10
 Keep map only in png screenshot 

Vayel:
When we call `m._to_png()` we get a sreenshot of the full window, even if the map is smaller. I changed that.

I didn't understand why the tests were marked as `xfail`. As they pass, I removed the decorator.


Vayel:
I guess the `xfail` is for Travis, right?

ocefpaf:
> I didn't understand why the tests were marked as `xfail`. As they pass, I removed the decorator.

> I guess the `xfail` is for Travis, right?

Now you know :smile: 

The weirdest thing is that they do pass on my Travis-CI instance!

Can you revert that change so we can merge this PR?

Vayel:
Done :)

ocefpaf (approved):

PR #1206, opened 2019-09-17
Add API_key param to folium.Map

kerrick-lyft:
`folium.Map`'s examples say that it takes an `API_key` param, but it's actually missing. Add it and pass it through to the `TileLayer`.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -135,6 +135,8 @@ class Map(MacroElement):
         rare environments) even if they're supported.
     zoom_control : bool, default True
         Display zoom controls on the map.
+    API_key: str, default None
+        API key for Cloudmade or Mapbox tiles.
     **kwargs
         Additional keyword arguments are passed to Leaflets Map class:
         https://leafletjs.com/reference-1.5.1.html#map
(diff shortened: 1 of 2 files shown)


kerrick-lyft:
@Conengmo @ocefpaf could you review this?

ocefpaf:
@kerrick-lyft I believe this was removed by design and we forgot to update the docs. The goal was to reduce the massive amount of kw in `folium.Map`. I'll let @Conengmo, who is the main dev on folium nowadays, confirm this and merge or close it.

kerrick-lyft:
@ocefpaf what is the correct way to pass an `API_key` if you're using `mapbox` tiles?

Conengmo:
That's right, we removed it when trying to reduce the number of kwargs in `folium.Map`. Relating to tilelayers this class now only accepts some basic arguments. If you want to do more fancy things with tilelayers, please use the `TileLayer` class in addition to `Map`.

Something like this:
```
m = folium.Map(tiles=None)
folium.TileLayer().add_to(m)
```

https://github.com/python-visualization/folium/blob/master/folium/raster_layers.py#L19

The `TileLayer` class currently still has an `API_key` argument, though we might deprecate it in the future to reduce maintenance burden. It's quite trivial for a user to put a key in a URL, but hard for us to keep up with ever changing paid API requirements.

PR #1224, opened 2019-11-12
Closes #1221

### LINKED ISSUE #1221: HeatMapWithTime can not work

```python
# Your code here
HeatMapWithTime can not work
version python:3.7.4
```
#### Problem description

https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/HeatMapWithTime.ipynb
this demo can not work

'0.10.0'

mapio:
This appears to close the issue in the subject, based on the comment https://github.com/python-visualization/folium/issues/1221#issuecomment-552280614 by @farisnanosoft

### DIFF
--- a/folium/plugins/heat_map_withtime.py
+++ b/folium/plugins/heat_map_withtime.py
@@ -162,7 +162,7 @@ class HeatMapWithTime(Layer):
                                             'if it is not in a Figure.')
 
         figure.header.add_child(
-            JavascriptLink('https://rawcdn.githack.com/socib/Leaflet.TimeDimension/master/dist/leaflet.timedimension.min.js'),  # noqa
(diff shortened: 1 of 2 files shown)


Conengmo:
Hi @mapio, thanks for opening this PR. I tried it out and it indeed solves the issue for me. But if you don't mind I would prefer to use the fix in https://github.com/python-visualization/folium/pull/1228, because that allows us to keep up to date with the Leaflet TimeDimension plugin, instead of staying on an older version.

As I said here https://github.com/python-visualization/folium/issues/1221#issuecomment-553972356 I'd appreciate somebody confirming this indeed solves the issue for them as well. Otherwise we'll resolve to your PR, to stay on the older version.

mapio:
No problem. Of course go with the fix that uses the most up-to-date version of the deps! I hadn't time to look into it… I just applied the suggestion I was reading few days ago in the bugrep.

PR #1050, opened 2019-01-09
Add SemiCircle Plugin

syberworld:

PR #1238, opened 2019-12-19
Add SemiCircle Plugin - reworked. 

kuaka:
I have tried to address the requested changes in pull request #1050 (Add SemiCircle Plugin).
I have also added the ability to change the colors etc for the semicircles.


folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+                        opacity: {{this.opacity}},
+                        startAngle:{{this.start_angle}},
+                        stopAngle:{{this.stop_angle}}
+                        })
+                        .addTo({{this._parent.get_name()}});
+                }
+            {% endmacro %}
+            """)
+
+    def __init__(self, location, radius=0, direction=None, arc=None, start_angle=None, stop_angle=None, fill=True,
```
Conengmo: In the `Circle` class that we have there is no default for the `radius` parameter. I think that makes more sense than having a default 0, because that may confuse users; there will be no circle and they won't see why. Can you remove the default 0? That will make radius a required parameter.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+    Map.add_plugin.
+    Use (direction and arc) or (start_angle and stop_angle)
+    Parameters
+    ----------
+    location: tuple of length 2, default None
+        The latitude and longitude of the marker.
+        If None, then the middle of the map is used.
+    radius: int, default 0
+        Radius of semicircle in meters
+    direction: int, default 0
```
Conengmo: Please update the docstring here to reflect that these defaults are not 0 but None.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+from folium.map import Marker
+from folium.utilities import _validate_location
+
+from jinja2 import Template
+
+
+class SemiCircle(Marker):
+    """
+    Creates a Semicircle plugin to append into a map with
+    Map.add_plugin.
```
Conengmo: `Map.add_plugin` is not a function we have... Maybe best to just remove that part.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, location, radius=0, direction=None, arc=None, start_angle=None, stop_angle=None, fill=True,
+                 fill_color='#3388ff', fill_opacity=0.5, color='#3388ff', opacity=1, **kwargs):
+        super(SemiCircle, self).__init__(_validate_location(location), **kwargs)
+        self._name = 'SemiCircle'
+        self.radius = radius
+        self.direction = json.dumps(direction)
+        self.arc = json.dumps(arc)
+        self.start_angle = json.dumps(start_angle)
+        self.stop_angle = json.dumps(stop_angle)
+        if fill:
```
Conengmo: This variable doesn't exist?

You could consider using the `path_options()` function from `vector_layers.py`. Look for example at the `Circle` class to see how it's used.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+            self.fill = 'true'
+        else:
+            self.fill = 'false'
+        self.fill_color = fill_color
+        self.fill_opacity = fill_opacity
+        self.color = color
+        self.opacity = opacity
+        self.kwargs = json.dumps(kwargs)
+
+        if not (direction and arc) and not (start_angle and stop_angle):
```
Conengmo: Suggestion: also raise an error if both direction and arc *and* start_angle and stop_angle are provided. If a users doesn't know that in that case the start_angle and stop_angle are ignored it may confuse them.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+        Line color
+    opacity: float 0-1, default 1
+        Line opacity
+    **kwargs:
+        Passed to Marker so a popup and tooltip can be provided.
+
+    """
+    _template = Template(u"""
+            {% macro script(this, kwargs) %}
+                if ({{this.direction}} || {{this.arc}}) {
```
Conengmo: Question: can you make this code more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)? If not that's okay, but if you want to do that that would be welcome.

Conengmo:
I forgot one thing: could you also add an entry to the `Plugins.ipynb` notebook? If not that's okay, we can do that later, but it would help users discover this new plugin.

folium/plugins/semicircle.py
kuaka: I see what you mean. Would you do this in python, by breaking the string into parts and using one part twice or would you do it in "JavaScript" within the template. JavaScript is not my forte to say the least. I tried a couple of things but they did not work. I realised today that maybe you were intending to do it directly with python. What do you suggest?

folium/plugins/semicircle.py
kuaka: I thought I had left a comment on this. However it appears not. I can't remember exactly what I had wanted to ask but in any case it is not clear to me what variable is the one in question.

kuaka:
Hi, I have made some of the changes requested and asked for clarification on the others.

kuaka:
Any chance of getting this looked at? I want to use this in my project ideally via folium otherwise I will have to use the semi circle code separately. I am sure it will benefit others if included in folium however. thanks.

Conengmo:
Hi @kuaka, thanks for the reminder, I'll take a look shortly!

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def render(self, **kwargs):
+        super(SemiCircle, self).render(**kwargs)
+
+        figure = self.get_root()
+        assert isinstance(figure, Figure), ('You cannot render this Element '
+                                            'if it is not in a Figure.')
+
+        figure.header.add_child(
+            JavascriptLink('https://cdn.jsdelivr.net/npm/leaflet-semicircle@2.0.2/Semicircle.min.js'),
```
Conengmo, line 107: Please use version 2.0.4 since it fixes [this compatibility issue](https://github.com/jieter/Leaflet-semicircle/pull/31)

folium/plugins/semicircle.py
Conengmo: I checked and what I think is the most sane approach is this:

- Pass `start_angle` and `stop_angle` into `parse_options`. If they are not None they will be added to the options.
- Store `direction` and `arc` in an optional tuple in the instance: `self.direction = (direction, arc) if direction and arc else None`
- In the template check if `this.direction` is set and apply `setDirection` if so: `{% if this.direction %}.setDirection({{ this.direction | tojson }}){% endif %}`

This way we don't need the big if/else.

In the Python part you already have a check if both angles and direction are not set at the same time.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+    **kwargs:
+        Passed to Marker so a popup and tooltip can be provided.
+
+    """
+    _template = Template(u"""
+            {% macro script(this, kwargs) %}
+                if ({{this.direction}} || {{this.arc}}) {
+                    var {{this.get_name()}} = L.semiCircle(
+                        [{{this.location[0]}},{{this.location[1]}}],
+                        {radius:{{this.radius}},
```
Conengmo, line 51: After using `parse_options` you can put the options in the template like this: `{{ this.options|tojson}}`. Like here: https://github.com/python-visualization/folium/blob/master/folium/plugins/polyline_text_path.py#L49

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+        Heading of the stop angle value between 0 and 360 degrees.
+    fill_color: string, default '#3388ff' (blue)
+        Fill color
+    fill_opacity: float 0-1, default 0.5
+        Fill opacity
+    color: string, default '#3388ff' (blue)
+        Line color
+    opacity: float 0-1, default 1
+        Line opacity
+    **kwargs:
```
Conengmo, line 42: The kwargs are not used in the code. I suggest to pass them into `parse_options` like we do [here.](https://github.com/python-visualization/folium/blob/master/folium/plugins/polyline_text_path.py#L60)

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import (absolute_import, division, print_function)
```
Conengmo, line 3: We dropped support for Python 2 so we don't need these future imports anymore.

folium/plugins/semicircle.py
```diff
@@ -0,0 +1,108 @@
...
+        Line opacity
+    **kwargs:
+        Passed to Marker so a popup and tooltip can be provided.
+
+    """
+    _template = Template(u"""
+            {% macro script(this, kwargs) %}
+                if ({{this.direction}} || {{this.arc}}) {
+                    var {{this.get_name()}} = L.semiCircle(
+                        [{{this.location[0]}},{{this.location[1]}}],
```
Conengmo, line 50:
```suggestion
                        [{{ this.location|tojson }}],
```

folium/plugins/semicircle.py
kuaka: I like that, it's elegant and reduces the template considerably.

kuaka:
I have made the requested changes and also updated the code in line with the latest code in master as somethings had changed quite a bit. I have also added a test file.

Conengmo:
Thanks @kuaka, nice work on this one!

PR #1251, opened 2020-01-22
Change input from deprecated data.to_crs(epsg=4326) to data.to_crs('E…

artnikitin:
The `process_data` method of the `folium.GeoJson` class is using a now deprecated input syntax for re-projecting in `data.to_crs(epsg=4326)`. This triggers a `FutureWarning: '+init=<authority>:<code>' syntax is deprecated. '<authority>:<code>' is the preferred initialization method.` in the `pyproj/crs.py` module. The actual reformatting to '+init=...' syntax is done by the `geopandas/geoseries.py` module, because the old syntax doesn't provide crs as input, only epsg. Warning can be silenced by changing the syntax to `data.to_crs("EPSG=4326")`. Discussion of this problem on stackoverflow is [here](https://stackoverflow.com/questions/59596835/open-street-map-pyproj-how-to-solve-syntax-issue/59703971#59703971).

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -492,8 +492,6 @@ class GeoJson(Layer):
         elif hasattr(data, '__geo_interface__'):
             self.embed = True
             if hasattr(data, 'to_crs'):
-                if isinstance(data.crs, dict) and 'init' in data.crs:
-                    data.crs = data.crs['init']
                 data = data.to_crs('EPSG:4326')
             return json.loads(json.dumps(data.__geo_interface__))
         else:
--- a/folium/features.py
+++ b/folium/features.py
@@ -492,7 +492,7 @@ class GeoJson(Layer):
         elif hasattr(data, '__geo_interface__'):
             self.embed = True
             if hasattr(data, 'to_crs'):
-                data = data.to_crs(epsg='4326')
+                data = data.to_crs('EPSG:4326')
             return json.loads(json.dumps(data.__geo_interface__))
         else:
             raise ValueError('Cannot render objects with any missing geometries'
--- a/folium/features.py
+++ b/folium/features.py
@@ -492,6 +492,8 @@ class GeoJson(Layer):
         elif hasattr(data, '__geo_interface__'):
             self.embed = True
             if hasattr(data, 'to_crs'):
+                if isinstance(data.crs, dict) and 'init' in data.crs:
+                    data.crs = data.crs['init']
                 data = data.to_crs('EPSG:4326')
             return json.loads(json.dumps(data.__geo_interface__))
         else:


Conengmo:
Thanks for your PR, your description and change look good. I’ll test it out next time I’m working on folium and merge if nothing comes up, which I suspect not.

artnikitin:
ok, thanks :)

Conengmo:
Your change is fine, but unfortunately we still get a futurewarning because of other libraries. Do you have an idea how we could deal with that? I was thinking of maybe surpressing it at that location. A futurewarning should be read by end-users, and since there is nothing here to do for them I don't like to give them false warnings.

artnikitin:
Could you please elaborate on the warnings you have, what other libraries trigger them, in particular? My FutureWarning was raised by the `pyproj` library, and after tracking it down, fixing the input syntax here seemed like a good idea.

Conengmo:
Changing the syntax is indeed a good idea, but I still get the warning:

```
>>> df = geopandas.read_file('tests/us-states.json')
>>> a = df.to_crs('EPSG:4326')
[...]\lib\site-packages\pyproj\crs.py:77: FutureWarning: 
'+init=<authority>:<code>' syntax is deprecated. 
'<authority>:<code>' is the preferred initialization method.
  return _prepare_from_string(" ".join(pjargs))
```
It seems because the current crs of the df still uses the old syntax. In the Stackoverflow post you linked to they suggest setting the current crs by hand. We could do that programmatically:

```
>>> df.crs
    {'init': 'epsg:4326'}
>>> if isinstance(df.crs, dict) and 'init' in df.crs:
...     df.crs = df.crs['init']
>>> df.crs
    'epsg:4326'
```

What do you think, good idea? Is this robust?

artnikitin:
Yeah, this looks robust to me. Just to be clear, there are two cases, in which the FutureWarning is triggered. The first one is when you set `df.csr` as a dictionary (a.k.a `{init: 'epsg:4326}`). Your code fixes that quite well. The other one is the re-projecting syntax: when you pass an `epsg` parameter to `df.to_crs`, instead of a `crs` parameter as a string in `epsg:4326` format.

Conengmo:
You’re right, there are indeed two warnings. Do you want to update your PR to address both?

artnikitin:
I've also noticed, that guys from `geopandas` updated their code couple of days ago to face the same problem: https://github.com/geopandas/geopandas/commit/1d212b2a0c2a993f7d3b9a505f314be37b99a7d7. It looks like, you will be able to safely use the `epsg=4326` syntax in `df.to_crs` without any FutureWarnings, because they are now converting it to a proper crs string. But, it probably still won't hurt to use the string version though.

Conengmo:
Hi Artem, good to hear the Geopandas people fixed it upstream. I checked and indeed no warnings are raised anymore. But I agree it would be good to use the `EPSG:xxxx` since that seems more future proof.

I reverted your second commit, since Geopandas no longer uses the `{'init': 'epsg:4326'}` syntax.

After the tests passed I'll merge this PR. Thanks for your help in investigating and resolving this!

Conengmo:
Notebooks test failures are unrelated and will be addressed separately.

artnikitin:
Wow, great! glad to help!

PR #1256, opened 2020-02-04
Update css url to https

sknzl:
Changing the URL of `leaflet.timedimension.control.min.css` to the https CDN.
Related to [this problem](https://github.com/python-visualization/folium/issues/1231#issuecomment-582025815) in #1231

PR #1259, opened 2020-02-08
Update url wms data

sknzl:
Problem already described [here](https://github.com/python-visualization/folium/pull/1256#issuecomment-582341337) in PR #1256 .
URL to wms data is outdated because server moved. This also makes Travis CI fail.


Conengmo:
Hi @sknzl , thanks for helping out with this. Can you run the notebook and include that in this PR? That way [the version we show here](https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/WmsTimeDimension.ipynb) is up to date.

Conengmo:
I just tried updating the notebook but the new server seems down as well... https://pae-paha.pacioos.hawaii.edu/thredds/ 

If it stays like this we should not test this notebook anymore, it's casing to many headaches.

Conengmo:
I updated the notebook, here's the updated version: https://nbviewer.jupyter.org/github/sknzl/folium/blob/update-url-wms-data/examples/WmsTimeDimension.ipynb

If the tests pass I'll merge this.

PR #1261, opened 2020-02-12
Update fill_color default value from 'blue' to 'Blues'

deepserket:
Some color codes were changed in [this commit](https://github.com/python-visualization/branca/commit/db7bd6cec3729242f7fa44e5c5a903a788d999bf) so now the value 'blue' does not exist and it [throws an error](https://stackoverflow.com/questions/60191092/blue-is-not-a-valid-colorbrewer-code-error-while-creating-choropleth-using-fol)

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1064,7 +1064,7 @@ class Choropleth(FeatureGroup):
         If `bins` is a sequence, it directly defines the bin edges.
         For more information on this parameter, have a look at
         numpy.histogram function.
-    fill_color: string, default 'blue'
+    fill_color: string, default 'Blues'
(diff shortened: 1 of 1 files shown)


Conengmo:
@deepserket thanks for looking into this! I checked this out and you're right, the default value doesn't work if you also pass in data. Unfortunately this class also supports the use-case where no data is passed, so we'll have to support that with the default `fill_color` as well. I made a PR that builds on your change, can you check it out and see if this works for you? https://github.com/python-visualization/folium/pull/1288

Conengmo:
I merged #1288 so this PR is no longer relevant. Thanks for your help with this one @deepserket !

PR #1267, opened 2020-03-05
Fix overlay=True by default in Choropleth's doc

ilyankou:


Conengmo:
That's a typo, good find! That test failure is unrelated to this change so I'll go ahead and merge this. Thanks!

Issue #1080, opened 2019-03-05
Add PolyLine Offset plugin


FabeG:
It would be nice to implement the PolyLine Offset plugin (can be found here : https://github.com/bbecquet/Leaflet.PolylineOffset), in order to draw parallel polylines.
I can propose a PR if you think it might be interesting.

ocefpaf:
@FabeG while I'm not against adding this to the folium plugin submodule you can create the plugin as a separete package. That would allow you to evolve the plugin faster, without waiting for folium releases and review process.

FabeG:
Thanks for your answer @ocefpaf. 
Actually, I had just developped this plugin for my own needs (not too difficult: inheritance from PolyLine class, just adding an offset argument), and I was wondering if it could be also useful for other folium users. No problem then if you think integrating this plugin in the folium plugin submodule is not that interesting.

ocefpaf:
> No problem then if you think integrating this plugin in the folium plugin submodule is not that interesting.

It is not that it is uninteresting. It is about maintenance burden and quick development. At the early stages of folium we incorporated all the plugins and that makes it hard for us now to update the underlying JS libraries and still get all the code to pass the tests. In hindsight we should've done the same model as leaflet itself, where each plugin is a different package. I do encourage you to make that plugin a separate package, if not, maybe @Conengmo may be a better person to ask if we should add that to folium itself or not.

Conengmo:
Hmm I agree on making sure we don't overburden ourselves with features. But then I don't think most of our time goes to light-weight plugins. If we only provide a simple wrapper around a Javascript plugin, we don't have that much to maintain, while providing possibly useful functionality. I also see the value of having a single authoritative package for discovery.

But some current plugins do make our lives hard. `TimeSliderChoropleth` comes to mind. Maybe we can define some criteria on which to decide whether new plugins are wanted. Here's a proposal:

* the plugin provides interesting new functionality.
* the wrapper template is simple.
* the wrapper class has not much logic, just passing some things to the template.
* no/little integration with other folium classes.

As well as these criteria for the process:
* the contributor communicates well.
* the first PR is of reasonably good quality.
* the contributor can add a docstring, tests, entry in the plugins notebook and possibly a standalone example notebook.

@FabeG I think the plugin you link to is quite simple and won't make folium that much more complex. I'm thinking it could be a class that just calls (in JS) `pl.setOffset()` on a `PolyLine`. That seems simpler than inheriting and providing the same arguments twice.
[...]

PR #1292, opened 2020-04-14
Add SideBySide plugin

fralc:
Basic parsing of the leaflet-side-by-side plugin following other plugins implementation


Conengmo:
Hi @fralc, thanks for your PR, it looks neat. We already got a PR a while back for the same plugin though: https://github.com/python-visualization/folium/pull/1032 At that time I didn't merge it because it seemed redundant with the `DualMap` plugin we already have. Do you think we should reconsider that?

fralc:
Hi @Conengmo, I have not seen `DualMap` plugin. Actually usages of Leaflet plugins [side-by-side](https://github.com/digidem/leaflet-side-by-side) and [sync](https://github.com/jieter/Leaflet.Sync) mostly coincide. One could expect to find both folium plugins since they are both available at pure Leaflet level and because their user experience is a bit different as sync let you drag a map as the other is synchronized, while side-by-side let you drag a separator between two maps.

However, I do not know which are the principles for introducing folium plugins, so I'm not sure if you should reconsider introducing it.

Conengmo:
I defined some criteria for plugin acceptance here: https://github.com/python-visualization/folium/issues/1080#issuecomment-471319344. I think we're good on all accounts except maybe for the interesting new functionality. But looking back on that previous PR I was maybe a bit too strict. If you would like to add this, then I suppose it's wanted.

To merge this, I think we require:
- An entry to the plugins example notebook.
- A different CDN. For github stuff we usually use http://raw.githack.com/.
- A docstring example that's simpler. Showing all this WMSTileLayer stuff is a bit unnecessary.
- Importing the class in `folium.plugins`.

fralc:
@Conengmo I don't see what is wrong with the last commit. It seems a problem with a notebook but I cannot spot the problem. Could you take a look?

Conengmo:
Travis build failing is not related to your change. We actually merged a PR yesterday that fixes it, so if you rebase onto master the tests should pass. But you can also leave it like it is now, that's okay.

So is this PR ready for review again?

fralc:
Yes, it is ready for a review

Conengmo:
Link to notebook changes: https://nbviewer.jupyter.org/github/fralc/folium/blob/master/examples/Plugins.ipynb It seems the notebook outputs are empty.

I have some comments on the controls with this plugin. 

Adding the SBS layer to the control doesn't seem to do anything. The slider will appear even if you untoggle it. So we might as well remove the `control`, `overlay` and `show` parameters.

By default if you create a `TileLayer` object it will have `overlay=False`. If both the tile layers you pass into SBS have `overlay=False` only one of them can be activated at the same time. I can think of two ways to resolve this:

- make it clear in the docstring and plugin notebook that tile layers you want to use in SBS have to have `overlay=True`.
- tile layers that are passed into SBS are set to be overlays automatically (e.g. `left_layer.overlay = True`)

Both is also possible. What are your thoughts on these comments?

Two things about the CDN link: I see I mentioned raw.githack.com before, but that is actually the development subdomain. We should use the rawcdn production subdomain instead. And better to use the minified JS version. https://rawcdn.githack.com/digidem/leaflet-side-by-side/6ea4567fe8d1b3e6ea3b0e9ca05d311f5be125a3/leaflet-side-by-side.min.js should do it.

giswqs:
Any updates on this SideBySide plugin? I would love to use this. See the demo below with ipyleaflet.

Conengmo (approved):
Looks good!

I updated the class to use our JSCSSMixin, which simplifies dependency statements. Also included a CND link using jsdeliver.net.

Since layer control does nothing for this plugin, I removed those arguments and hard-coded the plugin to not appear in layer control.

Let's merge this when the tests pass.

giswqs:
I think it would be great to have an option that allows users to keep the layers control. This is particularly useful when users want to display multiple layers on the map when the two layers for the layer control do not cover the entire area.

giswqs:
Here are some user cases where the layers control would still be useful. 
https://leafmap.org/notebooks/49_split_control/

Conengmo:
In the first example on the page you shared, I'm talking about the 'Split Control' checkbox which does nothing.

giswqs:
Sorry for my misunderstanding. That makes sense. The Split Control itself should not show up as a layer in the layers control It should be removed.

Conengmo:
No worries, thanks for looking out!

Conengmo:
For anybody checking this out in the future, here's an example of having a base layer and two overlays that can be slided.

```
m = Map(tiles='cartodbpositron')

layer_right = TileLayer('openstreetmap', overlay=True)
layer_left = TileLayer('stamen toner', overlay=True)

sbs = plugins.SideBySideLayers(layer_left=layer_left, layer_right=layer_right)

layer_left.add_to(m)
layer_right.add_to(m)
sbs.add_to(m)

LayerControl(collapsed=False).add_to(m)
```

giswqs:
Looking forward to the new release. Thank you!

Conengmo:
and thanks to @fralc for the PR! Sorry it took so long.

fralc:
Sorry for not helping with the final steps, thank you for completing this work!

PR #1294, opened 2020-04-16
Remove iter_points utility function

Conengmo:
We have a utility function `iter_points` that's used [in only one plugin](https://github.com/python-visualization/folium/search?q=iter_points&type=Code). We don't need it, because we have a better function called `iter_coords`. Furthermore, in that one plugin we can even remove a bunch of code and replace it with the same functionality we already have in the `get_bounds` function.


Conengmo:
@ocefpaf I cleaned up one of the utility functions, can you review this one? There is a test that verifies the bounds of `TimestampedGeoJson` so I'm fairly sure this is safe to merge.

ocefpaf:
Thanks!

PR #1297, opened 2020-04-19
Type hints [WIP]

Conengmo:
I tried out adding type hints to the main modules (not the plugins).

**Rationale**
For users that use folium in an ad-hoc way the type hints may help in finding out what input a function accepts. And they may use an IDLE that gives those warnings automatically.

Users that write software with folium can now run Mypy on their software and get warnings for when their code is not compatible with folium.

For folium developers it becomes easier to work, because with explicit types you can change things with more confidence. And it forces you to think better about how a function should work.

**Remarks**
- This makes folium in compatible with Python 3.5 because of some variable type hints that are necessary.
- Many of the data inputs can be anything, a list, dict, Numpy array, Pandas dataframe, or something else... I typed these with `Any`, which is fine.

**Changes**
- Split the `validate_locations` function into a version for normal lines and one for multi-lines. This makes it more robust for classes that use normal lines and the type hints more accurate.
- `path_options`: define the two keyword arguments on usage. This is necessary for mypy and is more explicit as well. Change the `radius` argument from a bool/float into an optional float.
- Mypy cannot handle function definitions that replace optional arguments, so I replaced these with lambdas.

**Unrelated changes**
- `get_obj_in_upper_tree` had a bug. It assumed the top of a tree was reached if there's no `_parent` instance attribute. But actually all `Element` instances have this attribute, but it's `None` if there's no parent.

**Merge plan**
Don't merge this until we did a release. Ideally have Selenium tests up first so we can merge this with more confidence.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ geckodriver.log
 geckodriver.tar.gz
 geckodriver/
 miniconda.sh
+.mypy_cache
--- a/folium/features.py
+++ b/folium/features.py
@@ -9,13 +9,14 @@ import json
 import warnings
 import functools
 import operator
+from typing import Any, Dict, Callable, Sequence, Optional, Union, List, Tuple, Iterable
 
 from branca.colormap import LinearColormap, StepColormap
 from branca.element import (Element, Figure, JavascriptLink, MacroElement)
 from branca.utilities import color_brewer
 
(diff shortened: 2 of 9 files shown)


Conengmo:
Let's revisit this after we ran `black` on everything.

PR #1298, opened 2020-04-19
Fix GeoJsonPopupAndTooltip example notebook

Conengmo:
This notebook is currently failing the tests. It's because of using `fiona.crs.from_epsg()`. We only import `fiona` but would also require `fiona.crs`.

I fixed this by not using Fiona to begin with. We can just pass `'EPSG:4326'` directly.

While I was there I did some slight reformatting as well.

The notebook can be seen here: https://nbviewer.jupyter.org/github/conengmo/folium/blob/fix-notebook-geojson-popup-tooltip/examples/GeoJsonPopupAndTooltip.ipynb

Compare with the current version, it seems unchanged: https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/GeoJsonPopupAndTooltip.ipynb


Conengmo:
Only the `WmsTimeDimension` notebook test failed, which will be addressed separately.

PR #1304, opened 2020-04-20
Enabled 'position' to control placement of the search box (issue 1302)

### LINKED ISSUE #1302: Cannot place search plugin in non-default locations
See https://nbviewer.jupyter.org/github/jjbenes/covid19/blob/master/jupyter/covid19_jhu_csse.ipynb#Set-up-search-box-for-Folium-maps.

```python
  folium.plugins.Search(
      layer=search_geo,
      position='topright',
      geom_type='Polygon',
      placeholder='Search for a State or a County',
      collapsed=True,
      search_label='Combined_Key',
      weight=1,
  ).add_to(folium_map)
```
#### Problem description
The `position` argument of `plugins.Search` has no effect unless `geom_type == 'Point'`. Not sure why this constraint is needed. I use `geom_type` `Polygon` in my GeoJSON.

See line 57 in `search.py`: https://github.com/python-visualization/folium/blob/v0.10.1/folium/plugins/search.py#L57.

#### Expected Output
Once I moved line 57 (`position:'{{this.position}}',`) out of the if statement and put it right after line 51 (`textPlaceholder: '{{this.placeholder}}',`) , I was able to move the search box to the top-right corner.  Please let me know if this is the right approach.    

Screen Shot 2020-04-19 at 5 50 09 PM

0.10.1

jjbenes:
The search plugin was stuck at the top-left corner in 0.10.1. Fixed by moving the "position" attribute out of the "if this.geom_type" statement since positioning of the search box shouldn't be conditional on the GeoJSON geometry type.

Closes #1302

### DIFF
--- a/folium/plugins/search.py
+++ b/folium/plugins/search.py
@@ -49,12 +49,12 @@ class Search(MacroElement):
                 {% endif %}
                 collapsed: {{this.collapsed|tojson|safe}},
                 textPlaceholder: '{{this.placeholder}}',
+                position:'{{this.position}}',                
             {% if this.geom_type == 'Point' %}
                 initial: false,
                 {% if this.search_zoom %}
                 zoom: {{this.search_zoom}},
                 {% endif %}
-                position:'{{this.position}}',
                 hideMarkerOnCollapse: true
             {% else %}
                 marker: false,


Conengmo:
Thanks for your help @jjbenes !

PR #1311, opened 2020-04-26
Selenium notebook tests

Conengmo:
Do selenium tests on our example notebooks. The steps are:

- execute the notebook so it's using our up to date code
- convert the notebook to html
- extract iframes from the html
- store each iframe in an html file
- load each html file with selenium
- check if the map loaded
- verify there are no errors in the JS console

After merging this we can see what other tests can be discarded.

In the contributing guide I removed the part about Firefox. We still use it in the code but those tests are marked to allow failure. I want to revisit that code after this.

Reuse the WebDriver instance so the cache is preserved. This means doing less requests for static content.


Conengmo:
@ocefpaf I can't seem to get the tests working on Travis... Can you help with this? I want to use Chrome, because checking for Javascript errors in Selenium only works for headed Chrome.

I keep seeing this error:

```
selenium.common.exceptions.WebDriverException: Message: Chrome failed to start: exited abnormally
(unknown error: DevToolsActivePort file doesn't exist)
(The process started from chrome location /opt/google/chrome/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

I don't know enough about Selenium and Travis to fix this.

amit1rrr:
@Conengmo @stickler-ci Sorry to intrude but I noticed this PR contains Jupyter notebook files that are hard to review on GitHub due to the JSON diffs.

You might want to checkout [ReviewNB](https://www.reviewnb.com/), code review tool for notebooks. Provides visual diffs & commenting for Jupyter notebooks on GitHub. Completely free for open source repositories.

Disclaimer: I built ReviewNB

diff_and_commenting

Issue #351, opened 2016-02-11
Use folium offline


ispmarin:
I'm planning on using folium to plot a few maps but noticed that everytime I replot something folium goes  back to the internet and fetch the map data. Is there a way to configure it to download all map data and use it offline?

ocefpaf:
Not really.  The whole goal of folium is to leverage the base map using a Web Tile Service (WTS). (And in some cases even more than just the base map is downloaded.)

Maybe someone could implemente a cache system to save a certain area/zoom level, like phones do, to work offline.  A PRs with such functionality would be welcomed :wink:

ocefpaf:
BTW, if you your goal is to produce static figures, take a look at [cartopy](http://scitools.org.uk/cartopy/).

ispmarin:
Thanks @ocefpaf. I'm doing an [evaluation](https://github.com/ispmarin/maps) of the mapping options ATM for Python and being interactive is one of my requisites. I'm looking at cartopy but so far there are several problems installing it.

ocefpaf:
>  I'm looking at cartopy but so far there are several problems installing it.

Do you use [`conda`](http://conda.pydata.org/docs/)? If so try:

`conda install -c ioos cartopy`

ispmarin:
Nope, using standard pip, and it seems that [the problem](https://github.com/SciTools/cartopy/issues/726) is something with the underlying libraries. It installs fine.

linwoodc3:
Is this a feature that's coming or has it been dropped?

ocefpaf:
> Is this a feature that's coming or has it been dropped?

Folium is a leaflet wrapper. Leaflet render map tiles from a server. Unless you set a local server for your tiles there is no way to display slippy maps without an internet connection and a map server.

linwoodc3:
There is a way to "hack" folium to work offline.  I used what's described below and was able to get `folium.Map()` working and rendering a geojson of a country without being online.

How to do it:
*  You will need to change the `_default_css` and `_default_jss` links (lines 225-257) to hit something stored on you local network.  The lines are in the `element.py` file. You'll need to run a few steps

    *  Download copies of all the files found in lines `225-257` of `element.py`
        *  For example, here is [the js markercluster](https://cdnjs.cloudflare.com/ajax/libs/leaflet.markercluster/0.4.0/leaflet.markercluster.js)
        *  You would simply save this file to your local computer (or get it on your network) and accessible

[...]

PR #1312, opened 2020-04-26
Refactor hardcoded JS CDN links to _protected variables

and-viceversa:
`folium.py` exposes `_default_js` and `_default_css` at a high level. This is good, because you can override the CDN URL to point elsewhere. In my case, to point to a static assets folder.

The plugins do not do this. This PR moves the URLs out of their respective `render()` methods to expose them in a similar way. This shouldn't change any existing behavior.

Also, 98% of the changes are fixing `flake8` results. Sorry about that but I went by the contributing guideline.

### DIFF
--- a/folium/plugins/timestamped_geo_json.py
+++ b/folium/plugins/timestamped_geo_json.py
@@ -5,29 +5,9 @@ import json
 from branca.element import CssLink, Figure, JavascriptLink, MacroElement
 
 from folium.folium import Map
-from folium.utilities import get_bounds, parse_options
+from folium.utilities import parse_options, get_bounds
 
 from jinja2 import Template
-
-_default_js = [
-    ('jquery2.0.0',
-     'https://cdnjs.cloudflare.com/ajax/libs/jquery/2.0.0/jquery.min.js'),
-    ('jqueryui1.10.2',
-     'https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.10.2/jquery-ui.min.js'),
-    ('iso8601',
-     'https://rawcdn.githack.com/nezasa/iso8601-js-period/master/iso8601.min.js'),
-    ('leaflet.timedimension',
-     'https://rawcdn.githack.com/socib/Leaflet.TimeDimension/master/dist/leaflet.timedimension.min.js'),    # noqa
-    ('moment',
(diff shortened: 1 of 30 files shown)


Conengmo:
Hi @and-viceversa, thanks for your PR, I think this is a good idea!

For anyone reading along, this was discussed in #351 and more recently in #1312.

Two things:

I'd prefer if each module had the same variable name for the JS and css links. Since we already have `_default_js` and `_default_css` in `folium.py` we can use those names.

Can you remove the style changes in code you otherwise wouldn't touch? It makes reviewing a lot more work, and since I'm doing this in my free time that's something to look out for. I'll update the contributing guide to reflect this as well.

and-viceversa:
Please see updated PR.

The plugins that use them now follow the `_default_js` and `_default_css` pattern.

The `render()` methods use the same `for` loop as `folium.py`.

Most plugins only use 1 JS or CSS file, so this is a bit overkill. But now it's uniform.

Thanks for your quick feedback. Let me know what else I can do on this.

Conengmo:
Looks good @and-viceversa ! I'll check in more detail before merging, but at this time there's nothing else for you to do. Don't worry about the commits, we squash and merge.

folium/plugins/timestamped_geo_json.py
```diff
@@ -9,6 +9,29 @@
...
+]
+
+_default_css = [
+    ('highlight.js_css',
+     'https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.4/styles/default.min.css'),
+    ('leaflet.timedimension_css',
+     'https://rawcdn.githack.com/socib/Leaflet.TimeDimension/master/dist/leaflet.timedimension.control.min.css'),
+    # noqa
+    ('bootstrap_theme_css',
+     'https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap-theme.min.css')
```
Conengmo, line 32: I don't see this one in the plugin. It's in `folium.py` though. Maybe a copy-paste error?

and-viceversa: Yeah looks like simple oversight on my part. Deleted Bootstrap CSS theme.

Conengmo:
Thanks for your work on this @and-viceversa !

PR #1322, opened 2020-05-07
condition for PyPI deployment should be tarball

ocefpaf:
Deployment did not work in the last release b/c of this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,4 +126,4 @@ deploy:
     repo: python-visualization/folium
     tags: true
     all_branches: master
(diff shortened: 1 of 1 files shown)


Conengmo:
Seems to work now: https://pypi.org/project/folium/#history

ocefpaf:
I just made that one. We will know if this worked when we issue a new release b/c the deployment task should run only when we issue a new tag.

Conengmo:
Okay, that will be a surprise for next time then.

PR #1323, opened 2020-05-08
add control_geocoder plugin

WBP20:
Add a leaflet plugin to search for cities by geocoding through OSM/Nominatim


Conengmo:
Great. Don't worry about stickler, I'll get those.

Conengmo:
@WBP20 I made some changes. Can you check if everything still works for you? Note that I renamed the class, you'll find it at `folium.plugins.Geocoder`.

PR #1325, opened 2020-05-10
added reference to free Leaflet tile providers

harabat:
Added a reference to the list of free tile providers available at http://leaflet-extras.github.io/leaflet-providers/preview/, with the advice to check terms and conditions. Only changed the docstring in folium.py and raster-layers.py. This should hopefully avoid hours and hours of research to those who are not familiar with folium, map charts, tile layers, etc.


Conengmo:
Awesome, thanks @harabat !

harabat:
Cheers :)

PR #1327, opened 2020-05-10
added ref to templates available in folium/templates/tiles to docstring

harabat:
The existing docstrings in for folium.Map and folium.TileLayer only specify some of the templates that are already available in folium/templates/tiles. I changed the docstrings in folium.py and raster-layers.py to fix that.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -75,12 +75,13 @@ class Map(MacroElement):
     to Folium. Pass any of the following to the "tiles" keyword:
 
         - "OpenStreetMap"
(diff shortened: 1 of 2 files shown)


Conengmo:
We are moving away from hosting references to these tilesets ourselves, instead relying on websites like http://leaflet-extras.github.io/leaflet-providers/preview/ and packages like https://github.com/geopandas/xyzservices. Our list of tilesets needs an update, but more to remove the paid tilesets we no longer support.

Issue #1318, opened 2020-04-30
Request Adding Leaflet.TileLayer.MBTiles Plugin Support


and-viceversa:
**Is your feature request related to a problem? Please describe.**
No. This feature request adds support for displaying raster tile base maps in .mbtiles format.

**Describe the solution you'd like**
1. Make [Leaflet.TileLayer.MBTiles](https://gitlab.com/IvanSanchez/Leaflet.TileLayer.MBTiles#) it's own Folium plugin.
2. Put the above plugin on a CDN so users aren't forced to store it locally. The original author's CDN link is broken I think.

Currently, you can use Folium and [Leaflet.TileLayer.MBTiles](https://gitlab.com/IvanSanchez/Leaflet.TileLayer.MBTiles#) to display .mbtiles format, but it's a bit hacky.

I used roughly the following snippet to override `TileLayer._template`

```
    # override defaults to use the plugin
    folium.raster_layers.TileLayer._template = Template(u"""
        {% macro script(this, kwargs) %}
            var {{ this.get_name() }} = L.tileLayer.mbTiles(         <--------
                {{ this.tiles|tojson }},
                {{ this.options|tojson }}
            ).addTo({{ this._parent.get_name() }});
        {% endmacro %}
        """)

   # make a Map with .mbtile basemap 
   m = folium.Map(
        location=[35.650787, -117.661728], 
        tiles='my_tiles.mbtiles', 
        attr=attr
    )

    # add open street map
    layer = folium.TileLayer(
        tiles='https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
        name='OpenStreetMap Online',
        attr='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> '
             'contributors',
        overlay=True,
        control=True,
        show=False
    ).add_to(m)

    # function to return _template back to default after the fact.
    set_tile_layer_default(layer)
```

Even then, you have to override `TileLayer._template` back to it's default before `m.save('index.html')` for each additional (non-mbtiles) basemap you add.

It would be really nice to:

```
from folium.plugins import MBTiles

m = MBTiles.Map(
    location = [14.0, 15.0],
    tiles = 'your_tiles_url.mbtiles',
    attr = 'attr text'
)
```

**Describe alternatives you've considered**
The only alternative I can see is something like the above snippet. There was some discussion in #351.

**Additional context**
[...]

PR #1341, opened 2020-05-26
Plugin mbtiles

and-viceversa:
PR for issue #1318 to add support for MBTiles.

@Conengmo is correct that the original plugin repo [here](https://www.npmjs.com/package/Leaflet.TileLayer.MBTiles) appears abandoned. However, MBTiles format under the hood is SQLite. The plugin just reads and returns tiles from the Mapbox-spec database. I think it's unlikely that either will change enough to break the plugin. Evidence is that this code still works 4 years later.

The good news is that this plugin is now being maintained [here](https://github.com/0nza1101/leaflet-tilelayer-mbtiles-ts), last updated May 2020. 

I think it'd be good to use the updated repo, but it's structure is beyond my web dev understanding.

Let me know what I can improve here.

### DIFF
--- a/folium/plugins/mbtiles.py
+++ b/folium/plugins/mbtiles.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+from branca.element import Figure, JavascriptLink
+from jinja2 import Template
+
+from folium.raster_layers import TileLayer
+
+_default_js = _default_js = [
+    ('sql-js',
+     'https://unpkg.com/sql.js@0.3.2/js/sql.js'),
+    ('MBTiles',
+     'https://unpkg.com/leaflet.tilelayer.mbtiles@latest/Leaflet.TileLayer.MBTiles.js')
+    ]
+
+
+class MBTiles(TileLayer):
+    """
+    Class to display raster tiles in MBTiles format.
+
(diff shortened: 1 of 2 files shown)


folium/plugins/mbtiles.py
```diff
@@ -0,0 +1,64 @@
...
+]
+
+
+class MBTiles(TileLayer):
+    """
+    Class to display raster tiles in MBTiles format.
+
+    See https://wiki.openstreetmap.org/wiki/MBTiles for overview.
+    See https://www.npmjs.com/package/Leaflet.TileLayer.MBTiles for original Leaflet plugin.
+
```
Conengmo, line 22: Maybe add a line that this class accepts any `TileLayer` argument.

Conengmo:
This PR is quite stale, so I'll close it now. If somebody wants to pick this up and address the open comments they are welcome to!

prusswan:
with the availability of PMTiles format, there is a "better" way (especially for large raster tileset):

1.  convert MBTiles to PMTiles (see https://docs.protomaps.com/pmtiles/create)

2.  use folium-pmtiles plugin: https://github.com/jtmiclat/folium-pmtiles

PR #1342, opened 2020-05-26
Draw export button refactor

and-viceversa:
This PR is similar to the scope change in #1312.

`draw.py` has an optional export button. As far as I can tell, there is no way to override its hardcoded `<style>`. Move `export_style` variable outside the Draw class and add Pythonic protected underscore. This shouldn't change default or expected behavior.

Plugins seem more open to interpretation than the core library ... is there any guidance to avoid this sort of refactor in the first place? To me, this sort of property could also belong to `self`.


Conengmo:
Good that you're taking this on, that export button is weird. 

- I agree the styling for the button should be editable. I don't think it should be a module variable. As you said, putting it in `self` seems like a better idea to me.
- Better default styling of the export button would be good. It now has an absolute position, that may fall under the layer control. But note that this aspect is not a requirement for getting this PR merged.

About guidance on plugin development: folium has grown organically over the years with different maintainers, so that may explain the different styles in the code.

PR #1349, opened 2020-06-05
Geojson zoom on click fixes

### LINKED ISSUE #1162: GeoJson make zoom on click optional
The default behavior when clicking on a region (e.g. a state) appears to have changed in version 0.9.1 compared to 0.8.3. On click in 0.9.1, the map zooms in close to the region. In 0.8.3, there is no zoom on click. This can be replicated using the choropleth sample code from the quickstart: https://python-visualization.github.io/folium/quickstart.html.

```python
import folium
import pandas as pd


url = 'https://raw.githubusercontent.com/python-visualization/folium/master/examples/data'
state_geo = f'{url}/us-states.json'
state_unemployment = f'{url}/US_Unemployment_Oct2012.csv'
state_data = pd.read_csv(state_unemployment)

m = folium.Map(location=[48, -102], zoom_start=3)

folium.Choropleth(
    geo_data=state_geo,
    name='choropleth',
    data=state_data,
    columns=['State', 'Unemployment'],
    key_on='feature.id',
    fill_color='YlGn',
    fill_opacity=0.7,
    line_opacity=0.2,
    legend_name='Unemployment Rate (%)'
).add_to(m)

folium.LayerControl().add_to(m)

m
```
This occurs for folium.GeoJson as well. Is there an option to disable this zoom?

### LINKED ISSUE #1345: GeoJsonPopup doesn't work (v 0.11.0) - TypeError: e.target.getBounds is not a function
Hey guys,

I try to use the GeoJsonPopup module to display names etc. When I run the code, no error is displayed in the shell, but when I open the html file and click on a marker, no popup show up and in the console there is following error:
```
Uncaught TypeError: e.target.getBounds is not a function
    at i.click (geojson_test.html:85)
    at i.fire (leaflet.js:5)
    at i._fireDOMEvent (leaflet.js:5)
    at i._handleDOMEvent (leaflet.js:5)
    at HTMLDivElement.s (leaflet.js:5)
```
here is a snippet of my code:

```
import folium
import pandas
import geopandas

tileset= "https://api.mapbox.com/styles/v1/f---/cj53iitnb0cxr2rp96gsee7dx/tiles/256/{z}/{x}/{y}?access_token=pk.eyJ1IjoiZjJkYXoiLCJhIjoiY2o1M2hrczV0MDc1bDJ2bjBiYWdha2c0diJ9.uEHH_ty------"
m = folium.Map(location=[50.032,10.519], zoom_start=14, tiles=tileset, attr="Mapbox")

data = geopandas.read_file('stationen.json', driver='GeoJSON')

stationen = folium.GeoJson(
    data,
    name='Stationen',
    popup = folium.GeoJsonPopup(
        fields=['shortname', 'voltage_level'],
        aliases=['Name', 'Spannungsebene'],
    )
).add_to(m)

m.save("geojson_test.html")
```

**Environment**
 - Browser: chrome
 - html file
 - folium version 0.11.0

Thanks in advance ;)

Conengmo:
Close #1345 and close #1162 

Apply two changes to the GeoJson class:

Don't zoom on click if the geometry type doesn't support it. This is the case for point geometries.

Add a parameter to control the zoom on click functionality. Disable it by default. I thought about this and I think it's best not to hijack the zoom by default. Users who do want that behavior can enable it as follows:

`GeoJson(zoom_on_click=True)`

or

```
choropleth = Choropleth()
choropleth.geojson.zoom_on_click = True
```


Conengmo:
Thanks for the review Jason!

> maybe a good thing would be to add the parameter usage to some of the notebook examples

you're totally right, I'll add that

folium/features.py
```diff
@@ -409,9 +411,13 @@ class GeoJson(Layer):
                     e.target.setStyle({{ this.get_name() }}_highlighter(e.target.feature));
                 },
                 {%- endif %}
+                {%- if this.zoom_on_click %}
                 click: function(e) {
-                    {{ this.parent_map.get_name() }}.fitBounds(e.target.getBounds());
+                    if (typeof e.target.getBounds === 'function') {
+                        {{ this.parent_map.get_name() }}.fitBounds(e.target.getBounds());
+                    }
```
jtbaker, line 418: Here, how about to handle zooming for point geometries as well? I'm not sure about best generalized options for the `.fitBounds()` [method](https://leafletjs.com/reference-1.6.0.html#fitbounds-options), it supports both `padding` and `maxZoom` options.
```
else {
    const bounds = L.latLngBounds([e.target.getLatLng()])
    {{ this.parent_map.get_name() }}.fitBounds(bounds, {padding: [5,5]});
}
```

Conengmo: I think your code is fine but I'd rather not introduce more fragility here. This may work for point geometries but I'm unsure what new bugs it might introduce. I'll just merge the PR as is, let's consider any new functionality separately.

PR #1409, opened 2020-11-09
Release via GH actions

ocefpaf:
@Conengmo this should be more robust than the Travis-CI deployment we had. All one need to do to release both wheels and sdist is to create a tag on GH.


ocefpaf:
Let's wait for #1410 first.

Conengmo (approved):
LGTM. Let's try this out next time!

PR #1411, opened 2020-11-09
py2k cleanup

ocefpaf:
@Conengmo this kinds of touch all the files in folium but any emerging conflicts should be easy to fix.


ocefpaf:
Ripping the bandaid on this one @Conengmo. Hope it is already :smile:

PR #1412, opened 2020-11-09
GH actions for tests

ocefpaf:
@Conengmo this is far from ready and needs a lot of work but I'm putting it here as a draft so I don't forget. Mostly we need to move away from Travis-CI. The .org domain is dying, only @wrobstory has the credentials to move to .com, but even so we'll be a bit limited with the new OSS Travis rules.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,60 @@
+name: Tests
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  run:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


ocefpaf:
@Conengmo actually, let's merge this so I can work on the next iteration with the GH actions enabled. It will be easier to debug it here than on my fork.

wrobstory:
👋 let me know if there's anything I need to do!

ocefpaf:
>  let me know if there's anything I need to do!

If you could move us to Travis-CI .com it would help until we make the transition to GH Actions. It should be just a click here: https://travis-ci.com/python-visualization/folium

wrobstory:
Arrrggggh `This repository may not exist or you may need to use another account to have access to it`, this is via signing in with my github account. I dont think you can sign in with an email anymore, so I'm not sure what to do.

ocefpaf:
No problem. Let's just focus on the GH Actions move. Thanks!

PR #1413, opened 2020-11-09
spellcheck

ocefpaf:
The failures here are related to things that will be removed in #1410, so no need to fix them.


Conengmo (approved):

tests/test_jinja.py
```diff
@@ -19,7 +19,7 @@
     ([[0, 0], [1, 1]], '[[0, 0], [1, 1]]'),
     ([(0, 0), (1, 1)], '[[0, 0], [1, 1]]'),
     ({'hi': 'there'}, '{"hi": "there"}'),
-    ({'hi': {'there': 1, 'whats': 'up'}}, '{"hi": {"there": 1, "whats": "up"}}'),
+    ({'hi': {'there': 1, "what's": 'up'}}, '{"hi": {"there": 1, "what's": "up"}}'),
```
ocefpaf, line 22: This one is failing. I'll check ASAP.

ocefpaf:
@Conengmo the spellcheck won't pass b/c there are false positives. Codespell is inspecting the notebook outputs and they contain all sort of strings in the embed images and other outputs.

When we start building a gallery from "clean" notebooks this will pass. Still, I believe that there is value in merging now so we can check new code for spell errors.

ocefpaf:
@Conengmo can you take a second look at this one? I removed all the outputs from all the notebooks. They should never be there in the first place. I removed the GHA check for the pre-commit and replaced it with pre-commit CI instead. Faster and it will update the pre-commit-ci file for us.

I believe that the examples should be run on binder or locally, so removing the outputs not only make the diffs more meaningful and the repo easier to manage but also "force" folks to run them first before copying and pasting from them.

Conengmo:
> I removed all the outputs from all the notebooks

Doesn't that mean the nbviewer links will be disabled as well? I don't think that's acceptable, as examples they should be easy to view, without having to wait to spin up a binder.

> so removing the outputs not only make the diffs more meaningful and the repo easier to manage 

that's true, Notebook diffs are terrible. We could also solve that by having a 'source' folder without rendering, and a 'main' folder that contains copies that get rendered on releases.

Or, I'd rather have to hard-to-use diffs then break our nbviewer examples.

> but also "force" folks to run them first before copying and pasting from them.

I don't think that's a benefit.

ocefpaf:
> We could also solve that by having a 'source' folder without rendering, and a 'main' folder that contains copies that get rendered on releases.

That is probably be best option. A while back I wanted to move all examples to another repo so we are not held accountable for maintaining them but I guess that ship has sailed.

Let's move this one to a draft again and I'll try to work on that solution.

Conengmo:
Superseded by https://github.com/python-visualization/folium/pull/1652

PR #1416, opened 2020-11-10
Move JS/CSS links to class attributes

Conengmo:
I wanted to do this before we release. We recently moved the links to external JS and CSS resources to module variables: https://github.com/python-visualization/folium/pull/1312. This was a good thing. But I was thinking that it would make things much more DRY if we move them to the classes instead, because then we can render them in one place.

I made a mixin class that renders JS and CSS links. I thought about subclassing `MacroElement` and using that subclass throughout the code, but I think using a mixin is a bit more clean here.

This change allows us to remove duplicated code while still allowing users to replace the JS/CSS links.

### DIFF
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -1,0 +1,21 @@
+from branca.element import Figure, Element, JavascriptLink, CssLink
+
+
+class JSCSSMixin(Element):
+    """Render links to external Javascript and CSS resources."""
+
+    default_js = []
+    default_css = []
+
+    def render(self, **kwargs):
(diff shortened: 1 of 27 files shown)


folium/elements.py
```diff
@@ -0,0 +1,21 @@
+from branca.element import Figure, Element, JavascriptLink, CssLink
+
+
+class JSCSSMixin(Element):
```
ocefpaf, line 4: Thanks! This reduces a lot of code duplication!

Issue #1424, opened 2020-12-09
Highlight `__geo_interface__` support in Documentation


kai-tub:
**Is your feature request related to a problem? Please describe.**
I was not really aware that `folium` supports to display objects with a `__geo_interface__` protocol until today.
I am relatively new to working with geographic data, so maybe it is obvious for other people,
but I was surprised that it worked without any issues. 

**Describe the solution you'd like**
A note in the documentation that the `__geo_interface__` protocol is supported with `folium.features.GeoJson`, which allows for extensive interoperability with various other Python libraries.

**Implementation**
I am not sure where the best place would be to highlight this feature. Maybe as a note in `Quickstart` with example usage of `GeoPandas`? And a note in the `GeoJson` documentation?

It is an awesome feature, but I feel like it could be easily overlooked (at least from my point of view as a beginner).

PS: Thanks for maintaining this awesome project! 👍

Conengmo:
I think the docstring of `GeoJson` might be a good place for this: https://github.com/python-visualization/folium/blob/master/folium/features.py#L314

Do you want to make a PR?

kai-tub:
I would be happy to :)

kai-tub:
Should we also add an example?
Like loading from shapely or geopandas? 
These seem to be the most prominent `__geo_interface__` libraries that could be frequently combined for analysis with `folium`. (Again from a perspective of a newcomer to this field)

Conengmo:
We already have this example notebook, would this be similar to what you are proposing: https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/Geopandas_and_geo_interface.ipynb

kai-tub:
Oh, I missed that example.
Ok, then I will only update the docstring

kai-tub:
As a side-note:
I think that these examples should be more prominently displayed.
I was always looking at the documentation page and not the GitHub Readme and I feel like a link at the end of the *Quickstart* section can be easily overlooked. Maybe we could add an extra section to the tables of content, where there are links to all current examples? Then new users could more easily find starting points for their problems?

Just my 2cts :)

Conengmo:
that's also a good idea. It has bothered me for a while that the example notebooks are not well ordered and have sometimes cryptic names. But I don't want to change the names, because there may be incoming links to them. Linking to them from the documentation sounds like a good way around this.

kai-tub:
Sure. :)
I will open both PRs later today.

[...]

PR #1425, opened 2020-12-09
Update GeoJson doc-string

kai-tub:
As discussed in #1424


kai-tub:
If desired, I can also add my updates to the documentation as discussed in #1424 to this branch.
Otherwise, I will open a new PR.

Conengmo:
Thanks! If you still want to help out with adding links to the example notebooks from the documentation, that would be great. Part of it would be thinking about how we could best do this. I'm thinking now to add a new file to `docs/` with links to the example notebooks, each with a one/two line description of whats in there. We could group them by topic as well. But I'm open for suggestions. Maybe before writing the whole thing, you can already open a draft PR to talk about.

Issue #212, opened 2015-10-09
get filled color for polygon_marker


PBrockmann:
I cannot get a flat filled color when I use polygon_marker. Instead I get a gradient that comes from Leaflet-DVF I suppose.
But many times, I would prefer to display a flat color.

Another point, I have noticed on the use of the property fill_color from polygon_marker is that the filled color is not displayed when you embed your map in a notebook.

Here is an example tested with 0.1.5

```python
import folium
print folium.__version__

from IPython.display import HTML, display

def inline_map(m):
    if isinstance(m, folium.folium.Map):
        m._build_map()
        srcdoc = m.HTML.replace('"', '&quot;')
        embed = HTML('<iframe srcdoc="{}" '
                 'style="width: {}px; height: {}px; '
                 'border: none"></iframe>'.format(srcdoc, m.width, m.height))
    else:
        raise ValueError('{!r} is not a folium Map instance.')
    return embed
```

```python
map_5 = folium.Map(width=600, height=400, location=[45.5236, -122.6750], zoom_start=13)

map_5.polygon_marker(location=[45.5318, -122.6745], popup='Broadway Bridge',
                     line_color='red',
                     fill_color='blue', num_sides=8, radius=10, fill_opacity=1)

inline_map(map_5)
``` 
![folium_02](https://cloud.githubusercontent.com/assets/5402758/10397222/c979ccc6-6ea6-11e5-922a-7976807107ef.png)

Inspecting the SVG I find a 
fill="url(#gradeac640f5-53f0-6c37-ea2b-10d33cbc9aa0)"
whereas in a saved html with map_5.create_map(path="myfile.html"), I get a 
fill='blue' as expected (event if it is a gradient that is finally displayed).

![folium_03](https://cloud.githubusercontent.com/assets/5402758/10397247/e63c8006-6ea6-11e5-87c4-461d93ac4bf8.png)

PBrockmann:
OK I get a solution after reading and testing https://github.com/python-visualization/folium/issues/135
on "How to use folium with iPython / Jupyter notebook".

```python
from IPython.display import HTML, display

def inline_map(m):
    if isinstance(m, folium.folium.Map):
        m._build_map()
        srcdoc = "data:text/html;base64," + m.HTML.encode('base64')
        embed = HTML('<iframe src="{}" '
                 'style="width: {}px; height: {}px; '
                 'border: none"></iframe>'.format(srcdoc, m.width, m.height))
    else:
[...]

PR #1433, opened 2021-01-04
add a gradient option to path_options

nocturnalAndroid:
Leaflet-dvf adds a gradient to RegularPolygonMarker by default, this change allows turning that gradient off (and also applying it to other elements like the CircleMarker which does not get it by default for some reason)
The only other reference I could find to this issue was here: https://github.com/python-visualization/folium/issues/212
Though it looks like the problem raised in 2015 is not the same one referred to by the last two comments from 2020.

In any case, I've applied this fix to my local install of folium and thought I might as well submit it as a pull request, this is my first ever pull request to a public project so please let me know if I've done something wrong ¯\_(ツ)_/¯


Conengmo:
Thanks for your PR @nocturnalAndroid! We were a bit slow with processing it, but better late than never :)

PR #1435, opened 2021-01-11
added optional inital position

jjbenes:
The slider in TimeSliderChoropleth is always 0, but in some apps, I want to show the latest data first. Users can then view older data by moving the slider to the left if they want to.

### DIFF
--- a/folium/plugins/time_slider_choropleth.py
+++ b/folium/plugins/time_slider_choropleth.py
@@ -26,13 +26,21 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
         Whether the Layer will be included in LayerControls.
     show: bool, default True
         Whether the layer will be shown on opening (only for overlays).
-
+    init_timestamp: int, default 0
+        Initial time-stamp index on the slider. Must be in the range
+        `[-L, L-1]`, where `L` is the maximum number of time stamps in
+        `styledict`. For example, use `-1` to initialize the slider to the
+        latest timestamp.
     """
     _template = Template(u"""
         {% macro script(this, kwargs) %}
             var timestamps = {{ this.timestamps|tojson }};
             var styledict = {{ this.styledict|tojson }};
-            var current_timestamp = timestamps[0];
+            {% if this.init_timestamp >= 0 %}
+            var current_timestamp = timestamps[{{ this.init_timestamp }}];
+            {% else %}
+            var current_timestamp = timestamps[timestamps.length+{{ this.init_timestamp }}];
+            {% endif %}
 
             // insert time slider
             d3.select("body").insert("p", ":first-child").append("input")
@@ -40,7 +48,7 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
                 .attr("width", "100px")
                 .attr("min", 0)
                 .attr("max", timestamps.length - 1)
-                .attr("value", 0)
+                .attr("value", current_timestamp)
                 .attr("id", "slider")
                 .attr("step", "1")
                 .style('align', 'center');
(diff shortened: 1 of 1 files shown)


folium/plugins/time_slider_choropleth.py
```diff
@@ -26,21 +26,29 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
...
+        `[-L, L-1]`, where `L` is the maximum number of time stamps in
+        `styledict`. For example, use `-1` to initialize the slider to the
+        latest timestamp.
     """
     _template = Template(u"""
         {% macro script(this, kwargs) %}
             var timestamps = {{ this.timestamps|tojson }};
             var styledict = {{ this.styledict|tojson }};
-            var current_timestamp = timestamps[0];
+            {% if this.init_timestamp >= 0 %}
```
Conengmo, line 39: Does it make sense to do this logic in Python instead? We don't need an if-statement to index `timestamps` there.

folium/plugins/time_slider_choropleth.py
```diff
@@ -156,3 +164,14 @@ def __init__(self, data, styledict, name=None, overlay=True, control=True,
 
         self.timestamps = timestamps
         self.styledict = styledict
+        if init_timestamp >= 0:
```
Conengmo, line 167: These error messages are very nice. I'm thinking though, could you combine them into a single check? It's less verbose that way, but will do the same. Just check if the value is between the expected lower and higher values.

Conengmo (changes requested):
It's clear to me what you want to do, and I think we can incorporate this. I added two minor comments to possibly make the change a bit simpler.

folium/plugins/time_slider_choropleth.py
```diff
@@ -26,21 +26,25 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
         Whether the Layer will be included in LayerControls.
     show: bool, default True
         Whether the layer will be shown on opening (only for overlays).
-
+    init_timestamp: int, default 0
```
Conengmo, line 29: I find this name a bit confusing. It suggests it requires a timestamp, but instead it's an index. Could we rename it to something more self-explanatory? Like `initial_index` or `initial_timestamp_index`.

folium/plugins/time_slider_choropleth.py
```diff
@@ -26,21 +26,25 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
...
+        `styledict`. For example, use `-1` to initialize the slider to the
+        latest timestamp.
     """
     _template = Template(u"""
         {% macro script(this, kwargs) %}
             var timestamps = {{ this.timestamps|tojson }};
             var styledict = {{ this.styledict|tojson }};
-            var current_timestamp = timestamps[0];
-
+            var current_timestamp = timestamps[
```
Conengmo, line 39: Shouldn't this be `var current_timestamp = timestamps[{{ this.init_timestamp }}];`?

folium/plugins/time_slider_choropleth.py
```diff
@@ -26,21 +26,25 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
...
+                (timestamps.length+{{ this.init_timestamp}}) % timestamps.length];
             // insert time slider
             d3.select("body").insert("p", ":first-child").append("input")
                 .attr("type", "range")
                 .attr("width", "100px")
                 .attr("min", 0)
                 .attr("max", timestamps.length - 1)
-                .attr("value", 0)
+                .attr("value", current_timestamp)
```
Conengmo, line 47: The `"value"` here is the index if I'm not mistaken

```suggestion
                .attr("value", {{ this.init_timestamp }})
```

Conengmo (changes requested):

Conengmo:
Updated the PR so it works on my end and added an entry to the example notebook.

Conengmo:
Thanks for your PR @jjbenes!

Issue #1438, opened 2021-01-14
Curious folium version 0.0.0 messes up checks for version requirements


niklasbohn:
When executing software tests on azure I run into trouble when checking for the installed folium version. If a package requires a particular folium version I get an error message like that:

`+ pip check`
`arosics 1.2.4 has requirement folium>=0.6.0, but you have folium 0.0.0.`

I just checked the folium [release history on PyPi](https://pypi.org/project/folium/#history) and it indicates that you uploaded folium version 0.12.0 to PyPi on Jan 6, 2021. But if I scroll down to the bottom of the release history, I recognize that the curious folium version 0.0.0 was also uploaded to PyPi on Jan 6, 2021. Would it be possible that you just misleadingly uploaded a folium version 0.0.0? If yes, could you check that and clean up if necessary?

Thanks in advance for any help!

Conengmo:
Thanks for the report! That doesn't look right indeed. @ocefpaf can you see what happened?

ocefpaf:
Probably a mistake in the release :-(

Working on it...

ocefpaf:
Looks like the wheel is fine:

```
> pip install folium
Collecting folium
  Downloading folium-0.12.0-py2.py3-none-any.whl (94 kB)

> python -c "import folium; print(folium.__version__)"0.12.0

```

But the sdist is not:

```
> pip install folium --no-binary :all:                                                                                                                                                                             ⏎
Collecting folium
  Downloading folium-0.12.0.tar.gz (92 kB)
     |████████████████████████████████| 92 kB 382 kB/s 
ERROR: Requested folium from https://files.pythonhosted.org/packages/17/fb/a36b985ccc1c1a7dbd2b92e82578bfeb7627622f986c8d8232df090c5075/folium-0.12.0.tar.gz#sha256=d45ace0a813ae65f202ce0356eb29c40a5e8fde071e4d6b5be0a89587ebaeab2 has different version in metadata: '0.0.0'
```

It is probably some change in setuptools_scm, maybe we need to add the version.py file in the sdist now.

niklasbohn:
@ocefpaf Thanks for dealing with this issue! However, the folium release history on PyPi hasn't changed after pull request #1439 and I still get the same error message as before on azure. I guess you would have to delete the curious version 0.0.0 on PyPi or alternatively, upload a new version so that there aren't two versions released on the same date anymore.

ocefpaf:
[...]

PR #1439, opened 2021-01-14
needed this for sdist

ocefpaf:
Should fix https://github.com/python-visualization/folium/issues/1438


Conengmo (approved):
I don’t know enough about the specifics here to really comment, but I’m on board with trying this

PR #1441, opened 2021-01-18
add a check version

ocefpaf:
@Conengmo this is just to debug problems in case we have a regression on #1438


ocefpaf:
Merging without the tests b/c this will only work on releases.

PR #1450, opened 2021-02-08
Update setting of CRS  for geopandas plots

Sangarshanan:
The API has been updated as on https://geopandas.org/projections.html
Also added a small test for the change

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -538,7 +538,7 @@ class GeoJson(Layer):
         elif hasattr(data, '__geo_interface__'):
             self.embed = True
             if hasattr(data, 'to_crs'):
-                data = data.to_crs('EPSG:4326')
+                data.crs = 'EPSG:4326'
             return json.loads(json.dumps(data.__geo_interface__))
         else:
             raise ValueError('Cannot render objects with any missing geometries'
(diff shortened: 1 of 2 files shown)


folium/features.py
```diff
@@ -538,7 +538,7 @@ def process_data(self, data):
         elif hasattr(data, '__geo_interface__'):
             self.embed = True
             if hasattr(data, 'to_crs'):
-                data = data.to_crs('EPSG:4326')
+                data.crs = 'EPSG:4326'
```
martinfleis, line 541: This changes the behaviour of the function. `folium` requires geometry in 4326 which is ensured by the `to_crs` call. This change overrides the CRS without reprojection.

The existing code seems to be perfectly fine and in line with the current geopandas API.

The only case where this may cause trouble is a GeoDataFrame with `None` as a CRS. In such a case `to_crs` will raise a warning and could be skipped here (assuming the geometries are in the correct CRS).

ocefpaf: > The only case where this may cause trouble is a GeoDataFrame with None as a CRS. In such a case to_crs will raise a warning and could be skipped here (assuming the geometries are in the correct CRS).

IMO that should fail and the user should add a crs explicitly to avoid implicit assumptions.

Sangarshanan: Thanks so much for the clarification! I will close this then

PR #1458, opened 2021-03-08
Expose GeoJson's web retrieval to its own function

beautah:
…so it can be overrriden or subclassed readily


beautah:
whoops, happy to re-submit if you'd like it cleaner, but feel free to delete the whitespace and merge too

PR #1459, opened 2021-03-17
fix: return rendered HTML from `Map.render`

paulswartz:


Conengmo (changes requested):
This is not enough to get the desired behavior. Both parents of `Map` don't return anything in their `render` methods, so `render` on `Map` will still return None.

paulswartz:
python-visualization/branca#106 has the fixes for the parent elements to return the HTML from their `render()` methods.

Conengmo:
With how the structure of Element, MacroElement, Map and Figure is built getting the html output is not as simple as slapping on some returns on the various `render` functions. 

If you want the html of a map, you have to render the `Figure` the map creates internally:

```
m = Map()
m.get_root().render()
```

Maybe to address the implicit desired functionally of this PR we could make a shortcut method for this on the `Map` class?

I'm closing this PR for now. If somebody wants to revisit, please first discuss what you'd like to achieve.

PR #1492, opened 2021-08-03
Polyline documentation

IamPhytan:
Added documentation about Polylines inside `Quickstart`, so that Polylines are also visible when starting using Folium

Note that I also added the following to `.gitignore`:
* VSCode's `.vscode` 
*  `venv/`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ miniconda.sh
 latest_logs/
 *.nbconvert.ipynb
 folium/_version.py
+.vscode/
+venv/


Conengmo:
the change on nbviewer: https://nbviewer.org/github/IamPhytan/folium/blob/polyline-tutorial/examples/Quickstart.ipynb

Conengmo (approved):

Conengmo:
Thanks for this addition @IamPhytan! We're a bit late, but better now than never :)

IamPhytan:
You're welcome ! Thanks for the merge and for maintaining `folium` !

PR #1498, opened 2021-08-07
ENH: support xyzservices

### LINKED ISSUE #1497: Support xyzservices.TileProvider as an input of tiles

With the @geopandas team, we have built a new package called [`xyzservices`](https://github.com/geopandas/xyzservices) to serve as a unified place for XYZ tile management. The idea behind this is that other packages will depend on that instead of having their own system and list of supported tiles hard-coded. It would be great to have `folium` on board with the idea, by supporting `xyzservices.TileProvider` as an input of `tiles`.

`xyzservices` itself does not have any dependencies which makes it a non-problematic dependency.

It is worth mentioning that `xyzservices` stores providers' metadata in JSON under `share`, accessible for anyone if you have a use case for that.

I am happy to do any changes in `xyzservices` needed to make it as useful for you as possible if that allows centralising maintenance of providers in one place.

I'd like to see the support of `TileProvider` as an input of `tiles` and `TileLayer`. 

```py
import xyzservices.providers as xyz

m = folium.Map(
    location=[53.4108, -2.9358],
    tiles=xyz.CartoDB.Positron,
)
folium.TileLayer(tiles=xyz.CartoDB.Positron).add_to(m)
```

All the info `TileLayer` require can be fetched from the `TileProvider` object under the hood.

You can pass individual attributes manually as we currently do in `xyzservices` [docs](https://xyzservices.readthedocs.io/en/latest/introduction.html#building-the-tile-url) but it can be streamlined with a built-in support.

```py
tiles = xyz.CartoDB.Positron

folium.Map(
    location=[53.4108, -2.9358],
    tiles=tiles.build_url(scale_factor="@2x"),
    attr=tiles.html_attribution,
)
```

**Additional context**

Support of `xyzservices` is being implemented across the ecosystem, from geopandas' own packages (new built-in foilum-based plotting (https://github.com/geopandas/geopandas/pull/1953) and `contextily`) to `leafmap` (https://github.com/giswqs/leafmap/issues/91), `ipyleaflet` (https://github.com/jupyter-widgets/ipyleaflet/issues/854) and holoviews (https://github.com/holoviz/holoviews/issues/5057) (+ hopefully more).

If there's anything we would need to implement in `xyzservices` to make all this work, I am all ears.

**Implementation**

I am happy to contribute this as a PR. 

In this case, it is actually quite simple and can be done without importing `xyzservices` as `TileProvider` is a subclass of dict. Something along this:

```py
if isinstance(tiles, dict):
     attr = attr if attr else tiles.html_attribution
     min_zoom = tiles.get("min_zoom", 0)
     max_zoom = tiles.get("max_zoom", 18)
     tiles = tiles.build_url()
```

martinfleis:
Closes #1497

The PR is ready, we just need to wait for xyzservices 2021.08.0 to be released as it currently requires functionality available only on main. That shouldn't take more than a couple of days.


martinfleis:
Looking at your management of built-in tiles, there's a possibility to simplify all that by relying on `xyzservices` directly.

In `TileLayer` we can map those 10 currently built-in on relevant `TileProvider` objects and get rid of all that complicated machinery you have to manage `templates/tiles`. We could even map any custom non-URL string to matching TileProvider (as "esriworldtopomap" would be mapped to `xyz.Esri.WorldTopoMap`, essentially expanding the number of built-in tiles to the whole library we have in xyzservices. That would essentially outsource the maintenance of tile sources to `xyzservices`.

It would require having `xyzservices` as a hard dependency though but keep in mind that it itself has no deps.

It is up to you. I can either expand this PR or make the refactor in another one if you'd like to have it.

martinfleis:
xyzservices 2021.08.0 is now available on PyPI so the tests should now pass.

Conengmo (approved):
Nice integration. Let's merge it when the tests pass. 

Additionally, we might want to include an example in our Quickstart example notebook or perhaps a separate notebook.

martinfleis:
> Additionally, we might want to include an example in our Quickstart example notebook or perhaps a separate notebook.

Do you want it to be a part of this PR or a follow-up?

Conengmo:
Let's do it as a follow-up, since this one is good to go!

PR #1507, opened 2021-08-29
Optimized Images

vardaan-raj:
I have simply compressed the size of the images without compromising their quality, to make them more space-efficient


Conengmo:
I don't think anybody is going to care about these few kilobytes, so I'm going to save myself the time of reviewing this PR.

PR #1511, opened 2021-09-04
Lazy popup - only load content on click

### LINKED ISSUE #1214: Prevent image caching in leaflet popup?
[Copied from SO](https://stackoverflow.com/questions/58357496/prevent-image-caching-in-leaflet-popup)

So I have a Python web service that creates maps with Folium/Leaflet.js. Each plotted point is a marker with a custom popup:

```python
folium.Marker(
            location=(row['lat'], row['lng']),
            icon=folium.Icon(color=color_mapper[type]),
            popup=folium.Popup(f'<p>{info_str} taken at {row["time"]} <br> \
                                    Description: {row["description"]}</p> \
                                    <img src={row["photoURL"]} width=300 height=500>'),
                tooltip=info_str
)
```
As of now when the generated HTML file is opened, any images in the popup will be loaded alongside the tilesets and plotted points. This has led to extremely slow loading times and excessive charges on each image loading call for each map. I there a way to override the popup's default behavior and only load the image when the popup is opened?

marciogranzotto:
Fixes #1214

https://user-images.githubusercontent.com/11080704/132108380-96b9c24d-b1f1-4c31-8264-05a5a0ba0d4b.mp4

Usage:
```python
marker = folium.Marker(
    [data['Latitude'], data['Longitude']],
    popup=folium.Popup(data['popup_html'], lazy=True),
    tooltip=tooltip,
    icon=folium.Icon(color=data['color'])
)
```


labtec901:
Bumping to express how useful I think this change is.  I published a repo today for one of my scripts and I am directing users to this open pull request to get the updated .py file because my application is prohibitively slow without this functionality.

PR #1528, opened 2021-11-12
Allow for simple Popup too

ocefpaf:
@Conengmo this used to be default but now it fails silently. This PR just add it back while keep the option for the more advanced `GeoJsonPopup`.

PR #1533, opened 2021-11-16
Test split

ocefpaf:
Do not merge this one yet! I''ll rebase against `main` after #1528 is merged to remove that commit from here.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,12 @@
 name: Publish to PyPI
 
 on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
   release:
     types:
       - published
@@ -14,7 +20,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.8
+        python-version: 3.x
 
     - name: Get tags
       run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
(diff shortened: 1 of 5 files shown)


ocefpaf:
@Conengmo do you mind taking a look? Let me know if you agree with the new test structure.

.github/workflows/publish.yml
```diff
@@ -1,6 +1,12 @@
 name: Publish to PyPI
 
 on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
Conengmo, line 9: is this correct? We want to publish to PyPI whenever there's a push or a pull request for the main branch?

.github/workflows/publish.yml
```diff
@@ -14,40 +20,34 @@ jobs:
...
-        python setup.py --version
-
     - name: Test wheels
       run: |
         cd dist && python -m pip install folium*.whl
         python -m twine check *
       shell: bash
 
     - name: Publish a Python distribution to PyPI
+      if: ${{ github.event_name == 'create' && github.event.ref_type == 'tag' }}
```
Conengmo, line 50: I suppose this means when a tag is created? Seems like this would be better suited in the `on` part.

setup.cfg
```diff
@@ -1,21 +1,11 @@
...
-parentdir_prefix =
-
 [tool:pytest]
 flake8-max-line-length = 121
 flake8-ignore =
     docs/* ALL
 select = C,E,F,W,B,B950
 ignore = E203, E501, W503
+markers =
+    web: marks tests require connection (deselect with '-m "not web"')
```
Conengmo, line 8: are we using this?

.github/workflows/publish.yml
ocefpaf: This will ensure the tarball tests here will run. We have na if clause below that will publish only when there is a tag.

.github/workflows/publish.yml
ocefpaf: Nope. We want to run all from above everywhere and this part below only on tags.

setup.cfg
ocefpaf: Not yet but I would like to sprinkle that on some tests that fetch data. Maybe not worth it? I'll let you decide :-)

Conengmo: maybe best to take that out of this PR and add it when we start using it.

.github/workflows/publish.yml
Conengmo: I guess I'm not a fan of having this switching logic outside of the `on` statement, but it's okay 👍

Conengmo (approved):

Issue #1541, opened 2021-11-23
examples on nbviewer returns 404


clausmichele:
**Describe the bug**
The link to the example notebooks is broken: https://nbviewer.org/github/python-visualization/folium/tree/master/examples/

ocefpaf:
> **Describe the bug** The link to the example notebooks is broken: https://nbviewer.org/github/python-visualization/folium/tree/master/examples/

We should update it to reflect the master to main move. Try https://nbviewer.org/github/python-visualization/folium/tree/main/examples/, it should work.

PR #1542, opened 2021-11-23
Master to main take 2

ocefpaf:
Closes https://github.com/python-visualization/folium/issues/1541


ocefpaf:
Let's merge #1533 before this one so we can make better sense of the failures.

Conengmo (approved):

PR #1552, opened 2021-12-26
GitHub actions badges

oefe:
Replace Travis badge with GitHub Actions badges


README.rst
```diff
@@ -1,12 +1,32 @@
-|PyPI| |Travis| |Gitter| |DOI| |binder|
+|PyPI| |Test| |Branca| |Selenium| |Notebook| |Notebook-Style| |Publish-PyPI| |Gitter| |DOI| |binder|
```
ocefpaf, line 1: We should probably add only the code tests to the README b/c it is what the contributors/users need to know. The others are more of an internal thing and rarely block a PR from being merged.

oefe: Yeah, the list got a bit too long.

I'm not sure what you mean with "only the code tests". Surely contributor would want to know if their change breaks a  Selenium test or an example notebook? Or if their change to a notebook follows style conventions? So I left these in, and only removed the Branca and Publish-PyPI badges.

That's still two rows of badges, though

ocefpaf: > Surely contributor would want to know if their change breaks a Selenium test or an example notebook? 

Yeah but that is shown in the PR status. The badges here are not really for devs/contributors. More like a display case for, everything is green, move along :-)

ocefpaf:
Thanks @oefe! Looks really nice!!

PR #1556, opened 2021-12-28
Fix test failures with latest branca

oefe:
Bianca removed uuid4 recently for performance reasons.

Rather than trying to mock what branch is using now to generate IDs, I think it's better not to mock id generation at all. This required a few changes on how to test the result of `to_dict`.

I wonder, though, wether we really should compare for equality. This will break again whenever branca adds new values to the dict, or changes internals. Maybe it would be better to just assert on specific keys that we are interested in, or not check `to_dict` at all?

### DIFF
--- a/tests/test_folium.py
+++ b/tests/test_folium.py
@@ -23,11 +23,6 @@ import pandas as pd
 
 import pytest
 
-try:
-    from unittest import mock
-except ImportError:
-    import mock
-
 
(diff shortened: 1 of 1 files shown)


ocefpaf:
LGTM but I would I'll leave this here for @Conengmo to take a second look. He has been the main folium maintainer for a while now and know way more than I do about the current tests.

Conengmo:
Thanks @oefe for the fix!

PR #1557, opened 2021-12-28
Fix broken links in two example notebooks

oefe:
Two example notebooks had broken (dead) links.

This is a quick and dirty fix that replaces the broken links with equivalent links from elsewhere (i.e. same data at different location).

However, I wonder whether it would be better to include the linked data in our own repository instead?


examples/FloatImage.ipynb
```diff
@@ -25,7 +25,8 @@
     "\n",
     "\n",
     "url = (\n",
-    "    \"https://github.com/ocefpaf/secoora_assets_map/blob/a250729bbcf2ddd12f46912d36c33f7539131bec/secoora_icons/rose.png\"\n",
+    "    \"https://raw.githubusercontent.com/ocefpaf/secoora_assets_map/\"\n",
```
ocefpaf, line 28: Thanks for the permalink. Note that we need the raw URL here to display the image.

examples/plugin-Search.ipynb
```diff
@@ -328,20 +328,11 @@
...
-     "text": [
-      "/usr/local/Caskroom/miniconda/base/envs/FOLIUM/lib/python3.10/site-packages/IPython/core/interactiveshell.py:3377: FutureWarning: The `op` parameter is deprecated and will be removed in a future release. Please use the `predicate` parameter instead.\n",
-      "  if (await self.run_code(code, result,  async_=asy)):\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "us_cities = geopandas.sjoin(cities, states, how=\"inner\", op=\"within\")\n",
+    "us_cities = geopandas.sjoin(cities, states, how=\"inner\", predicate=\"within\")\n",
```
ocefpaf: This fixes the warning.

PR #1570, opened 2022-02-15
add width to css for FloatImage

### LINKED ISSUE #1001: Styling Imported Images
New to Python, kinda dove into it headfirst to create an interactive map with data. Really enjoying it, and really loving what Folium can do.

I've added a legend.png to create a legend (temporarily, or perhaps permanently, unless something else can be recommended) that I'd like to style by adding a box-shadow as well as adding a radius to curve the edges if I so desire.

Perhaps it's already in the notes somewhere, but I couldn't find it!

The legend itself works as it should and doesn't interfere with interacting with each site bubble (each site is expected to have more than just a "name" at some point. Actual data/graphs/etc. will be contained for nearly each site.)

```python
# Create Map Legend
from folium.plugins import FloatImage
image_file='images/Legend.png'

FloatImage(image_file,bottom=5,left=5).add_to(map)

```

Another quick question: can the output HTML file be modified with a viewport tag to assist with scaling on a mobile environment? I haven't tried it yet, and I assume each time I compile the app after I make changes the subsequent HTML file is entirely overwritten.

Thank you!

beautah:
Allows for setting relative size of an image without changing default behavior,  fixes #1001 imo...

### DIFF
--- a/folium/plugins/float_image.py
+++ b/folium/plugins/float_image.py
@@ -12,6 +12,7 @@ class FloatImage(MacroElement):
                         position:absolute;
                         bottom:{{this.bottom}}%;
                         left:{{this.left}}%;
+                        width:{{this.width}}%;
                         }
                 </style>
             {% endmacro %}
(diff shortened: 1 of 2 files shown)


beautah:
I chased down the all the failed checks and the remaining seem to be not related to any changes I've made. Happy to fix anything else up if needed though

PR #1576, opened 2022-02-27
feat: add Leaflet.VectorGrid plugin

iwpnd:
Hi 👋 

I wanted to use vector map tile layers in folium with streamlit and was surprised that [Leaflet.VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid) is not yet part of the folium plugin ecosystem.

Let me know what you think and if you don't want to consider this PR I'd be happy if [folium-vector](https://github.com/iwpnd/folium-vector) could be linked instead.


Conengmo:
This is a good candidate for a plugin, given our plugin acceptance criteria. https://github.com/python-visualization/folium/blob/main/.github/CONTRIBUTING.md#plugin-acceptence-criteria PR is also good. I'll rebase so we can run some more tests.

Conengmo:
Thanks @iwpnd!

PR #1585, opened 2022-05-05
update selenium tests after new branca version

Conengmo:
With https://github.com/python-visualization/branca/pull/96 we changed the way html is stored in notebooks. Update our notebook tests to take this into account.


Conengmo:
We still have to fix `test_heat_map_with_weights`, but at least the notebook tests now pass.

Issue #1025, opened 2018-11-22
Radio control instead of checkboxs for layer control


joaopalotti:
Hi all,

First of all, thanks for this wonderful library.

I am planning to visualize different data slices on top of the same map.
For example, let's say that for the US Unemployment example (https://python-visualization.github.io/folium/docs-v0.6.0/quickstart.html#Choropleth-maps) I have information of unemployment by gender and education. 

How can I create an exclusive group control like was done by this plugin (https://github.com/ismyrnow/leaflet-groupedlayercontrol) here using a radio control instead of a checkbox?

ocefpaf:
Every layer that has the option `overlay=True/False` can be set to a radio button and cycle the layers available. However, what you want is to group those layers with that plugin and this is not possible without the plugin ;-p

That functionality is a nice candidate for an external plugin but I don't believe it belongs in the core folium. (Happy to be convinced otherwise with a nice PR!)

Conengmo:
This was added in https://github.com/python-visualization/folium/pull/1592, which will be available in the 0.14.0 release.

PR #1592, opened 2022-05-19
add groupedlayercontrol plugin

chansooligans:
This PR adds the [leaflet-groupedlayercontrol plugin](https://github.com/ismyrnow/leaflet-groupedlayercontrol). Also addresses a closed issue that requested this PR: https://github.com/python-visualization/folium/issues/1025. 

Purpose of plugin is to allow grouping overlays and allowing within-group radio buttons. My motivation for this was because I had created folium maps with several choropleths (e.g. population, density, demographics) where it would make no sense to have multiple choropleths on simultaneously and confuse some users. 

Current implementation creates a second layer control but it would be possible to change so that base, grouped overlays, and ungrouped overlays are all contained in a single layer control.

The template and logic are modified copies of the parent class LayerControl.


Conengmo (commented):
[edit: this comment is actually about a different plugin that was mentioned in a related issue]

Though I'd like to include the Leaflet.Control.Layers.Tree plugin in folium, I don't think this PR is the way we should do it, for three reasons:

- The code in this PR contains a specific use case of that JS plugin: radio buttons in separate groups. The JS plugin itself doesn't only contain that, but accepts a generic layer tree.
- It contains custom code in Python as well as the JS template. Ideally, a folium plugin should only wrap the JS plugin. But maybe some is unavoidable here.
- It partially reimplements code of the normal LayerControl class.

IMO an adoption of this JS plugin should have as little code on the folium side as possible and expose the full feature. It should either replace the existing LayerControl class or be compatible with it.

I wonder how that can be done best in folium, since on the Python-side we structure layers a bit different than Leaflet. My best suggestion would be to pass a dictionary with layers to this plugin, then pass that to the JS plugin. And then not use the normal layer control. Something like:

```
layer1 = Layer()
layer2 = FeatureGroup()
layer 3 = Marker()
GroupedLayerControl(
    base_layers=[layer1],
    overlays={
        'Group': [layer2, layer3]
    }
).add_to(m)
````

Note that this mimics the arguments of the JS plugin. That means little code on the folium side. Also note that this doesn't use the `overlay` and `control` parameters of the folium objects, because the tree structure passed to GroupedLayerControl` defines how the layers are used.

I'd be interested to hear how others think about this.

Conengmo (commented):
Worked on this for a bit:

- Make it work together with LayerControl, so we don't have to re-implement that here.
- Make radio buttons optional, enabled by default.
- Make the user pass in objects instead of layer names. I think that's easier and less error prone.
- In the test use the rendered template.


https://nbviewer.org/github/chansooligans/folium/blob/plugins-groupedlayercontrol/examples/plugin-GroupedLayerControl.ipynb

Conengmo:
Thank you @chansooligans! Nice to have this plugin added to folium.

PR #1618, opened 2022-09-17
Send events as console logs

joaquinperaza:
Hi, I share this code that I am using to receive events in Python using Qt library.

These modifications allowed me to receive:
 - Mouse Click
 - Polygon Creation
 - Polygon Click
 
It should be really easy to add more events in different layers, elements, and features using this logic.

I feel that folium would be much more useful with two-way communication with python.

For reading the vents in PyQt6 I use:

```
class WebEnginePage(QWebEnginePage):
    def javaScriptConsoleMessage(self, level, msg, line, sourceID):
        event = (json.loads(msg))
        print(event)

web_map = QWebEngineView()
page = WebEnginePage(web_map)
web_map.setPage(page)
m = folium.Map(location=settings.COORDINATES, zoom_start=13)
tile = folium.TileLayer(
    tiles='http://mt1.google.com/vt/lyrs=s&x={x}&y={y}&z={z}',
    attr='Google',
    name='Google Satellite',
    overlay=False,
    control=True
).add_to(m)

data = io.BytesIO()
m.save(data, close_file=False)
web_map.setHtml(data.getvalue().decode())

```

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -191,7 +191,7 @@ class Map(JSCSSMixin, MacroElement):
             );
             
             {{ this.get_name() }}.on('click', function(e) {        
-                console.log(JSON.stringify({polygon_created: e.latlng}));    
+                console.log(JSON.stringify({MAP_CLICK: e.latlng}));    
             });
 
             {%- if this.control_scale %}
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -191,7 +191,8 @@ class Map(JSCSSMixin, MacroElement):
             );
             
             {{ this.get_name() }}.on('click', function(e) {        
-                console.log(JSON.stringify({MAP_CLICK: e.latlng}));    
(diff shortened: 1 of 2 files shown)


Conengmo:
Thanks for sharing your code here! I haven't worked with folium in Qt myself, but I understand what you're doing, it's an interesting approach.

I'm a bit hesitant to add this to the scope of folium, because it would mean:

- Adding console logs to our classes
- Coming up with a syntax those logs should have
- Adding documentation for it all
- Supporting it all for years to come

So I'd rather avoid it if possible.

Alternative thought: Leaflet generates events for most things already in Javascript. If you can read the Javascript console from Qt, I'm pretty sure you also should be able to catch the Leaflet Javascript events directly! And Leaflet already has excellent documentation on its events, for example https://leafletjs.com/reference.html#tooltip-click. A quick search suggestions looking at the `Qt WebChannel` class, maybe that's something you can check out?

I'll close this PR for now, but feel free to continue the discussion here. We can always re-open or start a new PR.

PR #1622, opened 2022-10-07
rename to avoid docs confusion and update

ocefpaf:
@Conengmo if this passes I'll self-merge and mint a new tag and release. Thanks for the new release BTW!


ocefpaf:
The test failure is unrelated to this PR. I'l send a new one to fix that. Merging this...

Issue #1316, opened 2020-04-29
Can't customize name for default layer


fallish:
I'd like to customize the name for default layer, and write code as below:

```python
import folium

m = folium.Map(tiles='OpenStreetMap', location=[41, -70], zoom_start=5, name='OSM')


folium.raster_layers.TileLayer(
    tiles='http://{s}.google.com/vt/lyrs=s&x={x}&y={y}&z={z}',
    attr='google',
    name='Google Maps',
    max_zoom=20,
    subdomains=['mt0', 'mt1', 'mt2', 'mt3'],
    overlay=False,
    control=True,
).add_to(m)

folium.LayerControl().add_to(m)

m

```
#### Problem description
The default layer name always is  lower case "openstreetmap" , which is not my expected name "OSM". 
![image](https://user-images.githubusercontent.com/1809740/80553893-019d3580-89fe-11ea-8bd6-747665015b44.png)

#### Expected Output
Add the ability to customize name for default layer.

Conengmo:
Hi @fallish , the way to do this is to disable the default layer and only add custom layers.

```
m = Map(tiles=None)
folium.raster_layers.TileLayer(tiles='openstreetmap', name='my name').add_to(m)
```

aleks-mariusz:
as per the refernced comment above, this is not an acceptable solution.. this case should be re-opened to create a proper workflow for those wanting to customize the name (such as the case in issue #1033 where the full url defaults to showing up in the layer control with no way to (properly) fix that)

Issue #1033, opened 2018-11-30
How to avoid whole tile layer url in LayerControl?


meteoDaniel:
How to avoid this dirty url in the layer control?

```python

    base_map = folium.Map(location=[50.377157, 9.36088], zoom_start=6,
                          tiles='https://api.mapbox.com/styles/v1/'
                                'mapbox/{theme}-v9/tiles/256/{z}/{x}/{y}?'
                                'access_token={token}'.format(theme='dark', z='{z}', x='{x}', y='{y}',
                                                              token=ACCESS_TOKEN_MAPBOX),
                          attr='Live Wetter Map')
```
![bildschirmfoto von 2018-11-30 16-14-09](https://user-images.githubusercontent.com/27021858/49297723-cc647f00-f4bb-11e8-99fb-8c6353900716.png)

ocefpaf:
You can use the `TileLayer` class instead and customize name.

ocefpaf:
> But this might be a bug if it is not posible to avoid this behaviour.

I don't follow.

> Otherwise you have to rename the "tiles" Option in Map Class. Because it is not working as expected.

The expected behavior is to use the URL for unnamed tiles. That is correct.

> My target is to use the mapbox layer as my basemap and not as an additonal tileLayer to the Openstreetmap.

Start with an empty map, tile=None, and add you tile.

eoefelein:
When I specify tile=None in instantiating the folium map object, and later specify with TileLayer, the state names disappear. But when I specify tile=my_custom_tile, I have the same issue, where the entire api name appears in the map's layer control. I haven't found a way to fix it.

aleks-mariusz:
btw, for anyone thinking issue #1316 could be a proper work-around for this: doing it that way seems to NOT have the intended effect, as when i do this, there is NO basemap selected on-loading the page (showing you a mapless page), and you then have to actually click on one of the tile layers in the layer control to display any map layer.. 

really seems like a bug (or poor workflow), just not sure whether it's a folium issue or a LeafLet issue?

i worked around it after m.save by using regex to in-place edit the file, changing the url being embedded as the name of the `layer_control....base_layers` array:

```
with open('index.html', 'r') as f:
    content = f.read()
    content_new = re.sub('"http.*" : tile_layer', r'"your mapname" : tile_layer', content)
with open("index.html", "w") as f:
    f.write(content_new)
```

as with any hacky "solution", YMMV

Conengmo:
[...]

PR #1624, opened 2022-10-07
allow passing TileLayer to Map

Conengmo:
Looking at https://github.com/python-visualization/folium/issues/1316 and https://github.com/python-visualization/folium/issues/1033, this seemed like an obvious improvement that doesn't actually increase the parameter footprint of the `Map` class.

With this change a user doesn't have to do the following:

```
m = Map(tiles=None)
TileLayer().add_to(m)
```

but can do this directly:

```
m = Map(tiles=TileLayer())
```

**Discussion**

I still think we should be careful to add more parameters to `Map`, to prevent it from becoming a god-class. If `Map` has the same parameters as `TileLayer`, we are doing something wrong. I think we should stick with `Map` having the `tiles` argument as a convenience, and have users use `TileLayer` when they want to do more advanced things.

Maybe a way to make this more clear is by allowing the `tiles` argument of `Map` to only be one of the pre-defined string arguments or a `TileLayer`, but not a custom url. Users wanting to use a custom url are thus forced to use `TileLayer`. That way we could also scrap the `attr` argument of `Map`.

Since that would mean breaking changes for some users, I don't think we should actually go ahead with that.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -97,9 +97,10 @@ class Map(JSCSSMixin, MacroElement):
         Width of the map.
     height: pixel int or percentage string (default: '100%')
         Height of the map.
-    tiles: str, default 'OpenStreetMap'
+    tiles: str or TileLayer, default 'OpenStreetMap'
         Map tileset to use. Can choose from a list of built-in tiles,
-        pass a custom URL or pass `None` to create a map without tiles.
+        pass a custom URL, pass a TileLayer object,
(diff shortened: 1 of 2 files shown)


ocefpaf:
Looks like a great improvement IMO and it is not really an extra parameter we are adding, just a valid input for an existing one.

ocefpaf (approved):

PR #1623, opened 2022-10-07
updated github actions

ocefpaf:
Pretty much update all actions, move to micromamba for speed and I'll rename some of them for consistency in the next commit if this passes.

PR #1628, opened 2022-10-10
CI: Add weekly scheduled run to all CI workflows

EwoutH:
Adds a weekly scheduled run to all CI workflows, which run each Monday at 06:00 UTC. A periodically scheduled run can detect errors and warnings appearing in changes in upstream dependencies or the build environment. By having a scheduled run they can be traced back easier to a dependency or environment change, then if they would pop up in a random PR.

Also enabled triggering a run manually (the `workflow_dispatch`) and updates [actions/checkout](https://github.com/actions/checkout) to v3.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,6 +1,11 @@
 name: Publish to PyPI
 
-on: ["push", "pull_request"]
+on:
+  pull_request:
+  push:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'
 
 defaults:
   run:
@@ -19,7 +24,7 @@ jobs:
 
     - name: Get tags
       run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
-      
+
 
     - name: Install build tools
       run: |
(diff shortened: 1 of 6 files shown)


ocefpaf:
@EwoutH sorry, I merged #1623 and clobbered most if your changes here. Do you mind rebasing?

BTW, I'm not a big fan on scheduled test. We kind of run the CIs a lot here in normal a "normal work day" and the scheduled ones will probably be ignored. With that said I'm +1 on the workflow dispatch option.

@Conengmo what is your opinion on this?

EwoutH:
No worries, conflicts were easily resolved :)

The scheduled run gives a reference run to compare (failed) CI runs to. Also, if there is a period no changes on the main branch, the scheduled run ensures failures introduced by dependency or environment changes can still be detected and traced back.

ocefpaf:
> The scheduled run gives a reference run to compare (failed) CI runs to. Also, if there is a period no changes on the main branch, the scheduled run ensures failures introduced by dependency or environment changes can still be detected and traced back.

Sure. I don't have a strong opinion. I just find them spammy and easily ignored, specially in active projects. BTW, we need to ensure it would run only here and not on forks to avoid it being even more spammy. Otherwise everyone with a fork would get a notification.

EwoutH:
GitHub thought of this, scheduled runs are disabled by default on forks :)

PR #1630, opened 2022-10-10
Build docs again.

ocefpaf:
Travis-CI is giving us a hard time. This moves the docs to GitHub Actions and fixes the build. We are losing the versioned docs here but I don't think that is too important b/c old docs do live in GH if folks need them.


ocefpaf:
@Conengmo I'd love a second pair of eyes here. I'd like to issue a patch release afterwards so we can ship functional docs with it.

Conengmo (approved):
The changes to the docs themselves look good.

I can't comment that much on the Actions workflow. It makes sense to me. Though I'm not that familiar with how the docs work currently. Were you able to test this workflow?

> We are losing the versioned docs here but I don't think that is too important b/c old docs do live in GH if folks need them.

That's okay with me if that's what it takes to get this working again without Travis.

I'll leave the merging to you.

Conengmo:
@ocefpaf unrelated note, would you mind squashing commits when you merge PR's?  I use https://github.com/python-visualization/folium/commits/main a lot and since we're very much PR oriented instead of commit oriented I'd love to have a clean overview here.

ocefpaf:
> @ocefpaf unrelated note, would you mind squashing commits when you merge PR's? I use https://github.com/python-visualization/folium/commits/main a lot and since we're very much PR oriented instead of commit oriented I'd love to have a clean overview here.

Done!

If this passes it should be good to go.

Conengmo (approved):
Let's give it a go!

ocefpaf:
The test needs to be updated. Working on it...

ocefpaf:
> The test needs to be updated. Working on it...

Ideally we should make a release of branca so this passes everywhere and not only on latest. I can work on that today.

Conengmo:
Done: https://github.com/python-visualization/branca/commit/9ec0805060e9ba5334e45b5e467ef7f4985d248c

PR #1567, opened 2022-01-26
Fixing flake8 warnings

pmains:
Fixing flake8 warnings, except unused folium.features.GeoJsonPopup import and _version.py newline error.

_version.py is not tracked and so changing that would have no effect on the codebase.

The unused import seemed possibly necessary inside of an __init__.py file.

PR #1632, opened 2022-10-10
Pre commit

ocefpaf:
Continuation of https://github.com/python-visualization/folium/pull/1567


ocefpaf:
@Conengmo, b/c this PR touches  several files I'll leave to you to merge it or not.

Conengmo:
Good initiative!

I understand using pre-commit locally. But what is the Github Action supposed to do, will it fail if pre-commit changes something?

By the way, did you see this warning? https://github.com/python-visualization/folium/actions/runs/3220435260

ocefpaf:
> I understand using pre-commit locally. But what is the Github Action supposed to do, will it fail if pre-commit changes something?

If it fails a user can fix it by reading the error message. We could configure pre-commit-ci to commit the fix but I find that very intrusive and leads to conflicts that may confuse new contributors.

Conengmo (approved):
Great, thanks for explaining. Let's merge this now and once we've got the open PRs out of the way add the same pre-commit stuff as branca has.

PR #1634, opened 2022-10-13
Jenks

pmains:
This feature change adds Jenks Natural Breaks Optimization to choropleths created using Folium.
https://en.wikipedia.org/wiki/Jenks_natural_breaks_optimization


ocefpaf:
I would prefer a PR that makes passing the pre-computed bins so users could use whatever they want. However, that may be too complex and would break a lot of code.

With that said, this PR is fine and it makes sense. My only request is to move the import inside the if-clause and add or add try/except clause for it at the top so we cab make it optional. I don't want to increase the number of dependencies in folium.

pmains:
Of course.

To be clear, you want to put the import inside of the if-clause _or_ add a try/except clause, not both, right?

folium/features.py
```diff
@@ -1212,7 +1213,14 @@ def __init__(self, geo_data, data=None, columns=None, key_on=None,  # noqa
         if color_data is not None and key_on is not None:
             real_values = np.array(list(color_data.values()))
             real_values = real_values[~np.isnan(real_values)]
-            _, bin_edges = np.histogram(real_values, bins=bins)
+            if use_jenks:
+                from jenkspy import jenks_breaks
+
+                if not isinstance(bins, int):
+                    raise ValueError(f'bins value must be an integer. Invalid value "{bins}" received.')
+                bin_edges = np.array(jenks_breaks(real_values, bins))
```
Conengmo, line 1221: I got an error with the color lookup because the last value in the array was not inclusive. Making it an array of floats solved the issue. I was using the `main` version of branca and your example notebook.

```suggestion
                bin_edges = np.array(jenks_breaks(real_values, bins), dtype=float)
```

examples/Jenks.ipynb
```diff
@@ -0,0 +1,191 @@
+{
```
Conengmo, line 1: Please rename this file to make it more clear what's it about. I would suggest: `Choropleth with Jenks natural breaks optimization.ipynb`.

folium/features.py
```diff
@@ -1174,6 +1174,7 @@ def __init__(self, geo_data, data=None, columns=None, key_on=None,  # noqa
                  line_weight=1, line_opacity=1, name=None, legend_name='',
                  overlay=True, control=True, show=True,
                  topojson=None, smooth_factor=None, highlight=None,
+                 use_jenks=False,
```
Conengmo, line 1177: could you include a line in the docstring above?

Conengmo (changes requested):
I like it! Have some comments, could you take a look?

Conengmo (commented):
Could you also include the notebook output in a commit? That way it will show up here: ~https://nbviewer.org/github/pmains/folium/blob/jenks/examples/Jenks.ipynb~

https://nbviewer.org/github/pmains/folium/blob/jenks/examples/Choropleth%20with%20Jenks%20natural%20breaks%20optimization.ipynb

examples/Jenks.ipynb
```diff
@@ -0,0 +1,191 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1bb0c3ba",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pip install jenkspy"
```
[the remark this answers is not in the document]
ocefpaf: Or, better yet, add `jenkspy` to our `requirement-dev.txt` to ensure it will be available in our dev environment. This environment is used to test the notebooks and the tests should pass with it.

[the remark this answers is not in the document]
Conengmo: Yes also that

Conengmo (approved):

Conengmo:
Thank you @pmains! Nice addition.

PR #1482, opened 2021-06-28
Upgrade bootstrap css 3.2 to 4.5

Shriukan33:
As requested in #1479

Didn't add any error on top of existing ones. 
(Done : 
run flake8 folium --max-line-length=120
run python -m pytest tests --ignore=tests/selenium
run python -m pytest tests/selenium)

This updates bootstrap's CSS.

PR #1644, opened 2022-11-10
Bootstrap 5

Conengmo:
Replaces https://github.com/python-visualization/folium/pull/1482.

Upgrading to Bootstrap 5 is not trivial, since the Awesome Marker package we use relies on Bootstrap 3 for glyphicons icons. Those are dropped in Bootstrap 4 and 5.

Maintain backwards compatibility by extracting the glyphicons css and fonts from Bootstrap 3 and including them in folium.

This way we can upgrade Bootstrap while still allowing the default icons of Awesome Marker.

### TODO in a next one:

include more documentation on what icons are available.

- https://getbootstrap.com/docs/3.3/components/#glyphicons
- Font Awesome

### To test this PR

In this PR I already link to the future location of the static css and fonts. If you want to test it locally, change the jsdeliver link to point to my repo and branch: https://cdn.jsdelivr.net/gh/Conengmo/folium@bootstrap-5/folium/templates/bootstrap_3_glyphicons/bootstrap-3-glyphicons.css


Conengmo:
All tests pass when I use the link to this feature branch. Now I'll revert it to use the future link again. Some tests will then fail, but it should be safe to merge.

Conengmo:
Yes, I included a CDN link to our main branch, so the files are not there yet until this PR is merged. But take a look at the `run tests with feature branch` commit, there the tests ran with the CDN link to this feature branch, and they passed.

ocefpaf:
Yeah. You mentioned that above and I missed it, sorry.

ocefpaf:
> Yeah. You mentioned that above and I missed it, sorry.

Can we do a permalink or something? I don't recall what is the best practice in this case. I'm OK merging this as-is but I have a feeling we dealt with this in the past.

Conengmo:
I don't think we solved this in the past... The best way would be to merge the static dependencies first, then do a second PR that uses them. But I guess I took the liberty to cut a corner here 😇

Conengmo:
Hmm this did break the Selenium tests on the main branch. Looking into it...

Conengmo:
Reverting this change in https://github.com/python-visualization/folium/pull/1648 since there's an issue with how we use the CDN.

Conengmo:
Yes, that's probably best!

`main` should be good now. I'm done for today, maybe more tomorrow. Only 7 PRs still open!

ocefpaf:
> Only 7 PRs still open!

\o/

PR #1646, opened 2022-11-10
Remove unused example index.html

Conengmo:
We have a file in our examples folder that seems to be unused.

The history of this file: https://github.com/python-visualization/folium/commits/main/examples/index.html

Looks like it was added in this commit by accident: https://github.com/python-visualization/folium/commit/6b0ae5ed9945cc54322b336dceab310dfe85a9a7

Remove this file since it is not used and appears to have been added by accident.


ocefpaf:
Should probably add to the .gitignore to avoid re-adding it in the future.

PR #1648, opened 2022-11-10
Revert "Bootstrap 5 (#1644)"

Conengmo:
This reverts commit 4e6e6e4d5d8cf5b03500531516495e918ff6b36b.

The jsdelivr CDN is not accepting this approach, saying: "Package size exceeded the configured limit of 50 MB." I guess it processes the entire folium repo, and complains that's too big? Let's look into it, but in the meantime make the main branch working again.

PR #1650, opened 2022-11-15
Bootstrap 5 (take 2)

Conengmo:
Second take of https://github.com/python-visualization/folium/pull/1644, which was reverted in https://github.com/python-visualization/folium/pull/1648.

This time don't host the icons ourselves, but use a standalone version on the official Bootstrap CDN. It was referenced here: https://stackoverflow.com/questions/18222849/bootstrap-3-glyphicons-cdn

If at some point this becomes unavailable I'm afraid we'll just have to drop Glyphicons. Bootstrap 3 is a bit old now, so while sad it's not completely unexpected support disappears at some point. But for now let's hope this CDN endpoint stays up.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -29,7 +29,7 @@ _default_js = [
     ('jquery',
      'https://code.jquery.com/jquery-1.12.4.min.js'),
     ('bootstrap',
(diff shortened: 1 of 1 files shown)


Conengmo:
Test failure is unrelated.

Issue #946, opened 2018-08-29
show folium outside of ipython notebooks


Demetrio92:
Hey guys i am trying to run one of the examples you provide. However, I am unable to render it. It's probably because I am using it in a normal python script inside the PyCharm. Is there any functionality similar to matplotlib `show()`? I tried `render()` but it does not do anything...

```python
import folium
from folium.plugins import HeatMap
import numpy as np
print(folium.__version__)
#0.6.0

data = (np.random.normal(size=(100, 3)) *
        np.array([[1, 1, 1]]) +
        np.array([[48, 5, 1]])).tolist()
m = folium.Map([48., 5.], tiles='stamentoner', zoom_start=6)
HeatMap(data).add_to(m)
m
# <folium.folium.Map object at 0x7f4c56e14b70>
m.render()
# nothing happens


```

Demetrio92:
So, the answer is no -- there is no such functionality in the package at the moment. Which is unfortunate, as it is easy to implement and e.g. `pyLDAvis` already solved the problem. 
I'll provide a custom solution below. 
For cross-reference, this issue is related to #781 and #455.

Demetrio92:
**Proof of concept.** 
Tested with python 3.5 and folium 0.6.0

```python
import numpy as np
import folium
from folium.plugins import HeatMap

# create map
data = (np.random.normal(size=(100, 3)) *
        np.array([[1, 1, 1]]) +
        np.array([[48, 5, 1]])).tolist()
folium_map = folium.Map([48., 5.], tiles='stamentoner', zoom_start=6)
HeatMap(data).add_to(folium_map)

# this won't work:
folium_map
folium_map.render()



# ------------------------------------------------------------------------------------------------
# so let's write a custom temporary-HTML renderer
# pretty much copy-paste of this answer: https://stackoverflow.com/a/38945907/3494126
import subprocess
import webbrowser
from http.server import BaseHTTPRequestHandler, HTTPServer


PORT = 7000
HOST = '127.0.0.1'
SERVER_ADDRESS = '{host}:{port}'.format(host=HOST, port=PORT)
FULL_SERVER_ADDRESS = 'http://' + SERVER_ADDRESS


def TemproraryHttpServer(page_content_type, raw_data):
    """
    A simpe, temprorary http web server on the pure Python 3.
    It has features for processing pages with a XML or HTML content.
    """

    class HTTPServerRequestHandler(BaseHTTPRequestHandler):
        """
[...]

PR #1651, opened 2022-11-16
Map.show_in_browser() using temp file

Conengmo:
Alternative for https://github.com/python-visualization/folium/pull/953/.

Related issue: https://github.com/python-visualization/folium/issues/946

Allow opening a Map in the browser directly. But instead of using a server, use a temporary file instead. In functionality it should be equivalent, but the code is simpler.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -384,7 +384,7 @@ class Map(JSCSSMixin, MacroElement):
                 'Your map should have been opened in your browser automatically.'
                 '\nPress ctrl+c to return.'
             )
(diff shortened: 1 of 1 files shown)


ocefpaf:
I kind of feel ashamed we never implemented that in the first place. We kind of had our heads buried deep in notebooks since day 1. Thanks for doing this!

Conengmo:
Feels pretty trivial now doesn't it :) I remember looking at https://github.com/python-visualization/folium/pull/953 and not liking the amount of code, but I didn't realize yet how it could be done otherwise 🤦‍♂️

PR #1652, opened 2022-11-17
Pre-commit with black, isort, spellcheck

Conengmo:
Apply the same packages we use in branca, see https://github.com/python-visualization/branca/pull/120. I didn't include https://github.com/asottile/add-trailing-comma though, because if I'm not mistaken Black already does this.

Picked some settings from https://github.com/python-visualization/folium/pull/1413, which is now superseded by this PR.

I ran `pre-commit run --all-files`, then afterwards manually cleaned up some line break artifacts.


Conengmo:
I'm opening more PRs, so I'll merge this, otherwise I'll have to keep updating it.

Issue #1471, opened 2021-04-20
Support rendering large GeoJSON datasets with WebGL


danwild:
Hi,

I'm interested in an open source alternative to [PyDeck/DeckGL](https://deckgl.readthedocs.io/en/latest/gallery/geojson_layer.html) for rendering large GeoJSON datasets on an interactive map (DeckGL uses MapBoxGL v2 for map, which is no longer open source).

I've made an initial javascript [plugin for leaflet ](https://github.com/onaci/Leaflet.glify.layer)(well, an extension for an [existing plugin](https://github.com/robertleeplummerjr/Leaflet.glify)) to handle this rendering.
I'm now interested in leveraging this functionality in a python context (using [Panel](https://panel.holoviz.org/index.html)).

Does this sound like something that might fit well in Folium? (my initial idea was to add this to [ipyleaflet](https://github.com/jupyter-widgets/ipyleaflet/issues/815), but unfortunately ipywidgets [don't work](https://panel.holoviz.org/FAQ.html) with panels bokeh server deployment option 😕 ).

If so, should be able to put a PR together. I've seen the plugin contrib guide, any other tips on how to get started much appreciated!

Cheers,
Dan

Conengmo:
Hi Dan,

I'm not familiar with that plugin, so I can't really say whether it's a good fit. If we can make it easier for people visualize large datasets, that does sound worthwhile.

Maybe a good way to go about this is to host this new folium plugin on a repo of your own first. That way you can build it according to your needs, make sure it works without having to deal with managing a folium PR as well. Then, if it works and you feel like it's something that would be useful for folium users in general we can look at a PR. If not, we can still link to your plugin code so people can use it.

Some thoughts:
- The trouble with geojson though is that it's hard to validate and give meaningful user feedback when the format is not correct for the application. If we have this as a plugin, I'd want to give users meaningful error messages if their geojson is not correct. But without having a heavy validation feature. Or maybe such a feature should be a core part of folium.
- As the developer of this feature you are an expert user. Keep that in mind when designing the API. If the API feels too simple for you, it's probably right for most users.

danwild:
Thanks,

As per @Conengmo's suggestion I've setup an initial standalone plugin here: https://github.com/onaci/folium-glify-layer

I'm happy to keep it standalone if others don't think it should be a 'core' folium plugin (although imo handling of larger datasets is only going to become more of an issue/requirement).

Some links to underlying tools/notes for anyone interested:
[...]

PR #1659, opened 2022-11-18
Link out to other packages and plugins

Conengmo:
Closes https://github.com/python-visualization/folium/issues/1471


ocefpaf (approved):
Should probably add to the docs but then I think about it, the README is good enough for this info.

Conengmo:
Yes, we can surely move this around if we want.

PR #1662, opened 2022-11-23
Leaflet Control

Conengmo:
Trying out an idea I had based on a number of issues I've encountered. Use the Leaflet Control class to put legends, images or other custom stuff on the map. That way they integrate better with other Leaflet features, such as full screen and potentially layer control.

Before this is ready I'll look at how this combines with layer control. And what the potential benefits are for other folium classes that are currently outside of Leaflet control.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -21,6 +21,7 @@ from folium.map import FeatureGroup, Icon, Layer, Marker, Popup, Tooltip
 from folium.utilities import (
     _parse_size,
     camelize,
+    escape_backticks,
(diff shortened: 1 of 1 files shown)


hansthen:
@Conengmo Would you mind if I continue this branch?

The feature in itself would be really useful. Also, from an architectural perspective, adding a `Control` base class would also be beneficial, especially for plugins with items that subclass `L.Control`. These now subclass directly from `branca.MacroElement` which is not the cleanest solution.

Issue #828, opened 2018-03-02
Folium can't parse TopoJSON objects with dashes `-` in name


kevhill:
This works just fine in an ipython notebook
```python
m = folium.Map([0, 0], zoom_start=7)
folium.TopoJson(
    {
      "type":"Topology",
      "transform":{
        "scale": [1,1],
        "translate": [0,0]
      },
      "objects":{ 
        "two_squares":{
          "type": "GeometryCollection",
          "geometries":[
            {"type": "Polygon", "arcs":[[0,1]],"properties": {"name": "Left_Polygon" }},
            {"type": "Polygon", "arcs":[[2,-1]],"properties": {"name": "Right_Polygon" }}
          ]
        }
      },
      "arcs": [
        [[1,2],[0,-2]],
        [[1,0],[-1,0],[0,2],[1,0]],
        [[1,2],[1,0],[0,-2],[-1,0]]
      ]
    },
    object_path='objects.two_squares'
).add_to(m)
m
```

but changing the object to include a `-` renders the map, but not the polygons
```python
m = folium.Map([0, 0], zoom_start=7)
folium.TopoJson(
    {
      "type":"Topology",
      "transform":{
        "scale": [1,1],
        "translate": [0,0]
      },
      "objects":{ 
        "two-squares":{
          "type": "GeometryCollection",
          "geometries":[
            {"type": "Polygon", "arcs":[[0,1]],"properties": {"name": "Left_Polygon" }},
            {"type": "Polygon", "arcs":[[2,-1]],"properties": {"name": "Right_Polygon" }}
          ]
        }
      },
      "arcs": [
        [[1,2],[0,-2]],
        [[1,0],[-1,0],[0,2],[1,0]],
        [[1,2],[1,0],[0,-2],[-1,0]]
      ]
    },
    object_path='objects.two-squares'
).add_to(m)
m
```
#### Problem description

Literally took me 3 hours to diagnose, as the first topojson file I ever tried to render used `-`s

#### Expected Output

Ideal fix would just be to render all objects, as there is no topojson specification that precludes `-`s. Consolation prize would be a friendly error message when `object_path` contains a `-`

#### Output of ``folium.__version__``
`0.5.0`

olavrel:
By saving the two cases to html and inspecting the results, it seems the culprit the usage of dot notation on accessing object properties. Line 69: `....objects.two-squares`
Using bracket notation would have worked fine I guess? `...objects["two-squares"]`.

[...]

PR #1665, opened 2022-11-26
Fix topojson object path

Conengmo:
Closes https://github.com/python-visualization/folium/issues/828

Fix an issue with Topojson object paths. With the current code each item in the path should be a valid Javascript identifier, but users may not know this.

I looked into JS identifier validation, but I don't think that's the way to go:

- Topojson doesn't demand the objects have valid JS identifiers as names.
- There are many ways in which a JS identifier can be invalid, so it's hard to test.
- Users may be unable to change their Topojson data easily.

Instead I opted to convert the path to an array notation: `obj1.obj2` -> `["obj1"]["obj2"]`. After we escape double quotes any string should work. With this, object paths that include hyphens work.

Alternatively, only look for hyphens and raise an error if we see one. I'd rather fix the complete problem instead of just one specific case. But if this change is too complex, that might be a small step to take instead.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -24,6 +24,7 @@ from folium.utilities import (
     get_bounds,
     get_obj_in_upper_tree,
     image_to_url,
+    javascript_identifier_path_to_array_notation,
     none_max,
     none_min,
     parse_options,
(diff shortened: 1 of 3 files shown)


Conengmo:
@ocefpaf or @martinfleis, would one of you mind reviewing this change?

martinfleis (approved):
This looks good.

> Alternatively, only look for hyphens and raise an error if we see one. I'd rather fix the complete problem instead of just one specific case. But if this change is too complex, that might be a small step to take instead.

It doesn't seem to be an overly complex solution. I am fine with it and agree that it is better to fix the whole issue where it starts rather than a specific subset of it.

Conengmo:
Thank you! Appreciate the review!

Issue #1248, opened 2020-01-14
support WMS parameter cql_filter


bhogan-mitre:
Not all WMS parameters are `camelCase`, and the current use of `camelize` by `parseOptions` prevents the proper application of some parameters through `**kwargs`.

```python
import folium

m = folium.Map(
  location=[40, -100],
  zoom_start=4
)
 
states = folium.WmsTileLayer(
  url = "https://ahocevar.com/geoserver/wms",
  layers = "topp:states",
  fmt = "image/png",
  transparent = True,
  cql_filter = "STATE_ABBR IN ('NY','IL')"
)

states.add_to(m)
m
```
![Screen Shot 2020-01-14 at 9 25 30 AM](https://user-images.githubusercontent.com/32336739/72352853-04735500-36b1-11ea-90ee-ea392c2f0ba9.png)

#### Problem description
Leaflet [`TileLayer.WMS`](https://leafletjs.com/reference-1.6.0.html#tilelayer-wms) options state that:
>If any custom options not documented here are used, they will be sent to the WMS server as extra parameters in each request URL. This can be useful for non-standard vendor WMS parameters.

One particularly useful parameter is [`cql_filter`](https://docs.geoserver.org/stable/en/user/services/wms/vendor.html#cql-filter).  The `cql_filter` parameter should be passed through `**kwargs` like any other parameter.  Currently `camelize` is treating that as `cqlFilter` which prevents it from being used.  `format_options` is another parameter that would be impacted by this.

#### Expected Output
`cql_filter` should be applied to limit the set of results from `WmsTileLayer`.

![Screen Shot 2020-01-14 at 9 35 50 AM](https://user-images.githubusercontent.com/32336739/72352996-48fef080-36b1-11ea-8740-e278738001b6.png)

#### Output of ``folium.__version__``
`0.10.1`

Conengmo:
Hi @bhogan-mitre, thanks for reporting this. I'm thinking this may be best solved by making a special case in the `WmsTileLayer` class when unpacking the kwargs. This should be fairly trivial. Maybe you want to make a PR?

kirodh:
Hi there

Was this fixed by any chance?

Kind regards

thiagomesUNB:
> Hi there
> 
> Was this fixed by any chance?
> 
> I got around this by adding in a special case for CQL_FILTER keyword where the kwargs are camelized.
> 
> But its still a hack i suppose and only works for an all caps (CQL_FILTER) keyword.
> 
[...]

PR #1673, opened 2022-11-29
WmsTileLayer: allow cql_filter parameter

Conengmo:
Closes https://github.com/python-visualization/folium/issues/1248

WmsTileLayer users requested a `cql_filter` parameter be passed through without converting to lowerCamelCase notation.

### DIFF
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -205,6 +205,7 @@ class WmsTileLayer(Layer):
         transparent=False,
         version="1.1.1",
         attr="",
+        cql_filter=None,
         name=None,
         overlay=True,
         control=True,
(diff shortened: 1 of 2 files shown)


folium/raster_layers.py
```diff
@@ -222,6 +223,9 @@ def __init__(
             attribution=attr,
             **kwargs
         )
+        if cql_filter:
+            # special parameter that shouldn't be camelized
+            self.options["cql_filter"] = cql_filter
```
ocefpaf, line 228: Can we pop it from the kw args to avoid adding an extra arg entry? Not sure if this is best but maybe something like:

```python
cql_filter = kwargs.pop("cql_filter", None)
if cql_filter:
    # special parameter that shouldn't be camelized
    self.options["cql_filter"] = cql_filter
```

ocefpaf (approved):

PR #1683, opened 2022-12-25
Fix test heatmap with selenium

Conengmo:
For some reason our Selenium test for the heatmap plugin broke. Looks like the image we test for changed every so slightly. Update the image to fix it.

Ideally we would replace this fragile test, but I don't have a good idea how. So for now just keep it. We might also consider removing it, since it gave more headaches than that it helped.


ocefpaf:
> Ideally we would replace this fragile test, but I don't have a good idea how. So for now just keep it. We might also consider removing it, since it gave more headaches than that it helped.

We can probably increase the tolerance for the image comparison.

PR #1685, opened 2022-12-27
Add FEMlium to packages

francesco-ballarin:
Hi,
I noticed that https://github.com/python-visualization/folium/pull/1659 added a packages section to the README.

If you are looking for new packages to be added, I can contribute with another package: FEMlium, https://femlium.github.io/.
My project has a quite specific audience, so I would understand if you feel that its scope is not broad enough to be added here. In that case, feel free to close this PR.


Conengmo:
Cool! Thanks for adding this.

Issue #1573, opened 2022-02-20
Performance hit when FeatureGroup is created with show=False.


spacediver99:
Thanks for the great library - fun to use, and beautiful!

Not sure whether this was best put in the bug or feature request category, so to be safe I chose the latter.

I've run into the following performance issue:

The following code renders the map very fast:
```
import folium

m = folium.Map(location=[35.11567262307692,-89.97423444615382], zoom_start=12, tiles='Stamen Terrain')

for i in range(200):
    feature_group = folium.FeatureGroup(i, show=True)
    feature_group.add_to(m)

folium.LayerControl().add_to(m)
m
```

However, if show is set to False, it takes much longer to render.

```
import folium

m = folium.Map(location=[35.11567262307692,-89.97423444615382], zoom_start=12, tiles='Stamen Terrain')

for i in range(200):
    feature_group = folium.FeatureGroup(i, show=False)
    feature_group.add_to(m)

folium.LayerControl().add_to(m)
m
```

The problem is, I don't want these layers shown by default. So I'm wondering if there's a way to render the map, and then automatically uncheck the layers (I know that in these examples, the layers are "empty", but that is just to illustrate that the performance issue has nothing to do with the "content" of the layers).

Failing that, are there any other ways to deal with this performance issue?

Thanks very much for any guidance!

Conengmo:
That's interesting. I can confirm this happens for me as well.

Know that `show=False` means we deactivate the layer after creating it. In other words, the layer will be created and shown, then removed again. I guess that's maybe causing the performance hit?

If we wanted to do this differently, it would mean we'd have to update the templates of all our classes! We'd have to change the line `.addTo({{ this._parent.get_name() }})` into something like `if (show) {.addTo({{ this._parent.get_name() }})}`. That's a lot of changes, so that's why we opted to remove them after creation.

I'll label this as a bug, since it's an unintended effect.

spacediver99:
Thanks for the reply!

Given that it seems to take significantly longer to deactivate a set of layers than it is to create them, I wonder if there's any room for speeding up the deactivation part. I tried digging through the python code a while back, but was unable to find the relevant sections - perhaps this is because the issue is in the leaflet library? (I have no experience with JavaScript)

Conengmo:
Removing the layers with `show=False` happens here: https://github.com/python-visualization/folium/blob/2e86b480bb77ce4b83f3a761233ec9dc0cceafec/folium/map.py#L137

[...]

PR #1690, opened 2023-01-04
Add layers in Layer

Conengmo:
Our current approach to adding layers to the map and how that works with LayerControl is:
- each Layer subclass adds itself to its parent
- then when rendering LayerControl we remove the layers that have `show=False`.

This has two issues, one big and one minor:
- Performance. Layers are added, then removed, which is overhead. See https://github.com/python-visualization/folium/issues/1573.
- Conceptually this doesn't correspond to how Leaflet is designed.

Address this by letting subclasses of `Layer` only add themselves to their parent if `show=True`.

We can remove the code from each of the subclasses and render it in `Layer` instead.

If you're asking, whoever did this in the first place? It was me, my first PR on this project :) https://github.com/python-visualization/folium/pull/772. Then slight changes were made in LayerControl in https://github.com/python-visualization/folium/pull/1101.

### Order of multiple base layers

The downside is that this changes behavior when you add multiple base layers. Currently only the first one is enabled. With this change the last one added will be enabled. 

I'm conflicted whether this is something we should want to avoid. It does fit more closely to how Leaflet works. If you want to add multiple TileLayers, you will have to use the `show` parameter to define which one you want to see. For example:

```
m = Map(tiles='cartodbpositron')
TileLayer('stamen toner', show=False).add_to(m)
```

### Downstream packages

Packages that use folium might suffer from this change, which I would call an internal breaking change. End-users shouldn't notice it, but others that develop using folium may. For example tests that test the correctness of renders will have to update.

### folium convention

It may confuse folium developers that Layer subclasses don't add themselves to their parent, while other classes do.

## Change parent class of two plugins

- `SideBySideLayers` currently inherits from `Layer`, but is not actually a layer-type class. Inherit from MacroElement instead.
- Same for TimestampedWmsTileLayers`, it's not a layer itself.

I'll probably split these two changes of into separate PRs.


ocefpaf:
I'm OK with this change. It made sense in the past but getting closer to leaflet and snappier maps are worth the breaking changes. We should probably aim to a 1.0 release anyway. So, if you have more breaking changes, now is the time ;-p

ocefpaf (approved):

Conengmo:
Thank you for the review.

> We should probably aim to a 1.0 release anyway.

Agreed. I'd like to include the improved documentation, other than that I don't have big ideas for 1.0.

> If you have more breaking changes, now is the time

That's a good point. I'll think on it.

I did consider doing a refactor of our modules, especially `map.py` and `features.py`. But in the end I don't think it's worth the hassle.

Conengmo:
@randyzwitch @giswqs since I know you both have projects that use folium, I wanted to give you a heads-up about this upcoming (breaking) change.

- If you have tests that use the 'raw' templates, you may need to make changes to those.
- It's not inconceivable this change introduces a bug in functionality, please let us know if that's the case.
- We will require the user to explicitly define what base maps they want to show. If you have code that adds base maps, you may need to add the `show` parameter.

Please let me know if you have questions or comments.

giswqs:
Thanks for the heads up. I will keep an eye on this.

PR #1693, opened 2023-01-07
Fit enabled overlays

Conengmo:
I took https://github.com/python-visualization/folium/pull/1317 and revived the change. Closes https://github.com/python-visualization/folium/issues/1313. Personally, I think this is a nice feature to have. 

Add a class that automatically fits the bounds to the enabled overlays. Include some options to customize it. I opted to not enable the fitting when *removing* an overlay, that seemed more natural to me.

Example notebook that demonstrates the behavior (and functions as automated test) here: https://nbviewer.org/github/Conengmo/folium/blob/zoom-selected/examples/FitOverlays.ipynb


ocefpaf (approved):
Nice! Works like a charm.

Conengmo:
Thanks for the review! Appreciate it.

PR #1704, opened 2023-01-17
add local file prefix for browser

RileyLeff:
The show_in_browser() method wasn't working at all for me -- no browser popped up, but no error thrown. I'm on a M1 13" Macbook Pro running Ventura 13.1 and Python 3.11.1. 

Adding a "file://" prefix to the fname makes the browser work as expected. This change doesn't break any tests on my end. Tested on MacOS and Debian, don't have access to a windows machine at the moment.

First PR here so if I've done something horrifyingly wrong, let me know!


Conengmo:
Thank you Riley! Much appreciated.

PR #1706, opened 2023-01-21
Fix documentation generation

Conengmo:
I noticed our documentation hasn't been updated since we moved away from Travis. Probably something wrong in our Github Actions workflow. I noticed we don't run the workflow on release, could that be the issue? To possibly solve this I copied the triggers from the 'Publish to PyPI' workflow.

If this would indeed solve the issue, I suspect we would then have to do a release to update the documentation. Or should I just run the documentation and update the `gh-pages` branch manually?

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,6 +1,13 @@
 name: Publish Docs
 
-on: ["push", "pull_request"]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  release:
+    types:
+      - published
 
 jobs:
   run:


.github/workflows/deploy-docs.yml
```diff
@@ -1,6 +1,13 @@
 name: Publish Docs
 
-on: ["push", "pull_request"]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
martinfleis, line 7: These will not do much apart from checking if sphinx did not fail as the last step will run only on release. But the lines below are needed here.

martinfleis:
> If this would indeed solve the issue, I suspect we would then have to do a release to update the documentation. Or should I just run the documentation and update the gh-pages branch manually?

I would run it manually and push to the branch.

Conengmo:
Thanks for the review Martin. I'll run the docs manually and add a PR to add them to the gh-pages branch. Then also merge this so we have the 'on release' trigger for our next release.

> These will not do much apart from checking if sphinx did not fail as the last step will run only on release

Agreed, but I suppose it's nice to check that on each PR.

PR #1707, opened 2023-01-21
add version 0.14.0 to docs on gh-pages branch

Conengmo:
Follow up on https://github.com/python-visualization/folium/pull/1706

- Update the main docs to version 0.14.0.
- Also add a folder with the same docs in the same way as older versions. We will link to these in the future so we have versioned docs again.
- Remove the 'dev' folder. It points to master branch version in 2021, so it's no longer relevant.


martinfleis:
Hey, sorry for not looking at this earlier but you have accidentally removed `[.nojekyll](https://github.com/python-visualization/folium/pull/1707#diff-a02e6962c695f8d4004cb2465a3785275fe8a1a056d797ded1bd329a3bf35934)` causing the docs to fail to load styles, JS etc. We need it to tell GitHub pages to not touch it.

Conengmo:
Oops! I'll get it back...

Conengmo:
Seems good now. I guess when I checked before I still had them in my browser cache. Thanks for looking out Martin!

PR #1721, opened 2023-02-08
python 3 language updates

marksmayo:
- replaced type with isinstance (faster calls)
 - updated format strings to f-strings
 - tidied imports

### DIFF
--- a/folium/__init__.py
+++ b/folium/__init__.py
@@ -2,47 +2,19 @@ import sys
 
 import branca
 from branca.colormap import ColorMap, LinearColormap, StepColormap
-from branca.element import (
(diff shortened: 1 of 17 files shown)


Conengmo:
Hi Mark, thanks for your effort for our project. Unfortunately I won’t merge this PR, for the following reasons:

- the imports already followed a certain style
- changing all formatted strings leads to a big diff that I don’t think is worth the effort: of reviewing, making sure lines and comments are still correct. And the extra steps it adds to the commit history.

If you’d made a PR with just the isinstance changes that would be acceptable to me.

Thanks again for your interest!

PR #1728, opened 2023-02-17
Use Ruff instead of pyflakes

ocefpaf:
@Conengmo maybe you may want to revert the last two commits but I added them here just in case.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
   hooks:
     - id: codespell
       args:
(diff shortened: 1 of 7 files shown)


martinfleis:
We can also replace isort with ruff and have one less tool to depend on.

ocefpaf:
> We can also replace isort with ruff and have one less tool to depend on.

Cool! I'm still learning how `ruff` works. I'll look into that now.

martinfleis (approved):

Issue #1730, opened 2023-02-18
SideBySideLayers plugin not working


giswqs:
**Describe the bug**
The SideBySideLayers plugin for folium v0.14.0 is not working properly. The slider can't be moved. 

@Conengmo  @fralc

**To Reproduce**

https://colab.research.google.com/github/python-visualization/folium/blob/main/examples/Plugins.ipynb

```
```bash
!pip install -U folium
```
```python
import folium
from folium import plugins

m = folium.Map(location=(30, 20), zoom_start=4)

layer_right = folium.TileLayer('openstreetmap')
layer_left = folium.TileLayer('cartodbpositron')

sbs = plugins.SideBySideLayers(layer_left=layer_left, layer_right=layer_right)

layer_left.add_to(m)
layer_right.add_to(m)
sbs.add_to(m)

m
```

![Peek 2023-02-18 14-43](https://user-images.githubusercontent.com/5016453/219885093-1bdcd18b-07fa-4170-9091-dcd9c3a0a305.gif)

It should work like this: 

https://ipyleaflet.readthedocs.io/en/latest/controls/split_map_control.html
![Peek 2023-02-18 14-45](https://user-images.githubusercontent.com/5016453/219885184-5c2ad085-e7d6-4a29-b253-7f7ae47af049.gif)

jtmiclat:
I was able to fix the issue by pinning to an older version of the minified js
```python
import folium
from folium import plugins

class FixedSideBySideLayers(plugins.SideBySideLayers):
    default_js = [
        (
            "leaflet.sidebyside",
            "https://cdn.jsdelivr.net/gh/digidem/leaflet-side-by-side@2.0.0/leaflet-side-by-side.min.js",
        ),
    ]
   


m = folium.Map(location=(30, 20), zoom_start=4)

layer_right = folium.TileLayer('openstreetmap')
layer_left = folium.TileLayer('cartodbpositron')

sbs = FixedSideBySideLayers(layer_left=layer_left, layer_right=layer_right)

layer_left.add_to(m)
layer_right.add_to(m)
sbs.add_to(m)

m
```

giswqs:
@jtmiclat Yes, it works.

jtmiclat:
Thanks for checking it! I opened a PR #1732 to merge the fix.

Conengmo:
Since https://github.com/python-visualization/folium/pull/1732 was merged we can close this issue.

mrgeorge:
[...]

PR #1732, opened 2023-02-26
Pin older version of leaflet side by side

jtmiclat:
Addresses issue raised in #1730

### DIFF
--- a/folium/plugins/side_by_side.py
+++ b/folium/plugins/side_by_side.py
@@ -39,7 +39,7 @@ class SideBySideLayers(JSCSSMixin, MacroElement):
     default_js = [
         (
             "leaflet.sidebyside",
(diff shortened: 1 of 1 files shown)


martinfleis:
Do we know what is going on apart from that the latest version does not work? If it is a bug it should be fixed (either here or there), pinning is only a temporary solution.

martinfleis:
Interesting. Are you able to pinpoint the source? If it is leaflet or the plugin and report it there? We can pin it here but we need to track the progress to unpin it once resolved in the upstream.

giswqs:
It seems the `leaflet-side-by-side` plugin is the root cause. 
https://github.com/digidem/leaflet-side-by-side/issues/48

giswqs:
Sent the wrong link. Should be this one: https://github.com/digidem/leaflet-side-by-side/issues/47

martinfleis (approved):
Approving as a temporary fix until the upstream solution is released. Thanks!

PR #1735, opened 2023-03-09
implemented vmin, vmax in ImageOverlay()

fedro86:
together with pull request https://github.com/python-visualization/branca/pull/129 in branca, this update will implement vmin vmax in image overlay.

### DIFF
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -307,6 +307,8 @@ class ImageOverlay(Layer):
         colormap: Optional[Callable] = None,
         mercator_project: bool = False,
         pixelated: bool = True,
+        vmin: Optional[float] = None,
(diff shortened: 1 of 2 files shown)


Conengmo:
Hi @fedro86, thanks for your PR! Though I appreciate your effort, I'm thinking right now not to merge this, because I think normalizing the image is something a user can do themselves in Python, outside of branca/folium. I don't think it's something within scope for folium. Especially since this needs a coordinated change in branca and folium, which is a hassle. Furthermore, this can also already be achieved by the colormap argument, which maps values to colors.

If you think I'm wrong here, please let me know. I hope you can appreciate the balance we need to strike between adding new features and preventing the code base from becoming too large and unmaintainable.

PR #1756, opened 2023-05-04
Support TagFilterButton in PolyLine, AntPath and GeoJson

### LINKED ISSUE #1751: How can I filter PolyLine or AntPath
I want to filter PolyLine or AntPath, 
I try to use the TagFilterButton but it only support the Marker class.
Is there any solution?

jamiechoi1995:
Support TagFilterButton in PolyLine, AntPath and GeoJson
Fix #1751


Conengmo:
Hi Jamie, thanks for your PR! Nice to see that the TagFilterButton works for more than just markers. Since they don't specify this in their own documentation (https://github.com/maydemirx/leaflet-tag-filter-button) I'm hesitant to 'officially' support this in Folium. 

I took your branch and added some changes, hope you don't mind.

- allow the `tags` parameter in any vector layer, not just PolyLine. Don't add it to the documentation though.
- allow any extra options in `GeoJson` as a keyword argument. This makes it possible to use a `tags` argument.

With this, you will be able to use `TagFilterButton` with GeoJson, PolyLine and Antpath. But Folium won't officially support it, so no guarantees this will continue working. Since we are dependent on an externally maintained plugin we don't have control over that.

jamiechoi1995:
Your solution is more elegant, looks good to me. 😄 

I think supporting this in Folium is needed.
In fact, TagFilterButton claim that they can also support geojson, 
https://github.com/maydemirx/leaflet-tag-filter-button/blob/cddc7491f1e3816c120bdf8ca6c0236a09fbe52d/README.md?plain=1#L2
although they have not shown this in their live demo.

Conengmo:
Sounds good. I added a specification to the docstring. Once the tests pass we can merge this one.

PR #1757, opened 2023-05-05
#1751 issue: add Filter polyline

adityakishor1:
To filter a PolyLine, you'll need to define what you mean by "filter." Here are a few common methods for filtering Polylines:

Filter by length: You can filter Polylines based on their length. For example, you might want to keep only Polylines that are longer than a certain threshold, or you might want to remove Polylines that are shorter than a certain threshold.

Filter by proximity: You can filter Polylines based on their proximity to other features or locations. For example, you might want to keep only Polylines that are within a certain distance of a particular point or feature.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -601,6 +601,9 @@ class GeoJson(Layer):
             {% if this.style %}
                 style: {{ this.get_name() }}_styler,
             {%- endif %}
+            {% if this.tags %}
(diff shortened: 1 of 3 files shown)


martinfleis:
The code contribution is a duplicate of #1756 and the text above seems to be irrelevant to what the code actually does. Closing.

PR #1760, opened 2023-05-17
Draggable layer control

Conengmo:
Add an option to drag the layer control around. I'm not sure how useful this is to have, but I thought to share it here anyway. Let's have a think if this is something we want to merge.


Conengmo:
Thanks! I'll add a test before merging.

PR #1761, opened 2023-05-17
Repair tests

Conengmo:
Seems like a new pandas version got out that deprecated some things in three notebook. Let's see if everything runs okay now.

- Altair 5.0.0 has a fix to work with the new Pandas version, so pin that.
- New Altair version doesn't have data built in, use the vega_datasets package instead.
- Some notebooks needed some syntax fixes to work with the new Pandas version.
- Don't test the code of examples/WmsTimeDimension.ipynb, the external resource makes it flaky. We already skip it in the Selenium tests.


Conengmo:
There’s an issue with installing the packages on Python 3.7. Since that version will be end of line soon, I’ll upgrade the workflow to 3.8.

Conengmo:
Everything passes! 🚀

PR #1793, opened 2023-08-25
FIX #1787 Bump Jquery Version

### LINKED ISSUE #1787: Update jquery
The default version of jquery used by folium is from 12 years ago and has several known vulnerabilities that make it unsuitable for use in production websites

https://security.snyk.io/package/npm/jquery/1.12.4

devbijay:


Conengmo:
I appreciate the effort, but unfortunately more work is needed. This change doesn't get all the definitions of jquery we use. Also, we need someone to verify a couple of features don't break with the new version. See https://github.com/python-visualization/folium/issues/1787#issuecomment-1678618696. So I'll close this PR. Help is still welcome, but unfortunately this change is more complex than most.

PR #1802, opened 2023-09-08
Update changelog for 1.0 release

Conengmo:
@ocefpaf let's do a 1.0 release! If you agree, here's what I think we should do:

I will:

- Add a changelog for Branca.
- Merge this PR.

Afterwards, you can:
- Make a Branca 0.7.0 release.
- Make a Folium 1.0.0 release.

What are your thoughts on this?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,46 @@
+1.0.0
+~~~~~
+
+Folium has been stable for a couple of years now. We hardly have breaking changes and the feature
+set is stable. We don't expect the scope of what Folium is and isn't to change significantly. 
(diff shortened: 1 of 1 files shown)


martinfleis:
Shall we include #1793 and possibly https://github.com/python-visualization/branca/pull/134 before releasing?

ocefpaf:
It is a long weekend here but I'll try to get to this ASAP. If not, next Monday.

martinfleis:
I guess that we should resolve #1803 before releasing as well.

Conengmo:
I'll use these when we make a 0.15.0, but this PR can be closed.

PR #1818, opened 2023-10-16
Update leaflet.fullscreen to 3.0.0

### LINKED ISSUE #1687: Click twice to exit full screen
**Describe the bug**
When I enter full screen mode, I need to click twice to exit full screen mode

afontenot:
Fixes #1687.

Note: the CDN was changed from cdnjs to jsdelivr because an issue with the former means that the fullscreen icon is broken for all versions >= 2.2.0 and <= 3.0.0.


Conengmo:
Thanks @afontenot!

PR #1827, opened 2023-11-06
use xyzservices instead of templates

martinfleis:
This removes built-in tile management with the template files and adds `xyzservices` as a hard dependency that now manages all tiles. It is fully backwards compatible with all the benefits of a large library in `xyzservices` and flexibility of `query_name`, allowing custom characters and cases in the input. E.g. all of these will work

```py
"CartoDB Positron"
"cartodbpositron"
"cartodb-positron"
"carto db/positron"
"CARTO_DB_POSITRON"
"CartoDB.Positron"
```

### DIFF
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Optional, Union
 
 import xyzservices
 from branca.element import Element, Figure
-from jinja2 import Environment, PackageLoader, Template
+from jinja2 import Template
 
 from folium.map import Layer
 from folium.utilities import (
@@ -16,8 +16,6 @@ from folium.utilities import (
     mercator_transform,
     parse_options,
 )
-
-ENV = Environment(loader=PackageLoader("folium", "templates"))
 
 
 class TileLayer(Layer):
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -128,8 +128,12 @@ class TileLayer(Layer):
                 name = tiles.name.replace(".", "").lower()
             tiles = tiles.build_url(fill_subdomain=False, scale_factor="{r}")  # type: ignore
 
-        self.tile_name = name if name is not None else "".join(tiles.lower().strip().split())
-        super().__init__(name=self.tile_name, overlay=overlay, control=control, show=show)
+        self.tile_name = (
+            name if name is not None else "".join(tiles.lower().strip().split())
+        )
+        super().__init__(
+            name=self.tile_name, overlay=overlay, control=control, show=show
+        )
         self._name = "TileLayer"
 
         self.tiles = tiles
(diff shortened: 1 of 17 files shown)


ocefpaf (approved):
Thanks!

folium/raster_layers.py
```diff
@@ -16,10 +17,6 @@
     parse_options,
 )
 
-if TYPE_CHECKING:
-    import xyzservices
-
-
 ENV = Environment(loader=PackageLoader("folium", "templates"))
```
Conengmo, line 20: we can get rid of this one as well if I'm not mistaken

folium/raster_layers.py
```diff
@@ -104,14 +106,26 @@ def __init__(
...
+                tiles = "OpenStreetMap Mapnik"
+                if name is None:
+                    name = "openstreetmap"
+            try:
+                tiles = xyzservices.providers.query_name(tiles)
+            except ValueError:
+                # no match, likely a custom URL
+                pass
+
+        if isinstance(tiles, xyzservices.TileProvider):
```
Conengmo, line 122: question, any particular reason to not use `elif` here?

folium/raster_layers.py
```diff
@@ -104,14 +106,26 @@ def __init__(
...
             attr = attr if attr else tiles.html_attribution  # type: ignore
             min_zoom = tiles.get("min_zoom", min_zoom)
             max_zoom = tiles.get("max_zoom", max_zoom)
             subdomains = tiles.get("subdomains", subdomains)
+            if name is None:
+                name = tiles.name.replace(".", "").lower()
```
Conengmo, line 128: are you following Folium convention here or is this another convention to create a name like this? I wonder if this is a chance to have a better name where we don't mush all words together in lower case. What do you think?

Conengmo (commented):
very happy with this change! I added a couple comments, but nothing stopping this.

folium/raster_layers.py
martinfleis: A string is used as an input to `xyzservices.providers.query_name` on l. 117 which produces TileProvider that needs to be parsed here. `elif` would not work.

folium/raster_layers.py
martinfleis: I was just ensuring the name remains the same it was. I am happy to use proper name that comes from xyzservices, i.e. `"CartoDB Positron` if backwards compatibility of this particular bit is not important.

folium/raster_layers.py
Conengmo: That makes sense, thanks!

folium/raster_layers.py
Conengmo: Good to know. I don't really have an overview of what the impact of changing the name format would be. This is used as JS object name as well if I'm not mistaken, so maybe better to not change it lightly. Let's indeed keep it then as it is, we can look into improving that later.

martinfleis:
I think that this should be ready now.

Conengmo (approved):
Let's go 🚀

PR #1834, opened 2023-11-21
Fix streamlit-folium incompatibility (add layer to map with new class)

Conengmo:
Fix the compatibility issue between Folium 0.15.0 and streamlit-folium, see https://github.com/randyzwitch/streamlit-folium/issues/148.

When adding layers to the map, don't add the relevant code to the figure object. Instead, add a new class instance to the Layer instance. That way it will get rendered even if only the map and not the figure is being rendered. This is also maybe a bit nicer, since it's more in line with how the rest of Folium works.

### DIFF
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -1,6 +1,8 @@
 from typing import List, Tuple
 
-from branca.element import CssLink, Element, Figure, JavascriptLink
+from jinja2 import Template
+
+from branca.element import CssLink, Element, Figure, JavascriptLink, MacroElement
 
 
 class JSCSSMixin(Element):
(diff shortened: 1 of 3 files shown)


BastienGauthier:
Works for me !

Conengmo:
pre-commit.ci autofix

BastienGauthier:
@ocefpaf, if you have time to do the release before the end of next week it could help me quite a lot for my project actually :)
Thanks !

Conengmo:
I think all relevant code to publish to PyPI is in our Github Actions, so I'll give it a try to bake a release later today.

ocefpaf:
I can do this today and I'll write a guide so others can do it in my absence.

PR #1836, opened 2023-11-21
GeoJSON: option to keep layer highlighted when popup is open

GiuseppeGalilei:
### Problem
Currently, the highlighting option available for GeoJson layers is active only when hovering with the mouse above the layer.
I'm working on a project where it is useful to keep the highlighting active after the user clicks on a layer, a popup is opened and the mouse is moved around. Then simply closing the popup will deactivate the highlighting.

### Solution
To achieve so, the boolean parameter "popup_keep_highlighted" has been added to the "folium.GeoJson" function, and a test has also been implemented to check if the user tries to use it without a popup, in which case an error is raised.
Then, based on this boolean parameter, the template gets slightly modified when dealing with the different mouse and popup events.
The standard behaviour of highlighting when hovering remains untouched.
But when `popup_keep_highlighted=True`: the `popupopen` and `popupclose` event control activation and deactivation of the highlighting, while `mouseout` deactivates the highlighting only if the popup is closed and `mouseover` remains untouched,

### Example of usage
```
        layer = folium.GeoJson(
                       path,
                       name = name,
                       highlight_function=lambda feature: {
                            "color": ("red"),
                            "opacity": 1,
                            "weight": 6
                            },
                        popup_keep_highlighted=True,
                        popup=popup,
                        style_function=lambda feature: {
                            "color": "black",
                            "weight": 4,
                            "opacity":0.5,
                        },
                  )
```  
### Demo
You can see the user first hovering over the track, then clicking on it to display the popup and moving the mouse elsewhere, to showcase the different behaviour.

https://github.com/python-visualization/folium/assets/44207906/37c28544-a137-4b9f-8bcb-a3e84567a39e

https://github.com/python-visualization/folium/assets/44207906/fbbd764a-6e3b-42e7-ba5a-76a3d8991bf0

It's my first pull request and I don't have much experience with this project, any advice is warmly welcomed.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -689,8 +689,7 @@ class GeoJson(Layer):
                     "Only Marker, Circle, and CircleMarker are supported as GeoJson marker types."
                 )
 
-        if popup_keep_highlighted:
-            if popup is None:
+        if popup_keep_highlighted and popup is None:
                 raise NameError(
                     "A popup is needed to use the popup_keep_highlighted feature"
                 )
--- a/folium/features.py
+++ b/folium/features.py
@@ -569,7 +569,7 @@ class GeoJson(Layer):
 
         function {{this.get_name()}}_onEachFeature(feature, layer) {
             layer.on({
-                {%- if this.highlight %}
+                {%- if this.highlight and not this.popup_keep_highlighted %}
                 mouseout: function(e) {
                     if(typeof e.target.setStyle === "function"){
                         {{ this.get_name() }}.resetStyle(e.target);
@@ -582,6 +582,34 @@ class GeoJson(Layer):
                     }
                 },
                 {%- endif %}
+
+                {%- if this.highlight and this.popup_keep_highlighted %}
+                mouseout: function(e) {
+                    if(typeof e.target.setStyle === "function"){
+                        if(!{{ this.get_name() }}.isPopupOpen()){
+                            {{ this.get_name() }}.resetStyle(e.target);
+                        }
+                    }
+                },
+                mouseover: function(e) {
+                    if(typeof e.target.setStyle === "function"){
+                        const highlightStyle = {{ this.get_name() }}_highlighter(e.target.feature)
+                        e.target.setStyle(highlightStyle);
+                    }
+                },
+                popupopen: function(e) {
+                    if(typeof e.target.setStyle === "function"){
+                        const highlightStyle = {{ this.get_name() }}_highlighter(e.target.feature)
+                        e.target.setStyle(highlightStyle);
+                    }
+                },
+                popupclose: function(e) {
+                    if(typeof e.target.setStyle === "function"){
+                        {{ this.get_name() }}.resetStyle(e.target);
+                    }
+                },
+                {%- endif %}
+
                 {%- if this.zoom_on_click %}
                 click: function(e) {
                     if (typeof e.target.getBounds === 'function') {
@@ -632,6 +660,7 @@ class GeoJson(Layer):
         data: Any,
         style_function: Optional[Callable] = None,
         highlight_function: Optional[Callable] = None,
+        popup_keep_highlighted: bool = False,
         name: Optional[str] = None,
         overlay: bool = True,
         control: bool = True,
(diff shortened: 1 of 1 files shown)


Conengmo:
Interesting! I think your PR is really clear, so no comments there now. I’ll review it shortly.

folium/features.py
```diff
@@ -582,6 +582,34 @@ class GeoJson(Layer):
                     }
                 },
                 {%- endif %}
+
+                {%- if this.highlight and this.popup_keep_highlighted %}
```
Conengmo, line 586: Is it better if we merge the two declarations of mouseout and mouseover? Then we would have less duplicate code. For example:

```
{%- if this.highlight %}
mouseout: function(e) {
    if(typeof e.target.setStyle === "function"){
        {%- if this.popup_keep_highlighted %}
        if (!{{ this.get_name() }}.isPopupOpen())
        {%- endif %}
            {{ this.get_name() }}.resetStyle(e.target);
    }
},
mouseover: function(e) {
    if(typeof e.target.setStyle === "function"){
        const highlightStyle = {{ this.get_name() }}_highlighter(e.target.feature)
        e.target.setStyle(highlightStyle);
    }
},
{%- endif %}
```

folium/features.py
```diff
@@ -659,6 +688,14 @@ def __init__(
                 raise TypeError(
                     "Only Marker, Circle, and CircleMarker are supported as GeoJson marker types."
                 )
+
+        if popup_keep_highlighted:
+            if popup is None:
+                raise NameError(
```
Conengmo, line 694: not sure if `NameError` is the right class here. Docs say it's for missing variables, which doesn't apply here: the variable exists, it just doesn't have the right value. I'd say go with a `ValueError` instead.

Conengmo (changes requested):
I think this is a nice addition. I've added some comments to hopefully sharpen the change a little bit. Do you have time to look into those?

Optionally, you could also consider adding an example to our documentation. But it's not required to get this PR merged.

GiuseppeGalilei:
Ok, thank you!
I'll fix the problems you pointed out, shortly. 
I'll be happy to contribute to the documentation, however you'll need to wait a few days for that.

Conengmo:
> I'll be happy to contribute to the documentation, however you'll need to wait a few days for that.

that’s great! Alright let’s not let that hold up this PR then though.

GiuseppeGalilei:
Done, can you review it?

GiuseppeGalilei:
Sorry for the bug. For my use case I build the map in a sligthly different way and I didn't consider this possibility. I'll take a look at it!

Conengmo:
That makes sense! I think your solution is good. Now I want to look whether it's possibly better to change something in `GeoJsonPopup`. There we bind the popup to the entire geojson, and put logic per element in the function. I'm curious if it's better to bind the popup to each element to begin with. So we don't have to do that post-hoc. 

If you're interested you could take a look at that. I don't know how you are timewise, but you're very welcome to. Otherwise I'll take a look myself.

GiuseppeGalilei:
Yeah, to be honest I had the same idea.
I tried to look into that, but it seemed something that could impact quite a bit of codebase and maybe project structure. 
So, given that I'm fairly new to the project and especially to javascript, I decided to adopt the cleanest and simplest approach (codewise).

Unfortunately I'm quite busy at the moment.
Maybe, given that you know the project throughly, you can brainstorm your ideas here and I'll try to give you some inputs if I can. I'm very curious to hear your insights, especially regarding how to integrate changes while keeping the design philosophy consistent.

Conengmo:
> Yeah, to be honest I had the same idea.
I tried to look into that, but it seemed something that could impact quite a bit of codebase and maybe project structure.
So, given that I'm fairly new to the project and especially to javascript, I decided to adopt the cleanest and simplest approach (codewise).

That makes a lot of sense, and I can totally agree with that. Let's say I'll give it a look over just to make sure. That will inform whether we go ahead with this approach and/or we add some more changes.

Conengmo:
@GiuseppeGalilei I looked into this a bit, and it seems pretty straight forward to bind the popups to each feature instead of the entire geojson. I made a PR, could you maybe review it? https://github.com/python-visualization/folium/pull/1844

Also, about this PR here, I'm now thinking maybe this shouldn't be a parameter, but we should just enable this behavior by default. Is there a use-case for keeping the old behavior? In what case would someone not want the highlight enabled when a popup is open?

GiuseppeGalilei:
#1844 progress will inform how to continue developing this.
As of now, it removes the 'bind/unbind' necessity for `geojson` but leaves behind regular `popups`, which still need the older method. It would be better to uniform this behaviour.

Conengmo:
Thanks for your contribution @GiuseppeGalilei! I'm happy we got to explore this problem and I'm also happy with your solution. Happy holidays!

GiuseppeGalilei:
Thank you @Conengmo! It was my first pull request and I really enjoyed the whole process. Happy holidays to you!
In the coming days I'll try to write the relative documentation.

Issue #1233, opened 2019-11-25
Unable to remove markers or polylines added to the Map


prashanthgpkr:
I am currently using Folium (0.10.0) to create a Map for a given data. In my specific use case, I would like to add markers and then delete the specific markers as the data gets updated. This also should work with respective polylines that are plotted between these markers.

I have been looking for the appropriate way to do this, but not able to find any solution so far. Could any one please let me know, if there is any way to solve this issue?

Conengmo:
I'm afraid folium is not really suitable for this. It's a one way street from Python to Javascript. Dynamically updating your map when your data in Python changes is not possible, other than rerendering the entire map.

You could check it this other library called ipyleaflet, it focuses more on interactivity in Jupyter notebooks: https://github.com/jupyter-widgets/ipyleaflet

pstatonwx:
@Conengmo Just curious if any progress has been made on this since posting - hoping it's possible now :)

Issue #848, opened 2018-04-23
Dynamicaly change/update markers


siezyj:
I`m want to make webapp with Django which displays a map for every user on the main page. What I want to have:

1.Every user can add a marker to map and every other user see those markers immediately.
2.Every user can delete marker if he is owner of that marker and other user will see those changes immediately.
3.If user A "is intrested" with user B (this information is stored in DB) he (A) sees markers of user B in a different way than others ( different color or marker image).

My idea is to create once a html file without markers and creat separate function which will be updateing map for every user separetly. I want to use Memecahce to dynamic update the map. If somebody add new marker, specific  function shoudl add new marker without makeing and reolading whole map. I need that becouse application can have a lot of user with a lot of markers so it could be too expensive for server(reolading all markers after every update).

My question : How to change/add markers to map dynamicaly after creating a map file or maybe you have better idea to solve that problem.

Conengmo:
Sounds like an interesting project! In short: unfortunately folium cannot do this, though it could be an interesting addition.

It's import to understand that folium generates the html with [Leaflet](http://leafletjs.com/) javascript and after that is finished. So anything that happens on the map after generation is outside of folium's scope.

Now there have been ideas of getting callbacks from javascript back to python, mainly for in the Jupyter environment. I can imagine something like that could also work for Flask or Django, where some javascript ajax calls retrieve new data. This is not available in folium though, but of course you're welcome to work on adding this.

folium works great for creating ad hoc maps with python data. For your application I would suggest not to use folium but to use Leaflet directly. You need a fixed map layout and very good integration with your web application, not to create various maps with differing Python data, which is folium's strong point. Of course you can use a map created by folium as inspiration for how to use Leaflet of course.

But that's my opinion. Hope it helps! Maybe others will chime in as well.

Conengmo:
A quick search shows that there is a lot written about for example [dynamic data for Leaflet](http://leafletjs.com/plugins.html#dynamiccustom-data-loading). Good luck with this! I'll close the issue if that's alright with you, feel free to reopen if you have something that could be done about folium. Also, it would be great if you could add any solutions you come up with here and help others who are coming from folium and want to do a similar thing.

Issue #735, opened 2017-09-26
Plugins ideas for a sprint


ocefpaf:
- [ ] https://github.com/turban/Leaflet.Graticule
- [ ] https://github.com/cloudybay/leaflet.latlng-graticule
- [ ] https://github.com/marslan390/BeautifyMarker
- [ ] https://github.com/coryasilva/Leaflet.ExtraMarkers
- [ ] https://github.com/calvinmetcalf/leaflet-ajax
- [ ] https://github.com/stuartmatthews/leaflet-geotiff
- [ ] https://github.com/IHCantabria/Leaflet.CanvasLayer.Field
- [ ] https://github.com/aparshin/leaflet-GIBS
- [ ] https://github.com/turban/Leaflet.Zoomify
- [ ] https://github.com/commenthol/leaflet-rastercoords
- [ ] https://github.com/commenthol/gdal2tiles-leaflet
- [ ] https://github.com/makinacorpus/Leaflet.FileLayer
- [ ] https://github.com/mapbox/leaflet-omnivore
- [ ] https://github.com/Flexberry/Leaflet-WFST
- [ ] https://github.com/daniellsu/leaflet-betterscale
- [ ] https://github.com/perliedman/leaflet-realtime
- [ ] https://github.com/makinacorpus/Leaflet.LineExtremities
- [ ] https://github.com/teastman/Leaflet.pattern
- [ ] https://github.com/albburtsev/Leaflet.geojsonCSS

AndyChallis1:
Just looked at https://github.com/albburtsev/Leaflet.geojsonCSS - looks similar to what I have just submitted for a pull request but in the TimestampedGeoJson Plugin. Happy to implement this wider using their layout

ocefpaf:
@ghandic That would be nice! Do you want to submit in the same PR or do you want to close that and send a new one?

ghandic:
I’ll do it in another pull request, also looking into the leaflet-search, I’m trying to maintain short branches so will do these in separate PR’s

Sent from my iPhone

On 13 Oct 2017, at 17:49, Filipe <notifications@github.com<mailto:notifications@github.com>> wrote:


@ghandic<https://github.com/ghandic> That would be nice! Do you want to submit in the same PR or do you want to close that and send a new one?

ocefpaf:
Awesome!

ghandic:
[...]

Issue #127, opened 2015-05-13
List of plugins that we would like to see in folium


ocefpaf:
- Leaflet.TimeDimension
- http://esri.github.io/esri-leaflet/
- https://github.com/aparshin/leaflet-GIBS (http://aparshin.github.io/leaflet-GIBS/examples/#2/18.3/13.4)
- https://github.com/heigeo/leaflet.wms
- https://github.com/joergdietrich/Leaflet.Terminator
- https://github.com/perliedman/leaflet-realtime
- http://arthur-e.github.io/Wicket/
- https://github.com/mapbox/leaflet-omnivore
- https://github.com/makinacorpus/Leaflet.FileLayer
- https://github.com/tomaszbrue/leaflet.boatmarker

themiurgo:
Great list, thanks! Isn't wicket redundant, considering Leaflet.Omnivore handles WKT?

birdage:
@ocefpaf ive worked with esri leaflet and wms and i think they would be easy to add, we might need some more functions and classes to support them tho

ocefpaf:
> wicket

I am not a `Wicket` user, but a colleague recommend it to me.  We can probably cross it out.

@birdage Adding just the layer is easy and I do not thing we should support that.  I put it there just remind myself to read the docs and check what else ESRI has to offer.

birdage:
@ocefpaf wicket is great! there are a whole host of esri options that are supported as part of the layer i.e queries.

ocefpaf:
Interesting.  Lets keep it there then!

ocefpaf:
I am closing this for now. I have a lot of work to do already regarding the plugins.

PR #1848, opened 2023-12-16
Add leaflet-realtime plugin

hansthen:
Based on: https://github.com/perliedman/leaflet-realtime

A current limitation of folium is that it is not possible to perform dynamic updates to an existing map (see e.g. #848, #1581, #1233). A way to overcome that limitation is to use the leaflet-realtime plugin. This plugin allows polling for a url that provides geojson data. 

I saw several suggestions in the past to implement this plugin in folium (see #735 and #127), so I decided to give it a go.

I did not test the full plugin functionality yet; just the basic `url` lookup of a static geojson datasource.  Most notably, the `container` parameter is not implemented. I am unsure what the best way would be to implement this. Should it refer to a folium object or should it refer to a javascript object?

If there is any interest in this plugin, I can add more tests and implement the `container` parameter. 

I also added a `JsCode` utility wrapper around strings. The idea was stolen from several streamlit plugins, that use a similar class to pass javascript functions from python to javascript. I mainly used it because I was too lazy to implement all the function parameters from `L.realtime` and `L.geojson` in the template.

### DIFF
--- a/folium/plugins/realtime.py
+++ b/folium/plugins/realtime.py
@@ -2,7 +2,7 @@ from branca.element import MacroElement
 from jinja2 import Template
 
 from folium.elements import JSCSSMixin
-from folium.utilities import JsCode, parse_options
+from folium.utilities import JsCode, camelize, parse_options
 
 
 class Realtime(JSCSSMixin, MacroElement):
@@ -13,33 +13,45 @@ class Realtime(JSCSSMixin, MacroElement):
 
     Parameters
     ----------
+    source :
+        The source can be one of:
+        * a string with the URL to get data from
+        * a dict that is passed to javascript's `fetch` function
+          for fetching the data
+        * a folium.utilities.JsCode object in case you need more freedom.
     start : bool, default True
         Should automatic updates be enabled when layer is added
         on the map and stopped when layer is removed from the map
     interval : int, default 60000
         Automatic update interval, in milliseconds
-    getFeatureId : function, default returns `feature.properties.id`
+    get_feature_id : folium.utilities.JsCode
+        A function with a geojson `feature` as parameter
+        default returns `feature.properties.id`
         Function to get an identifier uniquely identify a feature over time
-    updateFeature : function
+    update_feature : folium.utilities.JsCode
+        A function with a geojson `feature` as parameter
         Used to update an existing feature's layer;
         by default, points (markers) are updated, other layers are discarded
         and replaced with a new, updated layer.
         Allows to create more complex transitions,
         for example, when a feature is updated
-    removeMissing : bool, default False
+    remove_missing : bool, default False
         Should missing features between updates been automatically
         removed from the layer
 
     Other parameters are passed to the GeoJson layer, so you can pass
-        `style`, `pointToLayer` and/or `onEachFeature`.
+        `style`, `point_to_layer` and/or `on_each_feature`.
 
     Examples
     --------
     >>> from folium.utilities import JsCode
-    >>> m = folium.Map()
+    >>> m = folium.Map(location=[40.73, -73.94], zoom_start=12)
     >>> rt = Realtime(
-    ...     "https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_10m_geography_regions_elevation_points.geojson",
-    ...     getFeatureId=JsCode("function(f) { return f.properties.name; }"),
+    ...     "https://raw.githubusercontent.com/python-visualization/folium-example-data/main/subway_stations.geojson",
+    ...     get_feature_id=JsCode("(f) => { return f.properties.objectid; }"),
+    ...     point_to_layer=JsCode(
+    ...         "(f, latlng) => { return L.circleMarker(latlng, {radius: 8, fillOpacity: 0.2})}"
+    ...     ),
     ...     interval=10000,
     ... )
     >>> rt.add_to(m)
@@ -52,8 +64,13 @@ class Realtime(JSCSSMixin, MacroElement):
             {% for key, value in this.functions.items() %}
             options["{{key}}"] = {{ value }};
             {% endfor %}
+
             var {{ this.get_name() }} = new L.realtime(
+            {% if this.src is string or this.src is mapping -%}
                 {{ this.src|tojson }},
+            {% else -%}
+                {{ this.src.js_code }},
+            {% endif -%}
                 options
             );
             {{ this._parent.get_name() }}.addLayer(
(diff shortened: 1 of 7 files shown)


Conengmo:
Hi Hans, good that you're asking. I'm busy at work with the end of the year, but I'll take a proper look after this week is over! Thanks for your PR!

StohanzlMart:
Dear Saints of the Open Source Community,

jesus

Thank you for making the world more awesome!

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+    default_js = [
+        (
+            "Leaflet_Realtime_js",
+            "https://cdnjs.cloudflare.com/ajax/libs/leaflet-realtime/2.2.0/leaflet-realtime.js",  # NoQA
+        )
+    ]
+
+    def __init__(self, src, **kwargs):
```
Conengmo, line 72: I'd rather explicitly name all the arguments here that are also in the docstring. That way it's more transparent for users and IDE's what the arguments of this function are. We can add type hints for each argument.

The JS functions are already added to `options` separately in the JS part, so that can still happen there.

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Parameters
+    ----------
+    start : bool, default True
+        Should automatic updates be enabled when layer is added
+        on the map and stopped when layer is removed from the map
+    interval : int, default 60000
+        Automatic update interval, in milliseconds
+    getFeatureId : function, default returns `feature.properties.id`
+        Function to get an identifier uniquely identify a feature over time
```
Conengmo, line 22: the type of these is `str`, not `function`. The user is expected to pass a Javascript function as a Python string. Good to put that here more clearly. Also what the argument of the function is.

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+    updateFeature : function
```
Conengmo, line 23: also not a function but a str

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+    Based on: https://github.com/perliedman/leaflet-realtime
+
+    Parameters
+    ----------
+    start : bool, default True
+        Should automatic updates be enabled when layer is added
+        on the map and stopped when layer is removed from the map
+    interval : int, default 60000
+        Automatic update interval, in milliseconds
+    getFeatureId : function, default returns `feature.properties.id`
```
Conengmo, line 21: in Folium we use snake_case convention for our arguments, even though the Javascript counter parts use camelCase. Please convert these to snake_case. The `parse_options` function converts these to camelCase. Or, for the functions, you can use the `camelize` function later on.

folium/plugins/realtime.py
Conengmo, line 72: the `src` argument is not listed in the docstring.

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+                self.functions[key] = value.js_code
+
+        # and remove them from kwargs
+        for key in self.functions:
+            kwargs.pop(key)
+
+        # the container is special, as we
+        # do not allow it to be set (yet)
+        # from python
+        self.options = parse_options(container=None, **kwargs)
```
Conengmo, line 90: IMO no need to explicitly add this comment and `container=None`. We don't generally list out every feature we don't support. If you want, you can add a note in the docstring above the parameters, that way it will show up in the documentation.

folium/plugins/realtime.py
```diff
@@ -0,0 +1,90 @@
...
+            var options = {{this.options|tojson}};
+            {% for key, value in this.functions.items() %}
+            options["{{key}}"] = {{ value }};
+            {% endfor %}
+            var {{ this.get_name() }} = new L.realtime(
+                {{ this.src|tojson }},
+                options
+            );
+            {{ this._parent.get_name() }}.addLayer(
+                {{ this.get_name() }}._container);
```
Conengmo, line 60: In the example the use `L.realtime().addTo(map)`. Why use `_container` here instead? Why not

```
{{ this.get_name() }}.addTo({{ this._parent.get_name() }}
```

Conengmo (changes requested):
This is a nice plugin to have. I think it's okay to not support the `container` argument in this first version, we can always add that in a next one.

This feature could really use an example in our documentation. The demo on https://github.com/perliedman/leaflet-realtime is actually broken for me :( Can we add a good example to our documentation?

The examples in our documentation also function as tests. But separate tests are welcome as well, though not a hard requirement.

StohanzlMart:
> I will work on better documentation. Would it be okay to add a static geojson to the repo? It would not really be realtime, but at least it will not break the tests.

Please forgive my ignorance, as I have no clue what I'm talking about. Could you use something like that for testing?

```python
def get_random_location_nearby(base_location, max_deviation=0.02):
    """
    Generate a random location near the base location.
    """
    lon, lat = base_location
    new_lon = lon + random.uniform(-max_deviation, max_deviation)
    new_lat = lat + random.uniform(-max_deviation, max_deviation)
    return [new_lon, new_lat]
```
Thank you for your endurance in implementing!

Conengmo:
I’m on mobile so only responding to the test data thing: we have a repo with example data at https://github.com/python-visualization/folium-example-data. You can use a file from there or add a new one if you want.

Conengmo:
> Alternatively, we could spin up a small webserver to serve geojson during testing, but that seems like overkill.

Interesting thought. We don’t have a webserver we can run things on. We do host our documentation on GitHub Pages, could that include something like that?

Conengmo:
Looks good! This is an interesting plugin to have. Especially like the option to add a custom source, I was playing around a bit with that, lots of flexibility. I did a test with random lat lons and that worked perfectly. 

Really good documentation you added also. Love the multiple examples.

Rather than doing another round of reviews to dot the i's, I took the liberty of making some changes on your branch myself. Mostly to get the documentation right. You can see the commits, I'll elaborate here as well:

- Don't define an `options` variable name in JS, since that may conflict with other instances. Use a unique name instead.
- Use default arguments in the Python side and add type hints.
  You mentioned not wanting to shadow the leaflet plugin default arguments, but to give a good experience to our users we've   opted throughout Folium to do shadow them in most cases.
- Update the JsCode docstring and our autodocs to have it show up in the documentation.
- Remove the 'parameters' section from the Realtime docs, since it's a copy of the docstring which is included in the API reference.
- Some small tweaks to the Realtime docs, like adding section headers and code indentation.

With this I'm ready to merge this PR. Do you want to take a look, see if you agree with these changes, or if you have any comments or issues? Let me know.

Conengmo (approved):

Conengmo:
Perfect! Thanks for your contribution @hansthen!

PR #1852, opened 2023-12-24
Geocoder plugin: support other built-in providers

prusswan:
tested with Nominatim, Photon, Google:
```Python
from branca.element import MacroElement
from jinja2 import Template

from folium.elements import JSCSSMixin
from folium.utilities import parse_options

import folium

class EnhancedGeocoder(JSCSSMixin, MacroElement):
    """A simple geocoder for Leaflet that by default uses OSM/Nominatim.

    Please respect the Nominatim usage policy:
    https://operations.osmfoundation.org/policies/nominatim/
    Parameters
    ----------
    collapsed: bool, default False
        If True, collapses the search box unless hovered/clicked.
    position: str, default 'topright'
        Choose from 'topleft', 'topright', 'bottomleft' or 'bottomright'.
    add_marker: bool, default True
        If True, adds a marker on the found location.
    geocode_zoom: int, default 11
        Set zoom level used for displaying the geocode result, note that this only has an effect when add_marker is set to False. Set this to None to preserve the current map zoom level.
    geocode_provider: str, default 'nominatim'
        Defaults to "nominatim", see https://github.com/perliedman/leaflet-control-geocoder/tree/2.4.0/src/geocoders for other built-in providers.
    geocode_provider_options: dict, default {}
        For use with specific providers that may require api keys or other parameters.

    For all options see https://github.com/perliedman/leaflet-control-geocoder

    """

    _template = Template(
        """

            var geocoderOpts_{{ this.get_name() }} = {{ this.options|tojson }};

            // note: geocoder name should start with lowercase
            var geocoderName_{{ this.get_name() }} = geocoderOpts_{{ this.get_name() }}["geocodeProvider"];

            var customGeocoder_{{ this.get_name() }} = L.Control.Geocoder[ geocoderName_{{ this.get_name() }} ](
                geocoderOpts_{{ this.get_name() }}['geocodeProviderOptions']
            );
            geocoderOpts_{{ this.get_name() }}["geocoder"] = customGeocoder_{{ this.get_name() }};

            L.Control.geocoder(
                geocoderOpts_{{ this.get_name() }}
            ).on('markgeocode', function(e) {
                var zoom = geocoderOpts_{{ this.get_name() }}['geocodeZoom'] || {{ this._parent.get_name() }}.getZoom();
                {{ this._parent.get_name() }}.setView(e.geocode.center, zoom);
            }).addTo({{ this._parent.get_name() }});

        {% endmacro %}
    """
    )

    default_js = [
        (
            "Control.Geocoder.js",
            "https://unpkg.com/leaflet-control-geocoder/dist/Control.Geocoder.js",
        )
    ]
    default_css = [
        (
            "Control.Geocoder.css",
            "https://unpkg.com/leaflet-control-geocoder/dist/Control.Geocoder.css",
        )
    ]

    def __init__(
        self,
        collapsed: bool = False,
        position: str = "topright",
        add_marker: bool = True,
        geocode_zoom: int | None = 11,
        geocode_provider: str = "nominatim",
        geocode_provider_options: dict = {},
        **kwargs
    ):
        super().__init__()
        self._name = "Geocoder"
        self.options = parse_options(
            collapsed=collapsed,
            position=position,
            default_mark_geocode=add_marker,
            geocode_zoom=geocode_zoom,
            geocode_provider=geocode_provider,
            geocode_provider_options=geocode_provider_options,
            **kwargs
        )

google_key = "placeholder"
gc2_default = EnhancedGeocoder() # nominatim
#gc2_photon = EnhancedGeocoder(geocode_provider="photon")
#gc2_google = EnhancedGeocoder(geocode_provider="google",
#                           geocode_provider_options={"apiKey": google_key})


m2 = folium.Map(location=[43.7798, 11.24148], zoom_start=13, tiles="cartodb positron")

gc2_default.add_to(m2)
#gc2_photon.add_to(m2)
#gc2_google.add_to(m2)

folium.LayerControl().add_to(m2)
m2
```

### DIFF
--- a/folium/plugins/geocoder.py
+++ b/folium/plugins/geocoder.py
@@ -1,10 +1,10 @@
+from typing import Optional
+
 from branca.element import MacroElement
 from jinja2 import Template
 
 from folium.elements import JSCSSMixin
 from folium.utilities import parse_options
-
-from typing import Optional
 
 
 class Geocoder(JSCSSMixin, MacroElement):
--- a/folium/plugins/geocoder.py
+++ b/folium/plugins/geocoder.py
@@ -19,6 +19,8 @@ class Geocoder(JSCSSMixin, MacroElement):
         Choose from 'topleft', 'topright', 'bottomleft' or 'bottomright'.
     add_marker: bool, default True
         If True, adds a marker on the found location.
+    geocode_zoom: int, default 11
+        Set zoom level used for displaying the geocode result, note that this only has an effect when add_marker is set to False. Set this to None to preserve the current map zoom level.
     geocode_provider: str, default 'nominatim'
         Defaults to "nominatim", see https://github.com/perliedman/leaflet-control-geocoder/tree/2.4.0/src/geocoders for other built-in providers.
     geocode_provider_options: dict, default {}
@@ -45,7 +47,8 @@ class Geocoder(JSCSSMixin, MacroElement):
             L.Control.geocoder(
                 geocoderOpts_{{ this.get_name() }}
             ).on('markgeocode', function(e) {
-                {{ this._parent.get_name() }}.setView(e.geocode.center, 11);
+                var zoom = geocoderOpts_{{ this.get_name() }}['geocodeZoom'] || {{ this._parent.get_name() }}.getZoom();
+                {{ this._parent.get_name() }}.setView(e.geocode.center, zoom);
             }).addTo({{ this._parent.get_name() }});
 
         {% endmacro %}
@@ -70,6 +73,7 @@ class Geocoder(JSCSSMixin, MacroElement):
         collapsed: bool = False,
         position: str = "topright",
         add_marker: bool = True,
+        geocode_zoom: int | None = 11,
         geocode_provider: str = "nominatim",
         geocode_provider_options: dict = {},
         **kwargs
@@ -80,6 +84,7 @@ class Geocoder(JSCSSMixin, MacroElement):
             collapsed=collapsed,
             position=position,
             default_mark_geocode=add_marker,
+            geocode_zoom=geocode_zoom,
             geocode_provider=geocode_provider,
             geocode_provider_options=geocode_provider_options,
             **kwargs
--- a/folium/plugins/geocoder.py
+++ b/folium/plugins/geocoder.py
@@ -4,6 +4,8 @@ from jinja2 import Template
 from folium.elements import JSCSSMixin
 from folium.utilities import parse_options
 
+from typing import Optional
+
 
 class Geocoder(JSCSSMixin, MacroElement):
     """A simple geocoder for Leaflet that by default uses OSM/Nominatim.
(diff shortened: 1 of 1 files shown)


folium/plugins/geocoder.py
```diff
@@ -19,6 +19,16 @@ class Geocoder(JSCSSMixin, MacroElement):
         Choose from 'topleft', 'topright', 'bottomleft' or 'bottomright'.
     add_marker: bool, default True
         If True, adds a marker on the found location.
+    geocodeProvider: str, default 'nominatim'
```
Conengmo, line 22: please use snake_case in the Python code. The `parse_options` function turns them into camelCase.

folium/plugins/geocoder.py
```diff
@@ -50,12 +155,24 @@ class Geocoder(JSCSSMixin, MacroElement):
         )
     ]
 
-    def __init__(self, collapsed=False, position="topright", add_marker=True, **kwargs):
+    def __init__(self, collapsed=False, position="topright", add_marker=True,
```
Conengmo, line 158: to satisify our linter these should all be on their own line. Please also add type hints. For example:

```
def __init__(
        self,
        collapsed: bool = False,
        position: str = "topright",
```

You can run the autoformatter/linter locally with precommit:

- Install precommit https://pre-commit.com/#install
- Enable precommit for this repo: `pre-commit install`
- Run precommit on all files: `pre-commit run --all-files`
  - or of course only on the files you touched

prusswan: Have applied the snake_case, type-hints and ran it through `pre-commit`. Please take a look when you have time.

Conengmo (changes requested):
Looks good overall! Few small comments before we can merge this.

folium/plugins/geocoder.py
```diff
@@ -50,12 +68,24 @@ class Geocoder(JSCSSMixin, MacroElement):
         )
     ]
 
-    def __init__(self, collapsed=False, position="topright", add_marker=True, **kwargs):
+    def __init__(
+        self,
+        collapsed: bool = False,
+        position: str = "topright",
+        add_marker: bool = True,
+        geocode_zoom: int | None = 11,
```
Conengmo, line 76: > as the default zoom level of 11 is not suitable for close-up locations (and only works when add_marker is False)

Question: is 11 a good default? Or would None be a better default?

And another question: should we check or correct this value when `add_marker` is False?

prusswan: 11 would preserve the current behavior so existing users would not be caught by surprise. The real gotcha would be the case when `add_marker`=True (when default handler in leaflet-control-geocoder will be using `fit_bounds` with no regard to current map zoom). What may not be obvious to normal folium users (with no knowledge of JS) is that both the default event handler for `markgeocode` as well as the one defined in `_template` would be triggered.

Not sure if there is a need to correct (beyond keeping the current default of 11). Can you explain more?

Conengmo (approved):

folium/plugins/geocoder.py
Conengmo: > Not sure if there is a need to correct (beyond keeping the current default of 11).

That's good then. It was more a question for you. We can keep it like it is now then.

Conengmo:
Thanks @prusswan! Nice addition!

PR #1854, opened 2024-01-02
 GeoJson: documentation for popup_keep_highlighted argument 

GiuseppeGalilei:
Related to #1836 
I added a small description along with an example, as a small section in the **Highlight function** paragraph.
I had some problems with building the documentation on my own.
However, I tested the example on my own and it is working correctly, so I believe there shouldn't be any problem.
Beware that, as I'm writing this, the popup_keep_highlighted feature has not yet been officially released.

### DIFF
--- a/docs/user_guide/geojson/geojson.md
+++ b/docs/user_guide/geojson/geojson.md
@@ -257,3 +257,27 @@ folium.GeoJson(
 
 m
 ```
+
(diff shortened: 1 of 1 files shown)


Conengmo (approved):
Looks good, thanks!

I'll fix the precommit complaint then merge it.

PR #1856, opened 2024-01-03
Process JS code separately [draft]

Conengmo:
Testing out how we can use the concept of the new `JsCode` class, which was added in https://github.com/python-visualization/folium/pull/1848. Idea is that we no longer have to manually put Javascript code in templates. 

Problem before was that a Javascript-function-as-Python-string would get transformed into a Javascript string. Because of that we had to put those functions into the template manually.

Now we have a `JsCode` class that we can use to mark strings as Javascript code. When putting stuff in templates, we can check for that class and prevent adding string quotes.

I looked into using a custom JSON encoder with the `json` module, but that turns out to be awkward, because what we're producing is not actually valid JSON.

So I've opted to make a custom filter like the `tojson` filter we use a lot, but now allowing for Javascript code. To register this filter I had to subclass `Template`.

Using it means:
- Importing the `Template` class from `folium.template`
- Wrapping any arguments that are JS functions with `JsCode.optional_create()`
- Stop using `parse_options`.
- Using `{{ options|tojavascript }}` filter in the template.

I've now applied this new workflow on the `LayerControl`, `MousePosition` and `Realtime` classes.

Some smaller included changes:
- Make `JsCode` a subclass of `str`, that makes it easier to put it in a template without additional needed code.
- Arguments that are JS functions can be either string or `JsCode`, I figured that's more user-friendly then requiring `JsCode`.
- The `MousePosition` example with the custom formatter no longer works with the trailing `;` character, so that's a breaking change.

If this seems like a good change I'll add tests as well.

Question is whether the added complexity of the new code weighs up to the more straightforward way in working with Javascript code.

### DIFF
--- a/docs/user_guide/plugins/mouse_position.md
+++ b/docs/user_guide/plugins/mouse_position.md
@@ -20,7 +20,7 @@ m
 ```{code-cell} ipython3
 m = folium.Map()
 
-formatter = "function(num) {return L.Util.formatNum(num, 3) + ' &deg; ';};"
+formatter = "function(num) {return L.Util.formatNum(num, 3) + ' &deg; '}"
 
 MousePosition(
     position="topright",
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -1,7 +1,8 @@
 from typing import List, Tuple
 
 from branca.element import CssLink, Element, Figure, JavascriptLink, MacroElement
-from jinja2 import Template
+
+from folium.template import Template
 
 
 class JSCSSMixin(Element):
--- a/folium/features.py
+++ b/folium/features.py
@@ -14,11 +14,11 @@ import requests
 from branca.colormap import ColorMap, LinearColormap, StepColormap
 from branca.element import Element, Figure, Html, IFrame, JavascriptLink, MacroElement
 from branca.utilities import color_brewer
-from jinja2 import Template
 
 from folium.elements import JSCSSMixin
 from folium.folium import Map
 from folium.map import FeatureGroup, Icon, Layer, Marker, Popup, Tooltip
+from folium.template import Template
 from folium.utilities import (
     TypeJsonValue,
     TypeLine,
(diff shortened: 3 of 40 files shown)


Conengmo:
Maybe I'll split off some concepts in this PR into separate PRs:

- Make `JsCode` a subclass of str.
- Allowing str type arguments in `RealTime`.

prusswan:
Slightly related: what are the recommended methods of string interpolation within `JsCode`? One issue I ran into: f-strings don't work so well for JS code since there will be too many curly braces to escape. I ended up using "%s" but there might be cases where some of the "%" needs to be escaped as well..

Conengmo:
>  what are the recommended methods of string interpolation within JsCode?

If escaping all those curly brackets is awkward, you could consider breaking up the string, and only apply the formatting on the part where you want to insert a variable. Example where I insert a custom id:

```
source = ("""
    function(responseHandler, errorHandler) {
        let geojson = {
            'type': 'FeatureCollection',
            'features': [{
                'type': 'Feature',
                'geometry': {
                    'type': 'Point',
                    'coordinates': [Math.random(), Math.random()]
                },
                'properties': {
                    'id': """ + f'{my_id}' + """
                }
            }]
        };
        responseHandler(geojson);
    }
""")
```

Conengmo:
Thanks for your comments @hansthen. 

> I thought it would be to big a change. It is cleaner though from the perspective of a plugin writer.

Well said, and I agree. I don't think this change is worth it. Maybe if we encounter this more in the future it could become worth it, but for now it's too much additional complexity for too little saved logic.

> On a side note will this pull request need to be finalized before you will make a new release containing the realtime plugin? 

No, that's not necessary. I split off a small idea from this PR into a separate PR (https://github.com/python-visualization/folium/pull/1862), maybe that would be nice to get processed before doing a release. I'll also check out https://github.com/python-visualization/folium/pull/1861. Afterwards, let's do a release.

PR #1858, opened 2024-01-08
Geojson: allow null geometry objects

hansthen:
GeoJson objects can have `null` `geometry` members according to the standard. However, if we pass a `null` value to folium, this will cause an error in `m.get_bounds()`. 

See https://datatracker.ietf.org/doc/html/rfc7946#section-3.2

A Feature object has a member with the name "geometry".  The value of the geometry member SHALL be either a Geometry object as defined above or . . . a JSON null value.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1120,7 +1120,8 @@ class GeoJsonDetail(MacroElement):
         geom_collections = [
             feature.get("properties") if feature.get("properties") is not None else key
             for key, feature in enumerate(self._parent.data["features"])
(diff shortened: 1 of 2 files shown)


Conengmo:
Thanks @hansthen! Nice enhancement, good PR. I wanted to merge it now already, but I was also thinking it would be nice to have a test case for this null case in `iter_coords`. That way we can prevent regressions in the future. If you have the time for it, that would be very welcome. But I didn't want to hold up this PR for that, so I already merged it. Thanks again!

PR #1861, opened 2024-01-18
Add interval support to timestamped_geo_json

hansthen:
This is a implementation of the leaflet.timeline plugin. See: https://skeate.dev/Leaflet.timeline. It has extra functionality above `Leaflet.TimeDimension` plugin currently used by `TimestampedGeojson`. One of these is the ability to specify both `begin` and `end` times for feature. I also liked the ability to show ticks under the timeslider.

This is still a work-in-progress, I just created this as a heads-up. And also to see if the maintainers are interested in a second plugin to support GeoJson with timestamps.

If you are okay with the general idea I will implement more parameters (and add documentation).

### DIFF
--- a/folium/plugins/timestamped_geo_json.py
+++ b/folium/plugins/timestamped_geo_json.py
@@ -1,59 +1,121 @@
 import json
+from typing import Optional, TextIO, Union
 
 from branca.element import MacroElement
 from jinja2 import Template
 
 from folium.elements import JSCSSMixin
 from folium.folium import Map
-from folium.utilities import get_bounds, parse_options
+from folium.utilities import JsCode, camelize, get_bounds, parse_options
 
 
 class TimestampedGeoJson(JSCSSMixin, MacroElement):
     """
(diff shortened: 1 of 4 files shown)


Conengmo:
If I'm correct the main difference with `TimestampedGeoJson` is that data points have a start time and an end time, whereas `TimestampedGeoJson` only allows a single timestamp. 

Two thoughts:

- I'd rather not have two plugins that have very similar, but slightly different functionality. That will make it harder for users to discover which Folium features they need. And gives us more code to maintain with little added benefit in terms of features.
- `TimestampedGeoJson` is not great in terms of maintainability, since it contains quite some Javascript logic that I'm not keen on.

I'm thinking, would it be possible to provide a single timeline/timestamped GeoJson plugin to our users? Maybe we can replace the internals of `TimestampedGeoJson` with your change, and add some Python logic to convert timestamps and durations to start and end dates?

Additionally, `TimestampedGeoJson` contains some features like popup support and `pointToLayer` definitions. Would we need to add the same in this new plugin?

What do you think?

Issue #1563, opened 2022-01-15
Make it possible for a marker to link to another marker


augusto-herrmann:
**Is your feature request related to a problem? Please describe.**
It may be a common use case that a marker might relate to other relevant makers. It would be great if a marker's popup could contain links to other markers.

**Describe the solution you'd like**
Have a standard way to include in a popup links to other markers. The link, once activated, would:

* deactivate its own marker's popup,
* center the map view on the target marker,
* and activate the target marker's popup.

**Describe alternatives you've considered**
Just referencing the other marker by an id, with no link, but that is confusing, awkward, unwieldy, and impractical for the end user, which not only would have to keep in mind an id but also have to search for it around the map.

**Additional context**
That would make it possible for users to easily navigate between relevant markers.

However, activating markers by links might not be a feature that *leaflet.js* itself supports, I haven't checked yet.

yschopfer19:
Hello @augusto-herrmann, 

I wanted  to let you know that I implemented the possibility to link markers by introducing a new feature in the marker's popup. Now, users can include links to other markers in a popup, which, when activated, closes the current popup, centers the map view on the target marker, and opens the target marker's popup.
You can see the changes I made in the #1863. 
I hope this implementation meets your requirements. I would appreciate any feedback you may have.

Best Regards,
@yschopfer19

augusto-herrmann:
Hi @yschopfer19,

thanks for the implementation! It is appreciated. I have yet to find some time to try it out in practice, but from your description and code it already does seem pretty good and useful.

It could be even better if we could define our own custom markup. That is, by just using an identifier in the link within a popup's markup (e.g. `<a href="#marker-2">go to marker popup 2</a>`, it would then create the navigation and activation as you've described. That way authors would have more flexibility when creating the popup content.

PR #1863, opened 2024-01-23
Possibility for markers to link to other markers #1563

yschopfer19:
This implementation introduces a feature allowing markers to be linked, specifically through their popups. This functionality also ensures an easy navigation between their popups.

Changes Made: 
 1. When creating a marker with a popup, users can now specify a linked marker using "linked_marker" and assign a name with "linked_marker_name".
 2. A clickable link is added to the popup. When clicked:
 - The current popup closes.
 - The linked marker gets centered, and its popup opens.
 3. If you have only specified the linked marker, it has a standard name. However, if you specified the name, the link is named after the given name.
 
**Related Issue:** #1563

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -467,7 +467,7 @@ class Popup(Element):
             **kwargs,
         )
         self.linked_marker = linked_marker
-        self.linked_marker_name =linked_marker_name
+        self.linked_marker_name = linked_marker_name
 
     def render(self, **kwargs) -> None:
         """Renders the HTML representation of the element."""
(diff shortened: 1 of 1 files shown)


Conengmo:
Thank you for your effort on this @yschopfer19! I’ve thought about it and I’ve come to the conclusion I’m not going to merge this, because it’s too much of an edge case to include it in the base marker class. We try to keep these base classes close to the functionality of Leaflet, and don’t overload them too much. I hope you can appreciate our effort to keep Folium maintainable, part of that is being picky with what new complexity we allow.

I do however see some other uses for this:

- I think your approach could be adapted to *using* the Folium library instead of being a part of it. Then, it would make a great advanced example in the documentation.
- You could host a ‘linkable marker class’ as a Folium plugin in a repo of your own, we would then link to it from our documentation. We already do this with some other repos.

If you want to write that documentation case, I’d be glad to help getting that in the docs.

Issue #1763, opened 2023-05-22
Polyline highlight color


fm-sys:
Polylines currently support showing a tooltip when clicking on them. Maybe it would be worth to add a `highlight-color` property, so the line can optionally be colored in a different style while the tooltip is shown.

Alternatively users need to add some custom JS code which might get ugly.

Conengmo:
That's a nice suggestion. If somebody is willing to pick this up, that's welcome.

Alternatively, you could convert your data to geojson format and use the `highlight_function` argument of the `GeoJson` class. We have an example of that here: https://nbviewer.org/github/python-visualization/folium/blob/main/examples/Highlight_Function.ipynb

fm-sys:
> Alternatively, you could convert your data to geojson format and use the `highlight_function` argument of the `GeoJson` class. We have an example of that here

Thanks for letting me know about the workaround, exactly what I need 👍

IndigoWizard:
@fm-sys As @Conengmo said, you can do it through the `highlight_function` of `folium.features.GeoJson` when using a GeoJSON for your polyline. I encountered the issue before and solved it through this, take a look at the solution I posted a while back:

[Folium draw and highlight polylines with distinct colours from unique GeoJSON](https://gis.stackexchange.com/questions/433810/folium-draw-and-highlight-polylines-with-distinct-colours-from-unique-geojson)

| Results - before hover | Results - on hover/highlight|
| -- | -- |
| ![](https://i.stack.imgur.com/5QpiU.png) | ![](https://i.stack.imgur.com/kuMLd.png) |

fm-sys:
I have successfully used the `GeoJson` workaround meanwhile... From my side this issue can be closed, but as I far as I understood @Conengmo we might want to track this idea for normal poly lines as well.

Conengmo:
Thanks for your input both. Indeed let's keep this issue open, I like the idea of having this on vector classes as well, as long as it can be done in a simple way.

fm-sys:
Let's close here. See https://github.com/python-visualization/folium/pull/1866#issuecomment-2200649948 for explanation

PR #1866, opened 2024-01-30
highlight_color for vector classes

yschopfer19:
The implementation introduces the possibility to highlight polylines, polygons, rectangles, circles, and circle markers if a tooltip is shown.
Changes Made:
 - Users can now specify the "highlight_color" when creating a vector layer.
 - When a user hovers over the vector layer and a tooltip is displayed, the vector layer then sets its color to the highlight color.
 
Related Issue: #1763

### DIFF
--- a/folium/vector_layers.py
+++ b/folium/vector_layers.py
@@ -133,6 +133,7 @@ class BaseMultiLocation(MacroElement):
         locations: TypeMultiLine,
         popup: Union[Popup, str, None] = None,
         tooltip: Union[Tooltip, str, None] = None,
+        **kwargs: TypePathOptions
     ):
         super().__init__()
         self.locations = validate_multi_locations(locations)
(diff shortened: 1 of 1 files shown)


Conengmo (commented):
Thanks for working in this @yschopfer19! I’m on mobile so I haven’t looked into this that detailed yet, but I wanted to share my first thought. Which is that I’d rather avoid adding the repeated code. Maybe you can think of a way to have that extra template code only in one place? 

I do like the idea of having highlight functionality on vector classes, but if the code becomes much longer or more complex I’m not sure if it’s worth it.

I hope you’re not discouraged by this. And if you want I can help with this as well later.

thanks again for your effort!

Conengmo:
Since https://github.com/python-visualization/folium/pull/1959 was merged maybe that can be used for this functionality.

hansthen:
@yschopfer19 By now we implemented #1959, which allows users to add javascript event handlers. It allows users to do what you proposed in your PR. 

If #1959 does not serve your need, please let us know.

PR #1869, opened 2024-02-04
Add a custom container to the realtime plugin

hansthen:
Add the container parameter from the leaflet `realtime` plugin to the Folium `realtime` plugin. So far, only tested with `MarkerCluster`. Also add tests and improved the documentation.


folium/plugins/realtime.py
```diff
@@ -98,11 +104,13 @@ def __init__(
         get_feature_id: Optional[JsCode] = None,
         update_feature: Optional[JsCode] = None,
         remove_missing: bool = False,
+        container: Optional[Layer] = None,
```
Conengmo, line 107: I have a sleight concern about this allowing any `Layer` object. In the Leaflet plugin documentation I read the container argument should be a (sub)class of L.LayerGroup. That's more strict than Layer.

I'm now thinking I'd rather limit this here to the exact classes that are allowed. I suppose those are not that many, so we can list them here and in the docs/docstring. And possibly add a check that raises an error on incorrect types. That way we avoid user questions about what is and isn't possible here, and save ourselves effort answering issues in the future. Looking at which classes that are, I'm thinking probably just `FeatureGroup`, `GeoJson`, `MarkerCluster`?

Conengmo (changes requested):
Nice addition to this plugin! I've added 3 comments, hope you have time to address them. Afterwards we can get this merged!

folium/plugins/realtime.py
hansthen: Good point about the `L.LayerGroup`. I changed it accordingly. 

I'd prefer not limit the container to only a limited set of classes. This would limit users of the package if they want to define their own containers (or extend an existing container). The python interface for the container is minimal. We only use `get_name` and  `add_to`. I updated the documentation to reflect the requirements on the JavaScript side.

folium/features.py
```diff
@@ -517,9 +517,9 @@ class GeoJson(Layer):
 
     >>> # Provide a style_function that color all states green but Alabama.
     >>> style_function = lambda x: {
-    ...     "fillColor": "#0000ff"
```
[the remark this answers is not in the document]
hansthen: My apologies, I think I used the wrong autoformatter here.

folium/plugins/realtime.py
Conengmo: Alright, that's fair. As long as it's well documented, we should be good.

I don't see your changes by the way, maybe you still need to push them?

hansthen: I messed up, I accidentally pushed something that did not work. There is no LayerGroup in folium.

Conengmo (approved):

Conengmo:
This looks good Hans, the docstring is pretty clear so I hope users can work with this. 

> More longterm, we can add LayerGroup as an abstract base class to indicate something that will generate a L.LayerGroup javascript object. That way we can make the Python class hierarchy mimic the Leaflet class hierarchy.

Though that does make sense, I'm hesitant from a complexity perspective. Adding a core Folium class doesn't seem to weigh up to facilitating a single type hint for an optional argument in a single plugin. Having too many layers of class inheritance makes things more complicated to understand. So I think I'm fine with how it's solved now.

I'll go ahead and merge it now. Thanks for your contribution to Folium!

PR #1874, opened 2024-02-07
add tile layer to example

danmaps:
looks like this was accidentally removed when stamen tiles were taken out all over the place


martinfleis (approved):
Thanks!

PR #1879, opened 2024-02-11
Add ability to set font size in plot

this-josh:
This PR permits you to set the leaflet font size. This is helpful to make plots more legible in certain circumstances.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -153,6 +153,9 @@ class Map(JSCSSMixin, MacroElement):
         rare environments) even if they're supported.
     zoom_control : bool, default True
         Display zoom controls on the map.
+    font_size : int or float or string (default: '1rem')
+        The font size to use for Leaflet, can either be a number of a
+        string ending in 'rem'
     **kwargs
         Additional keyword arguments are passed to Leaflets Map class:
         https://leafletjs.com/reference.html#map
@@ -236,7 +239,6 @@ class Map(JSCSSMixin, MacroElement):
         height: Union[str, float] = "100%",
         left: Union[str, float] = "0%",
         top: Union[str, float] = "0%",
-        font_size: str = "1rem",
         position: str = "relative",
         tiles: Union[str, TileLayer, None] = "OpenStreetMap",
         attr: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


folium/folium.py
```diff
@@ -235,6 +236,7 @@ def __init__(
         height: Union[str, float] = "100%",
         left: Union[str, float] = "0%",
         top: Union[str, float] = "0%",
+        font_size: str = "1rem",
```
martinfleis, line 239: This needs to be on the bottom as the last argument so it does not break potential use of positional args.

It also needs to be added to the docstring above.

martinfleis (commented):
Nice change! Minor comment in the code.

folium/folium.py
this-josh: Good catch, I've implemented these changes

martinfleis (approved):

Conengmo (commented):
Thanks for your effort @this-josh! It’s a nice improvement and I’m sure we can merge it soon.

 I found a typo, and I added a question, hope to hear your thinking on it.

folium/utilities.py
```diff
@@ -425,3 +425,13 @@ class JsCode:
...
     def __init__(self, js_code: str):
         self.js_code = js_code
+
+
+def parse_font_size(value: Union[str, int, float]) -> str:
+    """Parse a font size value."""
+    if isinstance(value, (int, float)):
+        return f"{value}rem"
+
+    if "rem" not in value:
```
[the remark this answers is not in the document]
this-josh: Good point, I've updated it to accept `em` and `px`.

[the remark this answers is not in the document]
Conengmo: Thanks! The default could still be rem if you like.
I noticed some test may fail, once that’s solved I’ll go ahead and merge it!

tests/test_utilities.py
```diff
@@ -218,7 +219,7 @@ def test_escape_double_quotes(text, result):
 def test_javascript_identifier_path_to_array_notation(text, result):
     assert javascript_identifier_path_to_array_notation(text) == result
 
-
+
```
Conengmo, line 222:
```suggestion

```

tests/test_utilities.py
```diff
@@ -218,7 +219,6 @@ def test_escape_double_quotes(text, result):
 def test_javascript_identifier_path_to_array_notation(text, result):
     assert javascript_identifier_path_to_array_notation(text) == result
 
-
 def test_js_code_init_str():
```
Conengmo, line 222:
```suggestion

def test_js_code_init_str():
```

Conengmo:
Thanks @this-josh!

PR #1880, opened 2024-02-14
Added how to update pip version in  README.rst

Nan227:
I did follow your steps, but my VS code requires the latest version of pip. It may help users understand the code more.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,10 @@ or
 
     $ conda install -c conda-forge folium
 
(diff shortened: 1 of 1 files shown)


Conengmo:
Hi Nan, thanks for your suggestion! I thought about it, and I’d rather keep our readme as light weight as possible. This is a more generic Python thing, so I don’t think we have to include it here.

Thanks again!

PR #1881, opened 2024-02-17
fixed typos of examples in code

S-Panta:
This PR fixes the typos of the example of tiles. The code cannot find  `CartoBD Voyager` in the `xyzservices.TileProvider`. It should be `CartoDB Voyager`. 
Reference: https://github.com/geopandas/xyzservices/blob/8c3f23830e9d0dd06406d3371610351aae77ed76/xyzservices/data/providers.json#L2075


martinfleis (approved):
Thanks!

PR #1882, opened 2024-02-17
changed NASAGIBS Blue Marble to NASAGIBS BlueMarble3413 in examples

S-Panta:
`NASAGIBS Blue Marble` was removed from xyzservices provider list as the endpoint doesn't work. (https://github.com/geopandas/xyzservices/pull/158) I suggest the example in Folium to be replaced with either `"BlueMarble3413` or `"BlueMarble3031`. 
 However, at the moment, the URL that's in their json file of these two tiles isn't working and throws 400 error.  
 Or, maybe this example can be thrown out as it would be very confusing.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -89,7 +89,7 @@ class Map(JSCSSMixin, MacroElement):
         - "OpenStreetMap"
         - "CartoDB Positron"
         - "CartoBD Voyager"
-        - "NASAGIBS Blue Marble"
+        - "NASAGIBS BlueMarble3413"
 
     You can pass a custom tileset to Folium by passing a
     :class:`xyzservices.TileProvider` or a Leaflet-style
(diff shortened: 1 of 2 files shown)


martinfleis:
Those will not work as they are in different CRS (EPSG 3413 and 3031 respectively). Folium supports only tiles in 3857. If you are able to figure out how to make the original BlueMarble work, happy to add it back to xyzservices but as long as it returns error 500 it'll stay out.

prusswan:
@martinfleis https://gibs-c.earthdata.nasa.gov/wmts/epsg3857/best/wmts.cgi?Service=WMTS&Request=GetCapabilities (WMTS endpoint used by https://worldview.earthdata.nasa.gov) has a list of active WMTS services. The closest one for Blue Marble is probably this:

`https://gitc.earthdata.nasa.gov/wmts/epsg3857/best/BlueMarble_NextGeneration/default/{TileMatrixSet}/{TileMatrix}/{TileRow}/{TileCol}.jpeg`

Conengmo:
We 'solved' this for Folium in https://github.com/python-visualization/folium/pull/1918.

Issue #1865, opened 2024-01-27
Map parameter to set position of zoom control buttons


berrfred:
I've been searching the past issues for such a feature but to my surprise I could not find anything.

All other plugins I've tried (Fullscreen, Geocoder, LocateControl, MousePosition) offer such a position parameter (topleft, topright, bottomleft, bottomright).

Why should we not be able to set it directly in folium without being forced to move a div in the html ?
Thanks

![image](https://github.com/python-visualization/folium/assets/50397929/6d52743f-0a9a-48fe-9b89-8443a7c64aee)

berrfred:
This is how I achieve it currently, it does work but it's not very cool ...

```python
map_script = '<script>document.addEventListener("DOMContentLoaded", function() {\
	document.querySelector("div.leaflet-bottom.leaflet-right").insertBefore(\
		document.querySelector("div.leaflet-control-zoom"), document.querySelector("div.leaflet-control-attribution")\
	);\
});</script>'

map.get_root().html.add_child(folium.Element(map_script))
```

berrfred:
The same request could be made to position a colormap legend ... currently solved with the same approach (moved to topleft in this case).

```python
map_script_legend_control = '<script>document.addEventListener("DOMContentLoaded", function() {\
	document.querySelector("div.leaflet-top.leaflet-left").prepend(\
		document.querySelector("div.legend.leaflet-control")\
	);\
});</script>'

fmap.get_root().html.add_child(folium.Element(map_script_legend_control))
```

Conengmo:
Hi Frederic, that’s an interesting question. I agree it would be nice to be able to set the position of the zoom control. Currently, they are controlled by setting a Boolean argument on the Map class.

 I’d rather not add any more arguments to the Map class. What would be another way to allow defining the position? I’m thinking now we could make the zoom_control argument not only allow a boolean, but also a string containing a position.

what do you think?

berrfred:
Your proposal makes sense to me, this way you keep backwards compatibility for True/False but also adds the usual four position strings that are an implicit True with additional positon information.

Conengmo:
Good to hear, would you like to make a PR?

berrfred:
I'd like to do that but it is not completely clear to me (yet) how to achieve that ... currently I see the zoom control creation is delegated to L.map() so does that mean I should pass False to the map leaflet and create myself the L.control.zoom()  with its position in the template part ?

berrfred:
[...]

PR #1884, opened 2024-02-24
Update Map zoom_control variable to also allow string to set position

berrfred:
This is referred to issue [#1865](https://github.com/python-visualization/folium/issues/1865)


berrfred:
This is referred to issue [#1865](https://github.com/python-visualization/folium/issues/1865)

folium/folium.py
```diff
@@ -150,8 +150,8 @@ class Map(JSCSSMixin, MacroElement):
         Forces Leaflet to not use hardware-accelerated CSS 3D
         transforms for positioning (which may cause glitches in some
         rare environments) even if they're supported.
-    zoom_control : bool, default True
-        Display zoom controls on the map.
+    zoom_control : bool or position string, default True
+        Display zoom controls on the map, eventually specifying its position.
```
Conengmo, line 154:
```suggestion
        Display zoom controls on the map. The default `True` places it in the top left corner. 
        Other options are 'topleft', 'topright', 'bottomleft' or 'bottomright'.
```

folium/folium.py
```diff
@@ -284,10 +288,17 @@ def __init__(
         self.crs = crs
         self.control_scale = control_scale
 
+        # Zoom control position specified ?
+        if isinstance(zoom_control, str):
+            self.zoom_control_position = True
+            self.zoom_control = zoom_control
```
Conengmo, line 294: maybe useful to add a check here for valid values? Now if a users makes an error, they just get a grey screen. For example something like:

```suggestion
            self.zoom_control = zoom_control
            if zoom_control not in {'topleft', 'topright', 'bottomleft', 'bottomright'}:
                raise ValueError("Incorrect value for `zoom_control`, choose from 'topleft', 'topright', 'bottomleft' or 'bottomright'.")
```

Conengmo (changes requested):
Implementation is good! I added two comments about the four possible string values. I'd like to make sure users won't make mistakes in these. I hope you can address these comments, afterwards this is good to go!

Conengmo:
Somethings going on with the tests, it's probably not related to this PR.

folium/folium.py
berrfred: I had considered introducing a check on valid position values but looking at various plugins I could not find such a check so I thought it was considered as unnecessary ...

Conengmo: I understand, this code base grew quite organically over time, so it's not always consistent. Thanks for putting it in!

Conengmo (approved):

Conengmo:
Thanks for your contribution @berrfred!

PR #1896, opened 2024-03-09
Fix typo in choropleth.md

andyforrest:


Conengmo:
Thanks Andy! Good catch

Issue #1519, opened 2021-10-16
The js/css override mechanism is not well explained in the docs


JeffAbrahamson:
**Describe the bug**
It is unclear from the docs (and from the code) how to override javascript and css that folium emits.  There's some very old discussion in (closed) issue #192 and (open) issues #1479 and #1482 on this subject.  The issue #1479 is particularly striking: it's nice to update folium to use bootstrap 4, but is the only way to do this to change folium?  What if I use folium on two sites and the sites use different versions of bootstrap for some historical reason?  This seems like it should be customisable, and #192 suggests it is.  And yet code such as

    figure.header.add_child(
        JavascriptLink('https://example.com/bootstrap/400.5.0/css/bootstrap.min.css'),
        name="bootstrap_css")

doesn't have the desired effect of changing the output html.  (Bootstrap 400 doesn't exist, of course, I want to make clear that I'm trying to change the default, even if the default is quite different when this issue is read by someone many years hence, as happens.)

It appears from #1479 that a different version of folium would solve my underlying issue (of wanting to use bootstrap 4), but it wouldn't address the issue that the docs don't tell me how to modify what version it uses.  And #1482 anyway hasn't been merged.

**To Reproduce**

As I'm trying to make this go in a django context, I've written a minimal reproducible example in django.  The (hopefully) only relevant bit is the code in this [view function](https://github.com/transport-nantes/map-mre/blob/correct-html-but-bootstrap-conflict/geo_mre/map/views.py).  That project can be cloned and run and it will provide two pages: an index page and a map page.  The map page is slightly borked because of the bootstrap mismatch.

I've asked this question on the [leaflet google group](https://groups.google.com/g/leaflet-js/c/8Bps-rr8iYM) with a fair bit of (by now slightly rambling) documentation of things I've tried to do and code I've looked at to try to understand it.  I've since realised this is probably more appropriate for folium than for leaflet, but that was less clear when I started.

**Expected behavior**
This is a documenation bug, not a code functionality bug, so the expected behaviour is that I'd be able to learn how to make folium use my bootstrap version by reading the docs.

**Environment (please complete the following information):**
 - Browser = firefox
 - django
 - Python version = 3.9.5 : `sys.version_info(major=3, minor=9, micro=5, releaselevel='final', serial=0)`
 - folium version = 0.12.1
 - branca version = 0.4.2

**Additional context**

**Possible solutions**
List any solutions you may have come up with.

[...]

PR #1898, opened 2024-03-09
Re: #1519 Override css and js links

hansthen:
Add a method to the JSCSSMixin that allows users to override a specific link to css or js. Unfortunately this still requires users to know the name of the of the link, which will require looking at the code or introspection.

### DIFF
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple
+from typing import List, Literal, Tuple
 
 from branca.element import CssLink, Element, Figure, JavascriptLink, MacroElement
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


Conengmo (changes requested):
Hi Hans, thanks for helping out with this one! Your code looks good, but I want to challenge using another approach here.

Now there's one function with a switch parameter, that has some logic in it to switch based on that parameter. In my opinion this is a pattern that indicates two separate functions would be better suited. Here's an example to illustrate what I mean:

```
def add_js_link(name, url):
     self._add_link(name, url, self.default_js)

def add_css_link(name, url):
     self._add_link(name, url, self.default_css)

def _add_link(name, url, link_array):
    # add the link to the array
```

This way we don't need switching logic. And I think it's a bit more user friendly than having a switching parameter the user must learn how to use.

I'd also opt to not have a separate method to add multiple links. This is easy for users to do themselves, and less code for us to maintain.

To really close the ticket we also need to document this feature in our documentation. I'm thinking maybe a new page in the advanced guide?

hansthen:
I updated the code as requested. As to the documentation, adding links for `javascript` and `css` is mostly useful for writing plugins.  Regular use will not typically require this API. What would you think if we added a guide for plugin writers to the advanced documentation and add the documentation there?

Conengmo:
> I updated the code as requested.

Thanks!

> As to the documentation, adding links for javascript and css is mostly useful for writing plugins. Regular use will not typically require this API.

I don't think that's true. We've had users ask for this feature so they can make self contained maps that use local resources only. Also, https://github.com/python-visualization/folium/issues/1519 talks about his as a user from a Django context.

For developers we have our contribution guide, which has some guidelines on plugin development. I don't really see the link with this topic here.

I'm also fine with merging this and leaving the ticket open to add documentation in a later stage. What do you prefer?

hansthen:
Yeah I see the other use cases now. I think I prefer to add the documentation in this PR.

Conengmo (approved):
Looks good! Thanks for adding that documentation!

PR #1895, opened 2024-03-08
Add layer control tree plugin

hansthen:
Implemented the `Leaflet.Control.Layers.Tree` in Folium. See https://github.com/jjimenezshaw/Leaflet.Control.Layers.Tree for more details.

Closes: #1606

PR #1912, opened 2024-04-03
Add custom Template class with tojavascript filter

Conengmo:
Based on https://github.com/python-visualization/folium/pull/1856, split off from the more complete version in https://github.com/python-visualization/folium/pull/1895.

Add a custom Template class that contains the `tojavascript` template filter that converts Python objects to Javascript notation. This may include Javascript object references or functions.

Instead of applying this template everywhere at once, we'll adopt is gradually where it is useful. This is mainly when one of the arguments is a Javascript function or when we want to reference another Folium object in the Javascript template.

### DIFF
--- a/folium/template.py
+++ b/folium/template.py
@@ -1,0 +1,51 @@
+import json
+from typing import Union
+
+import jinja2
+from branca.element import Element
+
+from folium.utilities import JsCode, TypeJsonValue, camelize
+
+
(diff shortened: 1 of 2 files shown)


Conengmo:
I'll go ahead and merge this one since it's basically @hansthen code of which I wrote the first draft, and has already been reviewed by me. The tests are new, but not that interesting :)

Conengmo:
@hansthen want to do a final review of this one?

Issue #1908, opened 2024-03-26
Folium loads two versions of bootstrap


bopjesvla:
**Describe the bug**

Folium currently loads two versions of bootstrap:

```html
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/css/bootstrap.min.css"/>
<link rel="stylesheet" href="https://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css"/>
```

**To Reproduce**

The first Folium iframe on https://python-visualization.github.io/folium/latest/getting_started.html contains the above snippet.

Conengmo:
That's correct, we require Bootstrap 3 to show Glyphicons. See https://github.com/python-visualization/folium/pull/1650.

But I did just notice in https://stackoverflow.com/a/18225474/4082914 they also list a css just for Glyphicons. It seems to work as well. Not sure why we didn't include that in https://github.com/python-visualization/folium/pull/1650. I'll open a PR.

PR #1914, opened 2024-04-03
Drop Bootstrap 3 but keep glyphicons

Conengmo:
In https://github.com/python-visualization/folium/pull/1650 we transitioned from Bootstrap 3 to 5. We also had to keep Bootstrap 3 for the Glyphicon icons in the AwesomeMarkers.

Turns out there's also a hosted css file with only the css necessary for the glyphicons, not the entirety of Bootstrap 3. See https://stackoverflow.com/a/18225474/4082914 under "If you want to use the Glyphicons separately". Use that file instead.
 
Closes https://github.com/python-visualization/folium/issues/1908.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -45,7 +45,7 @@ _default_css = [
     # glyphicons came from Bootstrap 3 and are used for Awesome Markers
     (
         "glyphicons_css",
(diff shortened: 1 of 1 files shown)


Conengmo:
Thanks Hans for the review!

PR #1917, opened 2024-04-04
Remove Python 3.7, add 3.12

Conengmo:
Python 3.7 is end-of-life, so remove it as a supported version. Add tests for 3.12 and add it as a supported version (if tests succeed).

I added a matrix to the Selenium tests as well, since they are important for catching regressions.


ocefpaf (approved):

Conengmo:
Seems all good. Thanks for the review @ocefpaf!

PR #1918, opened 2024-04-04
Remove NASAGIBS Blue Marble example

Conengmo:
https://github.com/python-visualization/folium/pull/1882 was very helpful in showing that the currently listed example tileset doesn't work anymore. But the proposed replacement is not suitable unfortunately. Simply solve this issue by removing the example. If users want more examples, they can visit the site we mention: http://leaflet-extras.github.io/leaflet-providers/preview/

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -90,14 +90,13 @@ class Map(JSCSSMixin, MacroElement):
         - "OpenStreetMap"
         - "CartoDB Positron"
         - "CartoDB Voyager"
-        - "NASAGIBS Blue Marble"
 
-    You can pass a custom tileset to Folium by passing a
+    You can also pass a custom tileset to Folium by passing a
     :class:`xyzservices.TileProvider` or a Leaflet-style
     URL to the tiles parameter: ``http://{s}.yourtiles.com/{z}/{x}/{y}.png``.
 
     You can find a list of free tile providers here:
-    ``http://leaflet-extras.github.io/leaflet-providers/preview/``.
+    http://leaflet-extras.github.io/leaflet-providers/preview/.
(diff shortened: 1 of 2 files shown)


folium/folium.py
```diff
@@ -90,14 +90,13 @@ class Map(JSCSSMixin, MacroElement):
...
-        - "NASAGIBS Blue Marble"
 
-    You can pass a custom tileset to Folium by passing a
+    You can also pass a custom tileset to Folium by passing a
     :class:`xyzservices.TileProvider` or a Leaflet-style
     URL to the tiles parameter: ``http://{s}.yourtiles.com/{z}/{x}/{y}.png``.
 
     You can find a list of free tile providers here:
-    ``http://leaflet-extras.github.io/leaflet-providers/preview/``.
+    http://leaflet-extras.github.io/leaflet-providers/preview/.
```
martinfleis, line 99: Shall we maybe point to `xyzservices` instead? Either to the docs (the list is [here](https://xyzservices.readthedocs.io/en/stable/introduction.html)) or with the code snippet 

```py
import xyzservices.providers as xyz

xyz
```

Now that we depend on it, it will be more straightforward to just pass the name of the tiles rather than figuring out on leaflet-providers which URL shall be passed where and how the attribution should look like.

Conengmo: @martinfleis I streamlined the docstring a bit more, does this look good to you? I hope it's more clear now that: users should mainly provide provider names, and that the xyzservices package is used underneath. Underneath, if they want to do more advanced stuff we still list a link to `xyzservices.TileProvider`, as well as how to use a custom url.

martinfleis (approved):
Looks good!

Conengmo:
Thanks Martin!

PR #1922, opened 2024-04-04
Use better dev version number in docs

Conengmo:
Locally this works great: use `git describe --tags` to return either the version number of the release (which is the current tag on release), or the latest version plus how for the commit is away from the version (e.g. `v0.16.0-24-g007fa8c7`). This should look much better than the `0.1.dev1.dev+gca4ce8f` we have now. I hope this works on Github Actions as well.

According to PEP440, the dev version should be like `v0.16.0.dev1`, so I added a replace to make it so.


martinfleis:
Isn't there a way to configure setuptools_scm to do this for us, instead of replacing the string ourselves?

Conengmo:
Good suggestion Martin, I found this snippet in their docs which seems to work well: https://setuptools-scm.readthedocs.io/en/stable/usage/#usage-from-sphinx.

```
Running Sphinx v7.2.6
Version: 0.16.1.dev37+g1296fdf3
```

It assumes we're working towards a 0.16.1 release instead of a 0.17.0 release, but I guess that doesn't really matter. Most important thing is that it's much better than 0.1.dev that's shown now :)

Conengmo:
Okay so turns out the issue was not in the conf.py code, but that there was no git history to find the latest tag in. So just updating the workflow to also fetch the history is enough to get the right version number. There's also no need to manipulate that string, it's already good as it is.

PR #1923, opened 2024-04-04
Fix small issues with Timeline and TreeLayerControl docs

Conengmo:
We added two new plugin classes today, but there are some small issues with the docs:

- they are missing in the plugins overview page
- the TreeLayerControl example is not showing up because the map is not returned at the end
- Sphinx is complaining about the docstrings of `Timeline` and `TimelineSlider` because the indentation is inconsistent.


hansthen:
@Conengmo Thanks for fixing. I also discovered how I can review the generated documentation, so I hope it will not happen again in the future.

PR #1924, opened 2024-04-04
docs: mark logo as dark mode safe

Conengmo:
Tiny tweak to fix something that bothered me: the transparent png logo in our docs has a white background when using dark mode. Turns out the theme we use does this on purpose and expects us to mark images that are safe for dark mode: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/light-dark.html#images-and-content-that-work-in-both-themes. Do that here by adding the 'dark-light' class.

** Before **

afbeelding

** After **

afbeelding

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,6 +5,7 @@ Folium
 
 .. image:: _static/folium_logo.png
    :height: 100px
+   :class: dark-light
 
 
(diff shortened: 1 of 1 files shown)


hansthen:
Looks good to me

PR #1928, opened 2024-04-10
Add `Leaflet.encoded` plugin: Enable creating PolyLine and Polygon from encoded string

### LINKED ISSUE #1886: Create a `PolyLine` object directly from the polyline string via the Leaflet.encoded plugin
Creating a `PolyLine` vector from an encoded string is currently a bit tedious as one needs to decode it prior to passing it to the constructor. 

[Leaflet.encoded](https://github.com/jieter/Leaflet.encoded) plugin provides an alternative way to directly use Leaflet with encoded strings and perhaps `folium` will benefit from having this plugin to squeeze out performance for users who are working with encoded strings rather than latitude-longitude pairs.

Maintain the status-quo.

**Additional context**
`Leaflet.encoded` doesn't seem to be under active maintenance (last commit was 4 years ago). If there's a policy against such plugins (is there a policy regarding qualifications of plugins?) then this might be problematic.

**Implementation**
Absolutely, I have limited experience with Javascript but if the maintainers reckon this to be doable without an extensive knowledge in js or leaflet, I'll be happy to contribute.

achieveordie:
Fixes #1886

- Create a plugin class called `PolyLineFromEncoded`, which accepts an encoded string to make a polyline. Also, add all the relevant docs/tests required for plugin approval.
- Create a plugin class called `PolygonFromEncoded`, which works the same way as above, but for polygons.

FYI @hansthen


hansthen (commented):
That is very clean and looks complete. 

I made one small change (added a type annotation). Since this is my first PR review, I will ask for a second check from @Conengmo if he has time. But to me it looks okay to merge.

achieveordie:
In hindsight, I'm also considering adding the second functionality the original author provides - `Polygon.fromEncoded()`. They will be in the same plugin file (renamed from `polyline_encoded` to `encoded`) but have separate test/doc files.

hansthen:
> In hindsight, I'm also considering adding the second functionality the original author provides - `Polygon.fromEncoded()`. They will be in the same plugin file (renamed from `polyline_encoded` to `encoded`) but have separate test/doc files.

Do you want to include this in this PR? If so, go ahead, but you can also do it in a later PR.

achieveordie:
done @hansthen :)

hansthen (approved):
Looks good to me.

achieveordie:
How long before this can be merged? I'd like to use the 'official' feature :)

folium/plugins/encoded.py
```diff
@@ -0,0 +1,120 @@
...
+        (
+            "polyline-encoded",
+            "https://cdn.jsdelivr.net/npm/polyline-encoded@0.0.9/Polyline.encoded.js",
+        )
+    ]
+
+    def __init__(self, encoded: str, **kwargs):
+        super().__init__()
+        self.encoded = encoded
+        self.options = parse_options(**kwargs)
```
Conengmo, line 48: Instead of `parse_options`, we have a function specifically for vector classes that parses options: `path_options`. 

For PolyLine it's called like this: `path_options(line=True, **kwargs)`

For Polygon: `path_options(line=True, radius=None, **kwargs)`

Maybe good to reuse that in the child classes? It contains some sensible defaults and does some checks. Or did you not include it on purpose?

docs/user_guide/plugins/polygon_encoded.md
```diff
@@ -0,0 +1,19 @@
+```{code-cell} ipython3
+---
+nbsphinx: hidden
+---
+import folium
+from folium import plugins
+```
+
+# PolygonFromEncoded
+
```
Conengmo, line 10: I'd expect this documentation to have one or two sentences that explain what this class is. Like the headline of the docstring "Create PolyLines directly from the encoded string" or something like that.

And maybe more importantly, something about, or a link to, how that encoding works.

Conengmo (commented):
Looks good, well done! I added two small comments, hope you can take a look. Shouldn't take long to get this merged!

docs/user_guide/plugins/polygon_encoded.md
achieveordie: done

folium/plugins/encoded.py
```diff
@@ -0,0 +1,123 @@
...
+    >>> encoded = r"w`j~FpxivO}jz@qnnCd}~Bsa{@~f`C`lkH"
+    >>> PolygonFromEncoded(encoded=encoded).add_to(m)
+    """
+
+    def __init__(self, encoded: str, **kwargs):
+        self._name = "PolygonFromEncoded"
+        super().__init__(
+            encoded,
+        )
+        path_options(line=True, radius=None, **kwargs)
```
Conengmo, line 118: Thanks for the quick response! This is not working yet, I suppose you mean to pass these to the super init? Also in the other class.

achieveordie:
@Conengmo I appreciate your input on the error. It was an oversight on my end. This PR should be good to go.

PR #1933, opened 2024-04-18
Update VectorGridProtobuf to include popups when clicked

christianaaronschroeder:
I've tried using `VectorGridProtobuf` several times in the past and almost always found myself making a copy of this class and adding functionality to click a polygon and get the properties.

Open to suggestions, or other options, just putting this out there.

### DIFF
--- a/folium/plugins/vectorgrid_protobuf.py
+++ b/folium/plugins/vectorgrid_protobuf.py
@@ -113,6 +113,20 @@ class VectorGridProtobuf(JSCSSMixin, Layer):
                     {{ this.options if this.options is string else this.options|tojson }}
                 {%- endif %}
             );
+            // Attach popup event
+            {{ this.get_name() }}.on('click', function (e) {
+                const properties = e.layer.properties;
+                var popupContent = "";
+                for (var property in properties) {
+                    if (properties.hasOwnProperty(property)) {
+                        popupContent += "<b>" + property + ":</b> " + properties[property] + "<br>";
+                    }
+                }
+                L.popup()
+                    .setContent(popupContent)
+                    .setLatLng(e.latlng)
+                    .openOn(map);
+            });
             {%- endmacro %}
             """
     )


hansthen:
@christianaaronschroeder Thanks for your contribution. This would be useful to add. I do have some questions and suggestions mostly having to do with configurability from the Python side. 

Before going any further, there is an open Pull Request that does something similar. Could you have a look https://github.com/python-visualization/folium/pull/1903?

It allows you to set event handlers (`on`) from the Python side. Would this also solve your needs?

christianaaronschroeder:
> @christianaaronschroeder Thanks for your contribution. This would be useful to add. I do have some questions and suggestions mostly having to do with configurability from the Python side.
> 
> Before going any further, there is an open Pull Request that does something similar. Could you have a look #1903?
> 
> It allows you to set event handlers (`on`) from the Python side. Would this also solve your needs?

I believe so. I'm open to trying that before moving forward with this one.

hansthen:
> > @christianaaronschroeder Thanks for your contribution. This would be useful to add. I do have some questions and suggestions mostly having to do with configurability from the Python side.
> > Before going any further, there is an open Pull Request that does something similar. Could you have a look #1903?
> > It allows you to set event handlers (`on`) from the Python side. Would this also solve your needs?
> 
> I believe so! I'm open to trying that before moving forward with this one.

Great!

I will try to get that PR approved.

Conengmo (changes requested):
Thanks for your PR Christian, great that you want to help improve Folium!

I understand this change is very useful for your situation, but I worry a bit about it applying to general users, since the use of a 'property' property as popup content is quite specific. I don't think that's a good fit, ideally interfaces are explicit (like the class parameters we often use) or are well documented. We also try to avoid custom code in plugins to avoid too much maintenance burden.

If we want to apply popups on this plugin, ideally it works with the `Popup` class Folium already has. That way we don't need custom code in this plugin.

Or take a look at the GeoJsonPopup class, which seems a bit similar to this usecase of showing a popup per feature. Maybe something like that could generalize to any layer-like object?

Alternatively, you could host this change as a custom plugin yourself, and we can link to it in our list of external Folium plugins.

hansthen:
@christianaaronschroeder By now PR https://github.com/python-visualization/folium/pull/1959 has been merged and I think also released into the new folium version. This feature allows you to add event handlers to arbitrary Layers. Can you have a look if this satisfies your need? If so, we can close this ticket.

christianaaronschroeder:
@hansthen looks good, thank you!

Issue #1932, opened 2024-04-16
GeoJson style is not being applied correctly


ocefpaf:
**Describe the bug**

Both in a loop or using a GeoDataframe, only the last style is applied to all geometries.
We can see that in our docs:

![Screenshot from 2024-04-16 21-23-04](https://github.com/python-visualization/folium/assets/950575/51c9334b-95e1-48dc-bf95-4225c6a37b49)

**To Reproduce**

https://python-visualization.github.io/folium/latest/user_guide/geojson/geopandas_and_geo_interface.html

I also believe https://github.com/python-visualization/folium/issues/1608 is related.

**Expected behavior**
We expect each geometry to have its own color defined by the style column.

**Environment (please complete the following information):**
 - Browser: all tested
 - Jupyter Notebook or html files? Both.
 - Python version: 3.11
 - folium version: 0.16.0
 - branca version: 0.7.1

**Additional context**
In a loop one can "hack" this problem by assigning a unique value using `exec` (note recommended though).

```python
import folium

m = folium.Map()
for k, glider in gdf.iterrows():
    exec(f"style_function = lambda x: {glider['style']}")  # hack it to make it work
    # style_function = lambda x: glider['style']  # shows the bug
    folium.GeoJson(
        data=glider["geometry"],
        style_function=style_function,
    ).add_to(m)
m
```
**Possible solutions**

I'll investigate more and see if I can come up with something.
I did try with latest `branca` and `folium` and this is not solved yet.

ocefpaf:
OK. We are hitting https://docs.python-guide.org/writing/gotchas/#late-binding-closures :-/
I'll see what we can do. At least we should amend the docs.

Conengmo:
Thanks for addressing this issue @ocefpaf!

PR #1935, opened 2024-04-19
fix __geo_interface__ styling

ocefpaf:
Closes https://github.com/python-visualization/folium/issues/1932.

- [X] Fixes the geopandas (or any `__geo_interface__` object) styling regression.
- [x] Amend the docs with a CAVEAT regarding Late Binding Closures when using `style_function` in a loop.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@ exclude: examples/data/|.*\.css|.*\.json|.*\.geojson|.*\.html
 
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.5.0
+  rev: v4.6.0
   hooks:
     - id: trailing-whitespace
     - id: check-ast
@@ -15,12 +15,12 @@ repos:
       files: requirements-dev.txt
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.3.4
+  rev: v0.4.1
   hooks:
     - id: ruff
 
 - repo: https://github.com/psf/black
-  rev: 24.3.0
+  rev: 24.4.0
   hooks:
   - id: black
     language_version: python3
--- a/docs/user_guide/geojson/geojson.md
+++ b/docs/user_guide/geojson/geojson.md
@@ -237,6 +237,23 @@ colormap.add_to(m)
 m
 ```
 
+>
+> **Caveat**
+>
+> When using `style_function` in a loop you may encounter Python's 'Late Binding Closure' gotcha!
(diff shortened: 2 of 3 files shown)


ocefpaf:
This one fixes the geopandas issue but I need to amend the docs for the pitfalls in the loop case and I also need to check if my simplistic solution is good enough and doesn't clash with the current styling options we have.

Screenshot from 2024-04-19 19-28-18

docs/user_guide/geojson/geojson.md
```diff
@@ -237,6 +237,23 @@ colormap.add_to(m)
 m
 ```
 
+>
+> **Caveat**
```
ocefpaf, line 241: @Conengmo do you have any suggestion to make this more prominent, like a red warning sign?

Also, I'd like to link the `geopandas_and_geo_interface.html#Adding-style` to the GeoPandas object suggestion below but I have no idea how to do that here.

folium/features.py
```diff
@@ -643,6 +643,10 @@ class GeoJson(Layer):
                 .done({{ this.get_name() }}_add);
         {%- endif %}
 
+        {%- if not this.style %}
```
ocefpaf, line 646: I checked all the docs/notebooks we have and none broke. If the user specifies a `style_function`, whatever is in the geojson will be ignored but, if nothing is specified and the geojson has style info, that will be used instead. I guess this is the best I could come up with but I'm open to suggestions.

docs/user_guide/geojson/geojson.md
hansthen: I think it is clear enough with the specified Caveat. It is a very tricky bug, but I also imagine it is not a very common software pattern to create style functions in a loop.

folium/features.py
hansthen: Is this change required to fix this bug? It seems like a new feature. If so perhaps it is better to separate this into a new PR.

Also, what happens if a Feature in the GeoJson does not have a `style` property? Will leaflet handle this correctly?

.pre-commit-config.yaml
hansthen, line 1: Are these changes required for the issue at hand? If not perhaps it is cleaner to open a separate PR.

hansthen (approved):

.pre-commit-config.yaml
ocefpaf: They are required to fix the styling for the geopandas case. This fixes a regression introduced during the GeoJson refactoring. I prefer to keep them both in the same PR b/c they are tied together by the same issue, even though they have different origins.

ocefpaf:
@Conengmo and @hansthen I'm planning on a new release but I guess we should also try to get https://github.com/python-visualization/folium/pull/1903 (and maybe others) in before that?

PR #1936, opened 2024-04-22
Minor docs fixes

ocefpaf:
This fixes some deprecation warnings that were appearing in our rendered HTML docs page and restores the WMS example that is working fine at the moment. (It wasn't in the past probably due to some downtime or network error.)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -47,8 +47,11 @@ source_suffix = ".rst"
 master_doc = "index"
 
 # General information about the project.
+import datetime
+
+year = datetime.datetime.now(tz=datetime.timezone.utc).date().year
 project = "Folium"
-copyright = "2013, Rob Story"
+copyright = f"{year}, Rob Story"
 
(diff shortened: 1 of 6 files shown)


docs/conf.py
```diff
@@ -47,8 +47,11 @@
 master_doc = "index"
 
 # General information about the project.
+import datetime
+
+year = datetime.datetime.now(tz=datetime.timezone.utc).date().year
 project = "Folium"
-copyright = "2013, Rob Story"
+copyright = f"{year}, Rob Story"
```
hansthen, line 54: A copyright statement with a single year indicates the year a document was first published. Better is "2013-{year}".

hansthen (approved):
I have one minor issue, but otherwise it looks good to me.

docs/conf.py
ocefpaf: Yep. Let's do it.

martinfleis (approved):

PR #1938, opened 2024-04-25
CI: include test against ngihtly numpy

### LINKED ISSUE #1937: NumPy 2.0 support

NumPy 2.0 is coming out soon ( https://github.com/numpy/numpy/issues/24300 ). NumPy 2.0.0rc1 packages for conda & wheels came out 3 weeks back ( https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395 ). It would be good to confirm folium works with NumPy 2.0

To prepare for NumPy 2.0, it might be worthwhile to start testing folium against NumPy 2 in CI

Also as NumPy is tracking ecosystem support for NumPy 2.0, it would be helpful to share folium current support status in issue (with any plans): https://github.com/numpy/numpy/issues/26191

NumPy has put out [a migration guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide ). More details are in [the release notes]( https://numpy.org/devdocs/release/2.0.0-notes.html ). As folium doesn't have C/C++ usage of NumPy, only the Python changes would be relevant

In the short term, folium may wish to set an upper bound on NumPy version until testing can happen. Alternatively folium could wait for user issues and address as those come up

Maybe there are other options that haven't been considered here

**Additional context**

NA

**Implementation**

This likely involves some tweaks to CI (or maybe a new CI job)

Unclear whether more changes would be needed beyond that without testing

martinfleis:
Closes #1937

### DIFF
--- a/.github/workflows/test_code.yml
+++ b/.github/workflows/test_code.yml
@@ -14,6 +14,9 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.8", "3.12"]
       fail-fast: false
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


martinfleis:
Once everyone is happy with the approach I took here, I'll mirror it to branca.

ocefpaf:
I'm OK with this approach but I'm afraid our tests are not robust enough to catch any failures. The way we use numpy here is unchecked and will produce passing status even when it is wrong. As you confirmed in https://github.com/python-visualization/branca/pull/163.

ocefpaf (approved):

Conengmo (commented):
Agree with Filipe that the code tests probably don't help here. But it doesn't hurt to test against the nightly releases either? 

Not sure if I would run the tests a second time. I'd think that makes the output harder to understand. Or is this a temporary addition? In that case it doesn't matter as much.

We might by the way have more success with the notebook/docs selenium tests, as those also check for Javascript errors (or at least I hope that's the case)

martinfleis:
> Not sure if I would run the tests a second time. I'd think that makes the output harder to understand. Or is this a temporary addition? In that case it doesn't matter as much.

Not sure what your suggestion here is then. Split it into two separate runs? Generally, it is better to test against nightly versions of our dependencies, assuming the test will catch issues, so I would try to keep it as permanent. In other projects I maintain, we have dev as another environment alongside the others (latest, oldest...)

Conengmo:
Looks good, let's do it!

Issue #1859, opened 2024-01-09
Max zoom level no longer working from 0.14.0 to 0.15.1 (defaults to 18)


Stephaneke:
**Describe the bug**
Since the update from 0.14.0 to 0.15.1, the max zoom level function does no longer work and defaults to 18?

**To Reproduce**

```
[ Include a code snippet that produces your bug. Make it standalone, we should be able to run it. ]

import folium
m = folium.Map(location=[51.505492, -0.075408], zoom_start=12, max_zoom=25)
# add a circle marker of 5 radius at scale at the center of the map
folium.CircleMarker(location=[51.505492, -0.075408], radius=5, color='#3186cc', fill=True, fill_color='#3186cc').add_to(m)
m.save('interactive_map_0.14.0.html')

[ Include a data sample or link to your data if necessary to run the code ]

```

**Expected behavior**
I expect the maximum zoom level to be 25 instead of 18

**Environment (please complete the following information):**

[maps.zip](https://github.com/python-visualization/folium/files/13875810/maps.zip)

**Possible solutions**
List any solutions you may have come up with.

folium is maintained by volunteers. Can you help making a fix for this issue?

Conengmo:
Can confirm. In the `TileLayer` class we now default to the `max_zoom` that's defined in the `xyzservices.TileProvider` object if available. 

Also note how this works in tandem with `max_native_zoom`.

We should fix this and improve documentation.

Stephaneke:
So there is still a possibility to increase the zoom level? Or is it disabled in the newest updates? If you are working on a small scale (e.g. meters), this is almost a necessity.

hansthen:
@Stephaneke The maximum zoom is defined by the tile provider. You can play around with different tile layers to see how that affects the max zoom level. E.g. `folium.TileLayer('CartoDB Positron').add_to(m)` gives me a maximum zoom level of 20. OpenStreetMap (the default) goes to 19.

Our documentation is off as the `max_zoom` and `max_native_zoom` parameters seem to be ignored.

hansthen:
@Stephaneke I did a quick experiment with a different provider. This one goes to level 22, which is the best I could find. It does require an api key. See https://github.com/leaflet-extras/leaflet-providers/blob/master/leaflet-providers.js for other providers.

```
folium.TileLayer(
     'https://tile.jawg.io/jawg-dark/{z}/{x}/{y}{r}.png?access-token=<get your own>',
      attr="jawg.io",
[...]

PR #1952, opened 2024-05-18
repair max_zoom and max_native_zoom

Conengmo:
Fix https://github.com/python-visualization/folium/issues/1859

When we introduced xyzservices, we changed the way we handle the `max_zoom` and `max_native_zoom` arguments of `TileLayer` (and `Map`). In this PR, restore how it worked before that change, but keep using the values from xyzservices.

To do that, make `min_zoom` and `max_zoom` optional arguments with default values None. That way, when a user doesn't provide them, we can use the value from xyzservices.

When a user does provide explicit values for e.g. `max_zoom`, honor that value and don't overwrite it from xyzservices. This allows zooming past the highest native zoom level.

Fill the `max_native_zoom` option from xyzservices, unless a user provides an explicit value.

I opted to keep the defaults of 0 and 18, because those are the defaults set by Leaflet as well. This way, we make them explicit to our users.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -112,10 +112,12 @@ class Map(JSCSSMixin, MacroElement):
         pass a custom URL, pass a TileLayer object,
         or pass `None` to create a map without tiles.
         For more advanced tile layer options, use the `TileLayer` class.
-    min_zoom: int, default 0
+    min_zoom: int, optional, default 0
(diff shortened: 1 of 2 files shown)


hansthen (approved):
Looks good to me.

Conengmo:
Thanks for the review Hans!

PR #1959, opened 2024-06-01
Add Evented base class

hansthen:
In Leaflet `Evented` is the parent class of `Layer` and `Map`. 

It adds the `on` method, which can be used to add event handlers for the two type of Leaflet objects. In Leaflet it also adds methods to fire events and turn `off` event handlers. I did not implement those as they would be less relevant within Folium.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -32,7 +32,7 @@ class Evented(MacroElement):
     method using python keyword arguments.
     """
 
-    def on(self, **event_map):
+    def on(self, **event_map: JsCode):
         for event_type, handler in event_map.items():
             self.add_child(EventHandler(event_type, handler))
 
--- a/folium/map.py
+++ b/folium/map.py
@@ -32,9 +32,6 @@ class Evented(MacroElement):
     method using python keyword arguments.
     """
 
-    def __init__(self):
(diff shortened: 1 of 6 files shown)


folium/map.py
```diff
@@ -21,7 +21,26 @@
...
+
+    See https://leafletjs.com/reference.html#evented for
+    more in depth documentation. Please note that we have
+    only added the `on(<Object> eventMap)` variant of this
+    method using python keyword arguments.
+    """
+
+    def __init__(self):
+        super().__init__()
+
```
Conengmo, line 37: this init is not necessary I think, without it it will just go directly to the init of the parent class

```suggestion
```

folium/map.py
```diff
@@ -21,7 +21,26 @@
...
+    def on(self, **event_map):
```
Conengmo, line 38: can we add a type hint for this?

```suggestion
    def on(self, **event_map: JsCode):
```

Conengmo (changes requested):
Looks good! Let's do this. I added two small comments, but should be good to afterwards. We'll look into adding more to the docs later.

Conengmo (approved):

PR #1960, opened 2024-06-02
Correct type annotations in Realtime

hansthen:
This PR corrects a few type annotations for the `Realtime` plugin.

1. Realtime superclass in leaflet is GeoJson (which is a subclass of featuregroup). In Folium I cannot make Realtime a subclass of GeoJson since GeoJson requires features to be present before rendering. I made it a subclass of FeatureGroup to more clearly document that features can be added to a Realtime layer.

2. The container parameter for Realtime cannot just be any `L.Layer`. It must be a `FeatureGroup` or something that allows adding features.

I created type annotations to reflect this on the Folium side.

### DIFF
--- a/folium/plugins/realtime.py
+++ b/folium/plugins/realtime.py
@@ -1,14 +1,14 @@
 from typing import Optional, Union
 
-from branca.element import MacroElement
 from jinja2 import Template
 
 from folium.elements import JSCSSMixin
-from folium.map import Layer
+from folium.features import GeoJson
+from folium.map import FeatureGroup
 from folium.utilities import JsCode, camelize, parse_options
 
 
-class Realtime(JSCSSMixin, MacroElement):
+class Realtime(JSCSSMixin, FeatureGroup):
     """Put realtime data on a Leaflet map: live tracking GPS units,
     sensor data or just about anything.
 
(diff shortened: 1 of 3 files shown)


folium/plugins/realtime.py
```diff
@@ -109,7 +109,7 @@ def __init__(
         get_feature_id: Union[JsCode, str, None] = None,
         update_feature: Union[JsCode, str, None] = None,
         remove_missing: bool = False,
-        container: Optional[Layer] = None,
+        container: Optional[Union[FeatureGroup, GeoJson]] = None,
```
Conengmo, line 112: this should also be updated in the docstring

Conengmo (changes requested):
Small comment about updating the docstring, good to go afterwards!

folium/plugins/realtime.py
hansthen: Done

Conengmo (approved):
Don't know why that test failed, maybe a hickup and we should try rerunning it? It's approved anyway.

Conengmo:
@ocefpaf is looking into that test failure in https://github.com/python-visualization/folium/pull/1978, so no need to do that here fortunately. Maybe you can revert the latest commit, then merge it?

ocefpaf:
Keep merging things without worrying about that Windows failure. It is a problem with the dependencies in conda-forge and we are working to fix it upstream.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 branca>=0.6.0
+fiona
```
Conengmo, line 2: @hansthen this got merged with this added dependency in the requirements. I don’t think that’s right, since it’s not a required dependency. Could you maybe make a PR to revert this?

PR #1965, opened 2024-06-07
Automatically create docs PR on release

merschformann:
# Description

This PR means to automatically open a PR with necessary changes to the `switcher.json` in order to integrate newly released docs.

## Prerequisites

In order to use this functionality the following needs to be configured on repository side:

- Requires _"Allow GitHub Actions to create and approve pull requests"_ to be enabled in the repository settings.

## Changes

- Adds `docs/update_switcher.py` which is used to include the new version  and bumping the _latest_ in `switcher.json`
- Opens a PR with above changes after a **release** and requests @Conengmo & @ocefpaf for reviews
  - I wasn't sure about the reviewers, happy to adapt the list :blush: 
- Introduces a _dev_ version in `switcher.json` which points to the latest docs from the _main_ branch


merschformann:
Let me know if I can be of any assistance here. It doesn't seem to me that the test fail is caused by my changes:

```txt
ValueError: 'M' is no longer supported for offsets. Please use 'ME' instead.
```

It seems to be emerging from [here](https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/tslibs/offsets.pyx#L4718-L4723) in pandas.

.github/workflows/deploy-docs.yml
```diff
@@ -9,13 +9,24 @@ on:
...
+  contents: write
+  pull-requests: write
+
 jobs:
   run:
     runs-on: ubuntu-latest
+
+    env:
+      # Define the reviewers that will be assigned to the automatically created PR.
+      REVIEWERS: "Conengmo,ocefpaf"
```
Conengmo, line 23:
```suggestion
```

.github/workflows/deploy-docs.yml
```diff
@@ -38,13 +49,30 @@ jobs:
...
+    - name: Create PR
+      if: ${{ github.event_name == 'release' }}
+      uses: peter-evans/create-pull-request@v3
+      with:
+        commit-message: "docs: Update switcher.json for ${{ github.ref_name }}"
+        title: "docs: Update switcher.json for ${{ github.ref_name }}"
+        body: "This PR updates the switcher.json file."
+        branch: "docs/update-switcher-${{ github.ref_name }}"
+        base: "main"
+        reviewers: ${{ env.REVIEWERS }}
```
Conengmo, line 66:
```suggestion
```

Conengmo (commented):
This is great, thanks!

That test issue is unrelated indeed, it should be fixed in main branch now. I'll update your branch before merging to be sure. 

I made a small change to not add a reviewer to the PR, no need, one of us will get it.

.github/workflows/deploy-docs.yml
```diff
@@ -9,13 +9,20 @@ on:
...
+
 jobs:
   run:
     runs-on: ubuntu-latest
+
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
+        ref: main
```
ocefpaf, line 25: @Conengmo and @merschformann do you need this? That means any change in the code won't be tested in the docs until it is on `main`.

merschformann: This is here to make sure that docs are only ever released from the _main_ branch, thus, reflecting docs in the state the _main_ branch is in. A push to _main_ is also a trigger of this workflow.
It does not feel fully necessary, if the events would change to only be pushes to _main_ and `release` (this one could technically result in none-_main_ being released if release is defined on branch). I am not sure what `pull_request` is for. It is just explicit about limiting docs releases to be coming from _main_.
(just my two cents)

Maybe the triggers of the workflow and the checkout should be revised? :thinking: 

cc @Conengmo

PR #1967, opened 2024-06-13
Fix tests

ocefpaf:
Should fix the failing test on latest numpy/pandas.

### DIFF
--- a/tests/plugins/test_time_slider_choropleth.py
+++ b/tests/plugins/test_time_slider_choropleth.py
@@ -5,9 +5,11 @@ tests TimeSliderChoropleth
 """
 
 import json
+import sys
 
 import numpy as np
 import pandas as pd
+import pytest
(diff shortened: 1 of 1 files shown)


tests/plugins/test_time_slider_choropleth.py
```diff
@@ -5,16 +5,19 @@
...
 import pandas as pd
+import pytest
 from branca.colormap import linear
 
 import folium
 from folium.plugins import TimeSliderChoropleth
 from folium.utilities import normalize
 
 
+@pytest.mark.xfail(sys.version_info[0:2] == (3, 8), reason="too modern for py38")
```
ocefpaf, line 20: We should probably drop Python 3.8 in October at its EOL. This should be fine for now.

ocefpaf:
Self-merging this one b/c it is mostly harmless.

PR #1968, opened 2024-06-14
Fix crs deprecation in geopandas

ocefpaf:
Another minor test update.


Conengmo:
This is for the upcoming Geopandas 1.0 release I assume? Good that you tested it.

PR #1978, opened 2024-06-17
Explicitly requires fiona for testing

ocefpaf:
Fixes the failures we are seeing in #1976 due to geopandas changes.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,6 +2,7 @@ altair>=5.0.0
 cartopy
 check-manifest
 descartes
+fiona
 flake8
 flake8-builtins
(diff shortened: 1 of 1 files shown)


ocefpaf:
Looks like we have a problem with the packages on Windows.

```
Importing fiona resulted in: DLL load failed while importing _env: The specified module could not be found.
```

martinfleis:
Coming from https://github.com/conda-forge/minizip-feedstock/issues/17

ocefpaf:
This PR is no longer needed b/c the root of the problem was not the lack of fiona but a bad Windows library in the dependency chain. That was resolved upstream. Closing this one.

PR #2001, opened 2024-10-05
Add feature_group parameter to Draw plugin

hansthen:
This can be used to pass existing layer objects into the Draw plugin, which the user can then edit.

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -12,6 +12,9 @@ class Draw(JSCSSMixin, MacroElement):
     ----------
     export : bool, default False
         Add a small button that exports the drawn shapes as a geojson file.
+    feature_group : FeatureGroup, optional
+        The FeatureGroup object that will hold the editable figures. This can
+        be used to initialize the Draw plugin with predefined Layer objects.
     filename : string, default 'data.geojson'
         Name of geojson file
     position : {'topleft', 'toprigth', 'bottomleft', 'bottomright'}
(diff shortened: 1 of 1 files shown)


folium/plugins/draw.py
```diff
@@ -50,26 +53,36 @@ class Draw(JSCSSMixin, MacroElement):
...
+                        var coords = JSON.stringify(layer.toGeoJSON());
+                        {%- if this.show_geometry_on_click %}
+                        layer.on('click', function() {
+                            alert(coords);
+                            console.log(coords);
+                        });
+                        {%- endif %}
+                        drawnItems_{{ this.get_name() }}.addLayer(layer);
+                    }
+            );
```
Conengmo, line 85: please don't mix styling changes with functional changes. Seems this code didn't change functionally, so don't include it in this PR then please.

Conengmo (changes requested):
Love this change, seems really useful. I added one comment about a code block change that doesn't seem necessary.

I feel like this could benefit from an example in the user guide regarding how this option can be used. Not necessary to get this merged, but if you feel like it: https://python-visualization.github.io/folium/latest/user_guide/plugins/draw.html

hansthen:
> Love this change, seems really useful. I added one comment about a code block change that doesn't seem necessary.
> 
> I feel like this could benefit from an example in the user guide regarding how this option can be used. Not necessary to get this merged, but if you feel like it: https://python-visualization.github.io/folium/latest/user_guide/plugins/draw.html

Will add the documentation when I have a bit more time.

folium/plugins/draw.py
hansthen: Reverted.

Conengmo (approved):

PR #2003, opened 2024-10-06
Add a date_option (format string) parameter to the choropleth timeline slider

### LINKED ISSUE #1997: Format time to show year only in choropleth time slider
There may already be a way to do this, but I was not able to stumble on it...

I would like to be able to control the data format in the slider bar in a time slider choropleth.  In my case (and I believe many others), I am describing annual trends and all that is required in the slider is the year.

I am requesting:
1) Ability to set the time format displayed in the choropleth time slider itself by passing a date format. or setting a "show year only" variable, or if this is already present,
2) Enhanced documentation describing how this can be modified.

hansthen:
Closes #1997

### DIFF
--- a/folium/plugins/time_slider_choropleth.py
+++ b/folium/plugins/time_slider_choropleth.py
@@ -19,6 +19,8 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
     styledict: dict
         A dictionary where the keys are the geojson feature ids and the values are
         dicts of `{time: style_options_dict}`
+    date_options: str, default "ddd MMM DD YYYY"
+        A format string to render the currently active time in the control.
     highlight: bool, default False
         Whether to show a visual effect on mouse hover and click.
     name : string, default None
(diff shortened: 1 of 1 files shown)


folium/plugins/time_slider_choropleth.py
```diff
@@ -19,6 +19,8 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
     styledict: dict
         A dictionary where the keys are the geojson feature ids and the values are
         dicts of `{time: style_options_dict}`
+    date_options: str, default "ddd MMM DD YYYY"
```
Conengmo, line 22: later on it's called `date_format`, why not call it that here as well?

folium/plugins/time_slider_choropleth.py
```diff
@@ -19,6 +19,8 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
...
+        A format string to render the currently active time in the control.
```
Conengmo, line 23: should we specify what formatting language is being used here? Like, what fields like MMM DD are available?

Conengmo (commented):
Looks good! Two comments that are more questions/suggestions on how to improve, but not critical.

folium/plugins/time_slider_choropleth.py
hansthen: To be consistent with the other plugins. `timestamped_geojson` and `timeline` also use `date_options`. I'm not so happy about the name

folium/plugins/time_slider_choropleth.py
hansthen: If you feel strongly about it, I can add a link to the formatting library used. However, I don't think it is necessary. The time string formatting language is fairly intuitive (and fairly universal). Also, I would like the documentation to be consistent between the different plugins.

folium/plugins/time_slider_choropleth.py
Conengmo: That makes sense 👍

Conengmo (approved):

PR #2007, opened 2024-10-17
make switcher match gh-pages

Conengmo:
https://github.com/python-visualization/folium/pull/1965 was a really helpful change to get our docs versioning updated automatically on release. It got us a large way there, but not fully. I noticed while the dropdown in the UI is updated correctly, it doesn't fully correspond to the folder situation on our `gh-pages` branch. See https://github.com/python-visualization/folium/tree/gh-pages.

The issue is we have a main `index.html` that redirects to `latest/index.html`. To make things easy, we'd ideally not have to update this redirect, but always put the latest release version in the `stable` branch.

Furthermore, I'd like it if there are not two entries in the dropdown for the same version. So 'latest' should now refer to v0.17.0, and there shouldn't also be a v0.17.0 option. And 'latest' should make clear it refers to v0.17.0, by including a name like "latest (v0.17.0).

- I updated @merschformann's script to include these two changes. 
- I updated the doc creation workflow to create two folders on release: one with the versioned folder name and one in `latest/`.
- I updated the `switcher.json` to be prepared for the new version of the script.

@merschformann don't know if you have time and are interested to review this one? Thanks again for your original work, made it easy to make this step.


docs/update_switcher.py
```diff
@@ -21,31 +22,32 @@ def main():
...
+            previous_version = re.search(r"latest \(([v.\d]+)\)", version["name"]).group(1)
+            if previous_version == args.version:
+                print(f"Version {args.version} already is the latest version. Exiting.")
+                return
+
+            # now replace the name of this one with the new version
+            switcher[i]["name"] = f"latest ({args.version})"
             break
-    if latest_index is None:
+    else:
```
merschformann, line 37: I didn't even realize this kind of `for` / `if` / `break` / `else` combo was possible. TIL, interesting. :+1:

.github/workflows/deploy-docs.yml
```diff
@@ -70,10 +70,19 @@ jobs:
...
         destination_dir: ${{ github.ref_name }}
+
+    - name: Publish to Github Pages on release (latest)
+      if: ${{ github.event_name == 'release' }}
+      uses: peaceiris/actions-gh-pages@v4
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: docs/_build/html/
+        destination_dir: latest
+        keep_files: false
```
merschformann, line 88: If I read this correctly, these 2 lines will "re-deploy" to _latest_ every time this executes, thus, having the latest docs always available in that dir. Nice and simple. Makes sense to me. :relaxed:

merschformann (approved):
Nice work! Makes a lot of sense to me. :raised_hands:

Conengmo:
Thanks Marius! Appreciate you taking the time to take a look at this. I'll give this a try once we fixed the publish docs workflow.

Conengmo:
pre-commit.ci autofix

Issue #2002, opened 2024-10-06
Publish docs test fails due to missing fiona dependency


hansthen:
**Describe the bug**
The automated tests are broken on main. There is an error in the publish docs test. 

**To Reproduce**
See this dependabot commit https://github.com/python-visualization/folium/commit/dc847327e02046befd87a3cee5f6b4e53e6866da.

**Additional context**
I tried to reproduce the issues locally. Locally, however, I do not have these errors. Perhaps someone can help write instructions on how to setup a sandboxed environment like in github to run these tests.

prusswan:
For the fiona import error, I suspect it has to do with geopandas dropping fiona as a dependency (now replaced with pyogrio) since 1.0.0 (see https://github.com/geopandas/geopandas/blob/main/CHANGELOG.md#version-100-june-24-2024) - so it has to be imported if any of the workflow runs require it. Can't "prove" this as I could not find the logs for a previous successful run to compare against.

Conengmo:
Fixed by https://github.com/python-visualization/folium/pull/2012

PR #2008, opened 2024-10-17
Fix docs workflow failure due to Fiona package

Conengmo:
Closes https://github.com/python-visualization/folium/issues/2002

Add `fiona` as an explicit dev dependency.

It is needed in this example in the docs https://python-visualization.github.io/folium/latest/user_guide/geojson/geopandas_and_geo_interface.html#Use-any-object-with-__geo_interface__

The other issue is with the heatmap verification. Seems like Chrome added some UI element to the headless browser that changed the window size. So I updated the verification image to correspond with the new size.

But now the size differ between my local computer and the remove env. I guess we should either rewrite this test to not be dependent on screen size or scrap it altogether.


Conengmo:
I added `fiona` as a dependency but it's not being installed in Miniconda... weird

Conengmo:
Can't get Fiona to install from the requirements. Instead I just install it with Pip explicitely. Not the nicest solution but it does work.

.github/workflows/deploy-docs.yml
```diff
@@ -33,6 +33,11 @@ jobs:
           --file requirements.txt
           --file requirements-dev.txt
 
+    # For some reason Fiona doesn't get installed from requirements-dev...
+    - name: Install fiona dependency
+      shell: bash -l {0}
+      run: python -m pip install fiona
```
ocefpaf, line 39: If you can wait I'll take a look at this next week. Something is not right here...

Conengmo: something's weird right! I can definitely wait, very interesting to figure out what's going on.

Note that https://github.com/python-visualization/folium/pull/2007 is waiting for the fix to get the docs going again. Waiting a week is not an issue, but if it's much longer then maybe a quick fix is better. Definitely not meant to rush you! More of an expectation thing.

I did also manage to fix the Selenium heatmap test issue in this PR, but I'll spin that out in a separate one and merge that earlier.

hansthen: I split the underlying issues for the Selenium heatmap and this change. Once we have both fixed shall we make a new release?

Conengmo: sounds good!

ocefpaf: > something's weird right! I can definitely wait, very interesting to figure out what's going on.

Found it. See #2012. Basically we are limiting to what is on `main` to build/test the docs.

Issue #2009, opened 2024-10-19
Error in selenium tests when generating heatmap


hansthen:
**Describe the bug**
The automated selenium tests are broken on main. There is an error in generating the heatmap.

**To Reproduce**
See this dependabot commit https://github.com/python-visualization/folium/commit/dc847327e02046befd87a3cee5f6b4e53e6866da.

**Additional context**
I tried to reproduce the issues locally. Locally, however, I do not have these errors. Perhaps someone can help write instructions on how to setup a sandboxed environment like in github to run these tests.

PR #2010, opened 2024-10-19
Fix Selenium heatmap test by using fixed figure size

Conengmo:
Closes https://github.com/python-visualization/folium/issues/2009. Taken out of https://github.com/python-visualization/folium/pull/2008.

It seems like Chrome added some UI element to the headless browser that changed the window size. It's now also inconsistent between my local machine and the CI/CD server. To fix it, I applied a fixed figure size, then crop the screenshot to the figure. That way if the window is a few tens of pixels heigher of shorter, it doesn't matter.

Pre-commit failure is unrelated, see https://github.com/python-visualization/folium/pull/2011


Conengmo:
pre-commit.ci autofix

Conengmo:
pre-commit.ci autofix

hansthen (approved):
LGTM

PR #2012, opened 2024-10-21
add fiona and ensure latest geopandas

ocefpaf:
Testing if we can build the docs properly. Alternative for #2008.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -25,7 +25,6 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-        # ref: main
 
     - name: Setup Micromamba env
       uses: mamba-org/setup-micromamba@v2
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -17,6 +17,9 @@ permissions:
 jobs:
   run:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -17,28 +17,30 @@ permissions:
...
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-        ref: main
+        # ref: main
```
ocefpaf, line 28: @Conengmo this is the culprit. I'm not sure if we need this though.

xref,: https://github.com/python-visualization/folium/pull/1965/files#r1808652059

Conengmo: good catch! Pretty sure we don't need that. We already have if-statements in that workflow to make sure we do certain things only on main branch or on release. I think it got in there because I missed it in review, my bad!

```suggestion
```

Conengmo:
Looks good, but now the code tests on Python 3.8 are failing. Looks like 3.8 is EOL since last week though, so maybe updating those tests from 3.8 and 3.12 to 3.9 and 3.13 will help?

Conengmo (approved):
Everything passes now!

Thanks for finding the issue @ocefpaf!

ocefpaf:
> Testing for 3.13 is not working yet it seems.

I'm working on this upstream. It should be OK in 1-2 days.

Conengmo:
If it's that quick then let's include it in the upcoming release.

PR #2013, opened 2024-10-21
Py313

ocefpaf:
We don't need `twine` in the `requirements-dev.txt` filem in fact, if we move to more pre-commits checks, we can remove even more dependencies listed in that file. I also removed the experimental tests, we no longer need them.

This PR should pass in a few hours as soon as jenkspy is available in the conda-forge channel.

xref.: https://github.com/conda-forge/jenkspy-feedstock/pull/31


Conengmo (approved):
Great!

PR #2018, opened 2024-10-23
Fix docs versioning relating to "v" prefix

Conengmo:
Some more tweaks to our docs versioning and the version switcher. Main issue is a mismatch in how the version is put into a string: whether a "v" prefix is included or not. In the git tag we include a "v" prefix, but in `folium.__version__` we don't have that prefix.

In the new situation:
- the git tag will include a `v` prefix
- the folder in `gh-pages` will include a `v` prefix
- all other places are without that prefix

This change is already live on the `gh-pages` branch, including the manually created docs for v0.18.0.

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -4,9 +4,13 @@
     "url": "https://python-visualization.github.io/folium/dev/"
   },
   {
-    "name": "latest (0.17.0)",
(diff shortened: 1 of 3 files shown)


Conengmo:
Will self-merge this one since it's only about the documentation, and I tested this already.

PR #2023, opened 2024-11-03
Implement the `once` method on Evented.

hansthen:


folium/elements.py
```diff
@@ -107,19 +107,20 @@ class EventHandler(MacroElement):
...
+            {{ this._parent.get_name()}}.{{ this.method }}(
                 {{ this.event|tojson}},
                 {{ this.handler.js_code }}
             );
         {% endmacro %}
         """
     )
 
-    def __init__(self, event: str, handler: JsCode):
+    def __init__(self, event: str, handler: JsCode, once=False):
```
Conengmo, line 118:
```suggestion
    def __init__(self, event: str, handler: JsCode, once: bool = False):
```

folium/map.py
```diff
@@ -34,8 +34,14 @@ class Evented(MacroElement):
     """
 
     def on(self, **event_map: JsCode):
+        self._add(False, **event_map)
```
Conengmo, line 37: I think it's nicer to use keyword arguments when the value doesn't inform what argument it's applying to.

```suggestion
        self._add(once=False, **event_map)
```

folium/map.py
```diff
@@ -34,8 +34,14 @@ class Evented(MacroElement):
...
+
+    def once(self, **event_map: JsCode):
+        self._add(True, **event_map)
```
Conengmo, line 40:
```suggestion
        self._add(once=True, **event_map)
```

folium/map.py
```diff
@@ -34,8 +34,14 @@ class Evented(MacroElement):
...
+
+    def _add(self, once, **event_map: JsCode):
```
Conengmo, line 42:
```suggestion
    def _add(self, once: bool, **event_map: JsCode):
```

Conengmo (approved):
Looks good! I have basically two comments, one about adding type hints, the other is more of a style thing. But nothing functional, so I'll approve it now already.

hansthen:
> Looks good! I have basically two comments, one about adding type hints, the other is more of a style thing. But nothing functional, so I'll approve it now already.

Okay thanks for those.

PR #2029, opened 2024-11-08
Issue 1943 allow jscode in options

hansthen:
Boring but big. The leaflet class hierarchy allows all classes to have optional arguments. These optional arguments can contain JavaScript code or references to other Leaflet elements.

To support this in Folium I replaced (almost) all instances of `parse_options` in the constructors with `this.options | tojavascript` in the templates. I left the `parse_options` in case someone outside of Folium uses it.

### DIFF
--- a/folium/plugins/realtime.py
+++ b/folium/plugins/realtime.py
@@ -100,7 +100,7 @@ class Realtime(JSCSSMixin, FeatureGroup):
         update_feature: Union[JsCode, str, None] = None,
         remove_missing: bool = False,
         container: Optional[Union[FeatureGroup, GeoJson]] = None,
-        **kwargs,
+        **kwargs
     ):
         super().__init__()
         self._name = "Realtime"
--- a/folium/plugins/realtime.py
+++ b/folium/plugins/realtime.py
@@ -1,11 +1,10 @@
 from typing import Optional, Union
 
-from jinja2 import Template
-
 from folium.elements import JSCSSMixin
 from folium.features import GeoJson
 from folium.map import FeatureGroup
-from folium.utilities import JsCode, camelize, parse_options
+from folium.template import Template
+from folium.utilities import JsCode
 
 
 class Realtime(JSCSSMixin, FeatureGroup):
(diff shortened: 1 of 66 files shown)


folium/plugins/realtime.py
```diff
@@ -124,12 +113,6 @@ def __init__(
         if update_feature is not None:
             kwargs["update_feature"] = JsCode(update_feature)
         kwargs["remove_missing"] = remove_missing
+        kwargs["container"] = container
```
Conengmo, line 116:
```suggestion
        kwargs["container"] = JsCode(container.get_name())
```

You  made this class right, so you probably know better. But I think it needs the Javascript name as JS code.

folium/utilities.py
```diff
@@ -365,7 +365,8 @@ def temp_html_filepath(data: str) -> Iterator[str]:
         yield filepath
     finally:
         if os.path.isfile(filepath):
-            os.remove(filepath)
+            pass
+            # os.remove(filepath)
```
Conengmo, line 369: Guess this is an artefact from testing? Should be restored.

```suggestion
            os.remove(filepath)
```

tests/plugins/test_grouped_layer_control.py
```diff
@@ -34,10 +35,11 @@ def test_grouped_layer_control():
...
-            {{"exclusiveGroups": ["groups1"]}},
+            {{"exclusiveGroups": ["groups1",],}},
          ).addTo({m.get_name()});
          {fg2.get_name()}.remove();
     """
-        )
-        in out
     )
+    print(out)
+    print(expected)
```
Conengmo, line 44:
```suggestion
```

tests/plugins/test_marker_cluster.py
```diff
@@ -61,6 +61,8 @@ def test_marker_cluster():
         in out
     )  # noqa
 
+    print(out)
+    print(expected)
```
Conengmo, line 65:
```suggestion
```

tests/test_folium.py
```diff
@@ -464,28 +464,31 @@ def test_global_switches(self):
         m = folium.Map(prefer_canvas=True)
         out = m._parent.render()
         out_str = "".join(out.split())
-        assert "preferCanvas:true" in out_str
+        for line in out.split("\n"):
+            print(line)
+        print(out_str)
```
Conengmo, line 469:
```suggestion
```

tests/test_raster_layers.py
```diff
@@ -65,6 +65,7 @@ def test_wms():
     )
     w.add_to(m)
     html = m.get_root().render()
+    print(html)
```
Conengmo, line 68:
```suggestion
```

Conengmo (changes requested):
Boring but big 😄 Kudos for taking this on. It looks good! I have two small comments, plus some suggestions to remove added print statements from tests. Other than that looks good!

folium/plugins/realtime.py
hansthen: We test specifically for Branca `Element`s in `tojavascript` so it's not necessary. 
```     
elif isinstance(obj, Element):
        return obj.get_name()
```

Conengmo: Nice!

Conengmo (approved):
Let's go! 🚀

hansthen:
@Conengmo Perhaps you can also have a look at the `raster_layers.WmsTileLayer` class. I still use `{{ this.options|tojson }}` there because there is an option that should remain lowercase. The inconsistency bugs me (a bit). I could not think of a way to make it consistent. But perhaps you have suggestions on how we could use `tojavascript` here?

Conengmo:
Yes, I noticed that one as well. Can't think of a good, simple solution right now, so maybe what you did is best for now.

Issue #901, opened 2018-06-30
Plugin to handle overlapping markers


psychemedia:
```python
import folium

loc=[50.65795773762609, -1.2263860950441539]

m=folium.Map(location=loc, zoom_start=12)
for i in range(5):
    folium.Marker(loc, popup='{}'.format( i )).add_to(m)
m
```

When plotting markers geocoded from things like UK postcodes, it's quite common to get collisions when placing markers (multiple markers at the same location).

The markercluster approach *will* separate out markers that are colocated, but the overall effect can be quite unsatisfactory because the clustering algorithm tries to cluster things all over the place.

One solution is to perturb markers randomly. For example:

```
import numpy as np

m=folium.Map(location=loc, zoom_start=12)
for i in range(5):
    loc[0]=loc[0]+np.random.uniform(0.001, 10**(-20))-0.00005
    loc[1]=loc[1]+np.random.uniform(0.001, 10**(-20))-0.00005
    folium.Marker(loc, popup='{}'.format( i )).add_to(m)
m
```
What would be really useful would be a plugin to separate out colocated markers. [jawj/OverlappingMarkerSpiderfier-Leaflet](https://github.com/jawj/OverlappingMarkerSpiderfier-Leaflet) looks like it might be a useful candidate?

#### Expected Output

*Separate out colocated markers* but  __not__ *cluster markers AND separate out colocated ones*.

2Shabby:
Is this issue resolved? Would like to help contribute to the same if possible.

eraduz:
Anything?

eraduz:
Is this issue resolved?

PR #2033, opened 2024-11-16
Feature overlapping markers

swtormy:
# Add OverlappingMarkerSpiderfier Plugin

### Overview

In the pull request, I added a new OverlappingMarkerSpiderfier plugin to Folium. It solves the problem of overlapping markers on maps by “stretching” them into a distinct, scattered pattern. This feature will allow the user to work with adjacent markers, eliminating the problem described in [#901](https://github.com/python-visualization/folium/issues/901).

### Key Features

- **Spiderfying Markers**: Automatically spreads out overlapping markers when clicked, making them individually accessible.
- **Customizable Options**: Supports options like `keepSpiderfied`, `nearbyDistance`, and `legWeight` (description [here](https://github.com/jawj/OverlappingMarkerSpiderfier-Leaflet)). 
- **Popup Integration**: Displaying their content when markers are clicked.

### Changes in This PR

1. **New Plugin**: Added `OverlappingMarkerSpiderfier` in `folium/plugins/overlapping_marker_spiderfier.py`.
2. **Documentation**: Added a detailed usage guide in `docs/user_guide/plugins/overlapping_marker_spiderfier.md`.
3. **Tests**: Included unit tests in `tests/plugins/test_overlapping_marker_spiderfier.py`.
4. **Import in Init**: Updated `folium/plugins/__init__.py` to include the new plugin.

### Example Usage

Here's a simple example to demonstrate the plugin:

```python
import folium
from folium.plugins import OverlappingMarkerSpiderfier

# Create a map
m = folium.Map(location=[45.05, 3.05], zoom_start=14)

# Generate overlapping markers
markers = [
    folium.Marker(
        location=[45.05, 3.05], 
        options={'desc': f'Marker {i+1}'}
    ) for i in range(10)
]

# Add the plugin to the map
oms = OverlappingMarkerSpiderfier(
    markers=markers,
    options={'keepSpiderfied': True, 'nearbyDistance': 20}
)
oms.add_to(m)

# Save the map
m.save("example_map.html")
```

## Addressed Issue

This PR resolves [Issue #901](https://github.com/python-visualization/folium/issues/901), which requested a solution for handling overlapping markers on maps. The proposed solution integrates the [OverlappingMarkerSpiderfier Leaflet plugin](https://github.com/jawj/OverlappingMarkerSpiderfier-Leaflet) into Folium.

## Testing

- Ran all existing and new tests to ensure the changes do not break existing functionality.
```bash
(venv) C:\Users\swtor\Desktop\python\folium>python -m pytest tests --ignore=tests/selenium
================================================================== test session starts ===================================================================
platform win32 -- Python 3.12.3, pytest-8.3.3, pluggy-1.5.0
rootdir: C:\Users\swtor\Desktop\python\folium
configfile: pyproject.toml
plugins: anyio-4.6.2.post1, nbval-0.11.0
collected 205 items

tests\plugins\test_antpath.py .                                                                                                                     [  0%]
tests\plugins\test_beautify_icon.py .                                                                                                               [  0%]
tests\plugins\test_boat_marker.py ..                                                                                                                [  1%]
tests\plugins\test_dual_map.py .                                                                                                                    [  2%]
tests\plugins\test_encoded.py ..                                                                                                                    [  3%]
tests\plugins\test_fast_marker_cluster.py ....                                                                                                      [  5%]
tests\plugins\test_feature_group_sub_group.py .                                                                                                     [  5%]
tests\plugins\test_float_image.py .                                                                                                                 [  6%]
tests\plugins\test_fullscreen.py .                                                                                                                  [  6%]
tests\plugins\test_grouped_layer_control.py .                                                                                                       [  7%]
tests\plugins\test_heat_map.py ...                                                                                                                  [  8%]
tests\plugins\test_heat_map_withtime.py .                                                                                                           [  9%]
tests\plugins\test_marker_cluster.py .                                                                                                              [  9%]
tests\plugins\test_minimap.py .                                                                                                                     [ 10%]
tests\plugins\test_overlapping_marker_spiderfier.py .                                                                                               [ 10%]
tests\plugins\test_pattern.py .                                                                                                                     [ 11%]
tests\plugins\test_polyline_offset.py ....                                                                                                          [ 13%]
tests\plugins\test_polyline_text_path.py .                                                                                                          [ 13%]
tests\plugins\test_realtime.py .                                                                                                                    [ 14%]
tests\plugins\test_scroll_zoom_toggler.py .                                                                                                         [ 14%] 
tests\plugins\test_semicircle.py .                                                                                                                  [ 15%]
tests\plugins\test_tag_filter_button.py .                                                                                                           [ 15%]
tests\plugins\test_terminator.py .                                                                                                                  [ 16%]
tests\plugins\test_time_slider_choropleth.py .                                                                                                      [ 16%]
tests\plugins\test_timeline.py .                                                                                                                    [ 17%]
tests\plugins\test_timestamped_geo_json.py .                                                                                                        [ 17%] 
tests\plugins\test_vectorgrid_protobuf.py ...                                                                                                       [ 19%]
tests\test_features.py ....................                                                                                                         [ 28%]
tests\test_folium.py ....................                                                                                                           [ 38%]
tests\test_jinja.py ..............                                                                                                                  [ 45%] 
tests\test_map.py ................                                                                                                                  [ 53%]
tests\test_raster_layers.py ......                                                                                                                  [ 56%]
tests\test_repr.py ......                                                                                                                           [ 59%]
tests\test_template.py .........                                                                                                                    [ 63%] 
tests\test_utilities.py ....................................................................                                                        [ 96%]
tests\test_vector_layers.py .......                                                                                                                 [100%]

================================================================== 205 passed in 48.18s ==================================================================
```
```bash
(venv) C:\Users\swtor\Desktop\python\folium>python -m pytest tests/selenium
================================================================== test session starts =================================================================== 
================================================================== test session starts =================================================================== 
platform win32 -- Python 3.12.3, pytest-8.3.3, pluggy-1.5.0
platform win32 -- Python 3.12.3, pytest-8.3.3, pluggy-1.5.0
rootdir: C:\Users\swtor\Desktop\python\folium
configfile: pyproject.toml
plugins: anyio-4.6.2.post1, nbval-0.11.0
collected 70 items                                                                                                                                         

tests\selenium\test_geojson_selenium.py
DevTools listening on ws://127.0.0.1:52531/devtools/browser/699f8db3-b007-48b7-866d-dc8f09834ce4
.                                                                                                           [  1%]
tests\selenium\test_heat_map_selenium.py .                                                                                                          [  2%] 
tests\selenium\test_selenium.py ....Created TensorFlow Lite XNNPACK delegate for CPU.
...................................................x............                                                [100%]

======================================================= 69 passed, 1 xfailed in 206.61s (0:03:26) ========================================================
```
- Successfully tested map visualizations with overlapping markers.
Screenshot 2024-11-16 в 10 44 05 AM

PS: Feedback is welcome! If there are additional improvements or refinements needed, I am happy to make the necessary adjustments.

### DIFF
--- a/folium/plugins/overlapping_marker_spiderfier.py
+++ b/folium/plugins/overlapping_marker_spiderfier.py
@@ -1,18 +1,20 @@
 from jinja2 import Template
+
 from folium.elements import JSCSSMixin, MacroElement
 from folium.utilities import parse_options
+
 
 class OverlappingMarkerSpiderfier(JSCSSMixin, MacroElement):
     """
     A plugin that handles overlapping markers on a map by spreading them out in a spiral or circle pattern when clicked.
-    
+
     This plugin is useful when you have multiple markers in close proximity that would otherwise be difficult to interact with.
-    When a user clicks on a cluster of overlapping markers, they spread out in a 'spider' pattern, making each marker 
+    When a user clicks on a cluster of overlapping markers, they spread out in a 'spider' pattern, making each marker
     individually accessible.
 
-    Markers must be added to the map **before** calling `oms.add_to(map)`. 
+    Markers must be added to the map **before** calling `oms.add_to(map)`.
     The plugin identifies and manages all markers already present on the map.
-      
+
     Parameters
     ----------
     options : dict, optional
@@ -25,16 +27,15 @@ class OverlappingMarkerSpiderfier(JSCSSMixin, MacroElement):
             Weight of the spider legs
         - circleSpiralSwitchover : int, optional
             Number of markers at which to switch from circle to spiral pattern
-    
+
     Example
     -------
-    >>> oms = OverlappingMarkerSpiderfier(options={
-    ...     "keepSpiderfied": True,
-    ...     "nearbyDistance": 30,
-    ...     "legWeight": 2.0
-    ... })
+    >>> oms = OverlappingMarkerSpiderfier(
+    ...     options={"keepSpiderfied": True, "nearbyDistance": 30, "legWeight": 2.0}
+    ... )
     >>> oms.add_to(map)
     """
+
     _template = Template(
         """
         {% macro script(this, kwargs) %}
--- a/folium/plugins/overlapping_marker_spiderfier.py
+++ b/folium/plugins/overlapping_marker_spiderfier.py
@@ -1,100 +1,67 @@
 from jinja2 import Template
-
-from folium.elements import JSCSSMixin
-from folium.map import Layer
+from folium.elements import JSCSSMixin, MacroElement
 from folium.utilities import parse_options
 
+class OverlappingMarkerSpiderfier(JSCSSMixin, MacroElement):
+    """
+    A plugin that handles overlapping markers on a map by spreading them out in a spiral or circle pattern when clicked.
+    
+    This plugin is useful when you have multiple markers in close proximity that would otherwise be difficult to interact with.
+    When a user clicks on a cluster of overlapping markers, they spread out in a 'spider' pattern, making each marker 
+    individually accessible.
 
-class OverlappingMarkerSpiderfier(JSCSSMixin, Layer):
-    """A plugin that handles overlapping markers by spreading them into a spider-like pattern.
-
-    This plugin uses the OverlappingMarkerSpiderfier-Leaflet library to manage markers
-    that are close to each other or overlap. When clicked, the overlapping markers
-    spread out in a spiral pattern, making them easier to select individually.
-
+    Markers must be added to the map **before** calling `oms.add_to(map)`. 
+    The plugin identifies and manages all markers already present on the map.
+      
     Parameters
     ----------
-    markers : list, optional
-        List of markers to be managed by the spiderfier
-    name : string, optional
-        Name of the layer control
-    overlay : bool, default True
-        Whether the layer will be included in LayerControl
-    control : bool, default True
-        Whether the layer will be included in LayerControl
-    show : bool, default True
-        Whether the layer will be shown on opening
     options : dict, optional
-        Additional options to be passed to the OverlappingMarkerSpiderfier instance
-        See https://github.com/jawj/OverlappingMarkerSpiderfier-Leaflet for available options
-
+        The options to configure the spiderfier behavior:
+        - keepSpiderfied : bool, default True
+            If true, markers stay spiderfied after clicking
+        - nearbyDistance : int, default 20
+            Pixels away from a marker that is considered overlapping
+        - legWeight : float, default 1.5
+            Weight of the spider legs
+        - circleSpiralSwitchover : int, optional
+            Number of markers at which to switch from circle to spiral pattern
+    
     Example
     -------
-    >>> markers = [marker1, marker2, marker3]  # Create some markers
-    >>> spiderfier = OverlappingMarkerSpiderfier(
-    ...     markers=markers, keepSpiderfied=True, nearbyDistance=20
-    ... )
-    >>> spiderfier.add_to(m)  # Add to your map
+    >>> oms = OverlappingMarkerSpiderfier(options={
+    ...     "keepSpiderfied": True,
+    ...     "nearbyDistance": 30,
+    ...     "legWeight": 2.0
+    ... })
+    >>> oms.add_to(map)
     """
-
     _template = Template(
         """
         {% macro script(this, kwargs) %}
-        var {{ this.get_name() }} = (function () {
-            var layerGroup = L.layerGroup();
-
+        (function () {
             try {
                 var oms = new OverlappingMarkerSpiderfier(
                     {{ this._parent.get_name() }},
                     {{ this.options|tojson }}
                 );
 
-                var popup = L.popup({
-                    offset: L.point(0, -30)
-                });
-
-                oms.addListener('click', function(marker) {
-                    var content;
-                    if (marker.options && marker.options.options && marker.options.options.desc) {
-                        content = marker.options.options.desc;
-                    } else if (marker._popup && marker._popup._content) {
-                        content = marker._popup._content;
-                    } else {
-                        content = "";
-                    }
-
-                    if (content) {
-                        popup.setContent(content);
-                        popup.setLatLng(marker.getLatLng());
-                        {{ this._parent.get_name() }}.openPopup(popup);
-                    }
-                });
-
-                oms.addListener('spiderfy', function(markers) {
+                oms.addListener('spiderfy', function() {
                     {{ this._parent.get_name() }}.closePopup();
                 });
 
-                {% for marker in this.markers %}
-                var {{ marker.get_name() }} = L.marker(
-                    {{ marker.location|tojson }},
-                    {{ marker.options|tojson }}
-                );
+                {{ this._parent.get_name() }}.eachLayer(function(layer) {
+                    if (
+                        layer instanceof L.Marker
+                    ) {
+                        oms.addMarker(layer);
+                    }
+                });
 
-                {% if marker.popup %}
-                {{ marker.get_name() }}.bindPopup({{ marker.popup.get_content()|tojson }});
-                {% endif %}
-
-                oms.addMarker({{ marker.get_name() }});
-                layerGroup.addLayer({{ marker.get_name() }});
-                {% endfor %}
             } catch (error) {
-                console.error('Error in OverlappingMarkerSpiderfier initialization:', error);
+                console.error('Error initializing OverlappingMarkerSpiderfier:', error);
             }
-
-            return layerGroup;
         })();
         {% endmacro %}
-
         """
     )
 
@@ -105,21 +72,9 @@ class OverlappingMarkerSpiderfier(JSCSSMixin, Layer):
         )
     ]
 
-    def __init__(
-        self,
-        markers=None,
-        name=None,
-        overlay=True,
-        control=True,
-        show=True,
-        options=None,
-        **kwargs,
-    ):
-        super().__init__(name=name, overlay=overlay, control=control, show=show)
+    def __init__(self, options=None, **kwargs):
+        super().__init__()
         self._name = "OverlappingMarkerSpiderfier"
-
-        self.markers = markers or []
-
         default_options = {
             "keepSpiderfied": True,
             "nearbyDistance": 20,
@@ -127,5 +82,4 @@ class OverlappingMarkerSpiderfier(JSCSSMixin, Layer):
         }
         if options:
             default_options.update(options)
-
         self.options = parse_options(**default_options, **kwargs)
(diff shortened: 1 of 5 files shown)


folium/plugins/overlapping_marker_spiderfier.py
```diff
@@ -0,0 +1,131 @@
...
+    >>> spiderfier = OverlappingMarkerSpiderfier(
+    ...     markers=markers, keepSpiderfied=True, nearbyDistance=20
+    ... )
+    >>> spiderfier.add_to(m)  # Add to your map
+    """
+
+    _template = Template(
+        """
+        {% macro script(this, kwargs) %}
+        var {{ this.get_name() }} = (function () {
```
hansthen, line 43: I really like how you used a n IFFE here. I think we should use these for all our templates.

hansthen (approved):
Looks good to me. Nice to have this ability.

folium/plugins/overlapping_marker_spiderfier.py
swtormy: Hi! I looked up this method in FastMarkerCluster, so at least one IFFE plugin already uses it)

Conengmo (changes requested):
Thanks for taking this up @swtormy! I love how complete your PR is!

One thing that tripped me up a bit though is whether `marker` objects should be added to the map or a feature group before being added to the `OverlappingMarkerSpiderfier` object. Turns out, looking at your implementation, they should not. Because the `OMS` class does create the markers in JS in its template. But in the docs example the markers are also added to the map.

That issue also led me to question whether we could simplify the template in this plugin class. Currently it contains code to create markers and popups. But we already have the `Marker` and `Popup` class. Ideally, we'd reuse more of those! That way, all the existing stuff around markers like tooltips, colors and icons works as well. It doesn't right now.

Suggestion, and I don't know if this is possible or reasonable: create `Marker` and `Popup` as usual, add them to a map or feature group as usual, and let the `OMS` class only handle the spiderfying and the click events. So in the `OMS` template overwrite the existing popup click events and bind them to the `oms`.

In short:
- in the current implementation, should markers be added to the map? I guess no, so documentation should be clear on that limitation.
- Or, can we take the marker and popup creation out of this plugin and use the existing marker and popup classes?
- if so, should this plugin actually be a `Layer`? Or should it just be a collector of existing markers?

Apologies if that's a bit much. I'd love to hear what you think about this. And let me know if I understood it correctly! Feel free to be open about what kind of changes you feel okay with.

folium/plugins/overlapping_marker_spiderfier.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+
+    default_js = [
+        (
+            "overlappingmarkerjs",
+            "https://cdnjs.cloudflare.com/ajax/libs/OverlappingMarkerSpiderfier-Leaflet/0.2.6/oms.min.js",
+        )
+    ]
+
+    def __init__(self, options=None, **kwargs):
```
Conengmo, line 76: I'd suggest to add the 4 listed in the docstring as explicit keyword arguments. That's more common in Python than having an `options` argument like in Javascript. This way, users get hints what arguments are available, and their types and default values are also included. 

```suggestion
    def __init__(
            self,
            keep_spiderfied: bool = True,
            nearby_distance: int = 20,
            leg_weight: float = 1.5,
            circle_spiral_switchover: int = 9,
            **kwargs
    ):
```

Then combine these together with the kwargs into `self.options`.

Note by the way that in Folium as a convention we use snake_case for all variables in Python, even if they map to camelCase variables in Javascript.

folium/plugins/overlapping_marker_spiderfier.py
```diff
@@ -0,0 +1,86 @@
...
+        super().__init__()
+        self._name = "OverlappingMarkerSpiderfier"
+        default_options = {
+            "keepSpiderfied": True,
+            "nearbyDistance": 20,
+            "legWeight": 1.5,
+        }
+        if options:
+            default_options.update(options)
+        self.options = parse_options(**default_options, **kwargs)
```
Conengmo, line 86: Then this becomes:

```suggestion
        self.options = parse_options(
            keep_spiderfied=keep_spiderfied,
            nearby_distance=nearby_distance,
            leg_weight=leg_weight,
            circle_spiral_switchover=circle_spiral_switchover,
            **kwargs
        )
```

I should add that your approach works perfectly fine too! This is just more in line with other classes in Folium.

folium/plugins/overlapping_marker_spiderfier.py
```diff
@@ -0,0 +1,86 @@
...
+                    {{ this._parent.get_name() }}.closePopup();
+                });
+
+                {{ this._parent.get_name() }}.eachLayer(function(layer) {
+                    if (
+                        layer instanceof L.Marker
+                    ) {
+                        oms.addMarker(layer);
+                    }
+                });
```
Conengmo, line 59: This works great when users add their markers directly to the map object. But it's quite common to add markers to a FeatureGroup, then add the FeatureGroup to the map. I think we should support that use case.

This made me think whether it's maybe better to explicitly pass the markers they want to have spiderified like before. That way, they can group them in FeatureGroups however they wish. Downside is it doesn't work automatically like here. 

Like for example

```suggestion
                {%- for marker in this.markers %}
                    oms.addMarker({{ marker.get_name() }});
                {%- endfor %}
```

Alternatively, expand this algorithm to recursively go through all layers. For example something like

```
let markers = [];

function findMarkers(layer) {
    if (layer instanceof L.Marker) {
        markers.push(layer);
    } else if (layer instanceof L.LayerGroup) {
        layer.eachLayer(findMarkers); // Recursively check nested layers
    }
}

map.eachLayer(findMarkers);
```

On the one hand making things easier for users is good. On the other hand, it's more code and functionality that can have bugs or can break.

What do you think?

Conengmo (changes requested):
Great work @swtormy, thanks so much for making these changes. I love how the code became simpler but everything, like the click events, still works.

I have basically two comments still. One about how we deal with class arguments in Folium that I should have mentioned the previous review, sorry about that. The other about a new missing use case I'd like to hear your opinion on.

I think with these out of the way we can get this merged!

folium/plugins/overlapping_marker_spiderfier.py
swtormy: Yes, it looks so much better that way!

swtormy:
I decided to try a recursive approach to finding markers, but not in the js template. In doing so, I wanted to keep the approach of minimal user interaction with the plugin so that the user doesn't have to add each token separately.

I tried modifying add_to by collecting all markers in it.

```python3
def add_to(
        self, parent: Element, name: Optional[str] = None, index: Optional[int] = None
    ) -> Element:
      self._parent = parent
      self.markers = self._get_all_markers(parent)
      super().add_to(parent, name=name, index=index)

 def _get_all_markers(self, element: Element) -> list:
      markers = []
      for child in element._children.values():
          if isinstance(child, Marker):
              markers.append(child)
          elif hasattr(child, "_children"):
              markers.extend(self._get_all_markers(child))
      return markers
```

In recursion I am looking for tokens within all children, it seems that this approach should help to find tokens not only in FeatureGroup but also in FeatureGroupSubGroup. 

I completely agree that adding more functionality increases the code and therefore the possibility of getting bugs. Therefore, if you still think that there is no need to complicate things here, I will gladly remove the recursive traversal and leave only manual addition of markers.

folium/plugins/overlapping_marker_spiderfier.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Parameters
+    ----------
+    keep_spiderfied : bool, default True
+        If true, markers stay spiderfied after clicking.
+    nearby_distance : int, default 20
+        Pixels away from a marker that is considered overlapping.
+    leg_weight : float, default 1.5
+        Weight of the spider legs.
+    circle_spiral_switchover : int, optional
```
Conengmo, line 29:
```suggestion
    circle_spiral_switchover : int, default 9
```

Conengmo (approved):
Lovely! This works great, good idea to do it this way. I have one tiny comment about a default value in the docstring that I'll merge myself, after that if there's nothing more from your end I'll go ahead and merge this one!

swtormy:
@hansthen, @Conengmo thank you for reviewing and approving the changes! 👍

I’ve tested the plugin locally, including Selenium tests, and everything passed successfully. However, if I understand correctly, there is an issue with the Selenium environment for python3.9 and 3.13 in the CI pipeline that doesn't bother with the plugin. Maybe, it might be resolved by simply restarting the failing runner? Or is the problem in my code?

Conengmo:
I’ve restarted the tests but they still fail… Ill take a closer look tomorrow.

swtormy:
> I’ve restarted the tests but they still fail… Ill take a closer look tomorrow.

Ok, I tried to get Micromamba up on my windows, with an Ubuntu container to run selenium tests on python 3.9 and 3.13. It took longer than I thought. I'll take a closer look tomorrow as well.

I want to try to rule out my tests first, something tells me that it might be because of them, even though there is no clear indication of them in the logs.

hansthen:
I also have problems running the tests. It would be cool if there were some instructions on how to setup the tests to run locally as part of the developer instructions.

Conengmo:
@hansthen theres this list in the contributors guide: https://github.com/python-visualization/folium/blob/main/.github/CONTRIBUTING.md#contributing-code. Is there anything missing there? If I’m not mistaken Selenium downloads drivers itself these days.

swtormy:
> @hansthen theres this list in the contributors guide: https://github.com/python-visualization/folium/blob/main/.github/CONTRIBUTING.md#contributing-code. Is there anything missing there? If I’m not mistaken Selenium downloads drivers itself these days.

Perhaps @hansthen means that this instruction doesn't have information on how to model python 3.9 and 3.13 in ubuntu if the user is using other OS. 

Perhaps I'm missing something, but this instruction describes the process in which the selenium driver is loaded for the OS the user is running on. In my case it is windows and macos. But on github the test actions are performed in ubuntu.

I haven't yet had time to recreate the conditions from `Selenium Tests` locally, it took longer than I thought.

Conengmo:
I reran the same test on the main branch and it fails there as well, so it's not because of your change: https://github.com/python-visualization/folium/actions/runs/11880066659/job/33603469550

Conengmo:
> Perhaps @hansthen means that this instruction doesn't have information on how to model python 3.9 and 3.13 in ubuntu if the user is using other OS.

That makes sense, but that goes a bit to far for a contributor guide IMO to go into containerizing different OS'es :)

Conengmo:
I made a temporary fix for the Selenium test failure in https://github.com/python-visualization/folium/pull/2034. Could one of you review it?

I'll go ahead and merge this PR since the test failure is unrelated.

Conengmo:
Thanks for your excellent work on this @swtormy!

PR #2036, opened 2024-11-27
Add OverlappingMarkerSpiderfier to plugins table

Conengmo:
I was looking at the added documentation for the new OverlappingMarkerSpiderfier plugin and noticed it wasn't added to the table on the plugins page. See https://python-visualization.github.io/folium/dev/user_guide/plugins.html. Add it in this PR.

Perhaps the reason why it wasn't there is because the contributing guideline is not explicit enough on this topic. So update that guide as well.


Conengmo:
@swtormy do you maybe want to take a look, see if you agree with the line for your plugin, and if the updated guide is clear?

swtormy (approved):

Conengmo:
Thank you!

Issue #2043, opened 2024-12-10
Invalid Zoom for TileLayer in version 0.19.0


pmd84:
I'm using a tiling endpoint based of TiTiler to serve map tiles on Folium. Since the upgrade last night, the tiles no longer render, and my tiling endpoint gives me this error: "Invalid ZOOM level 0", and the tile request is showing as "tiles/WebMercatorQuad/0/0/0" 

```
geom = shape(AOI)
centroid = geom.centroid

m = folium.Map(location=[centroid.y, centroid.x], zoom_start=12)

# Initialize Tile Layer and add to map
TileLayer(
    tiles=tilejson["tiles"][0],
    opacity=1,
    min_zoom=8,
    max_zoom=14,
    attr="Sentinel-2 COG",
).add_to(m)

folium.GeoJson(AOI).add_to(m)

m

```

Typically, the folium map generates tile at whatever x, y, z are visible on the map. Only this single 0,0,0 tile is being requested.

Environment is in a Databricks notebook with Folium 0.19.0

lux-orc:
I have got a similar issue, the following is the reproducible codes that people can test on folium 0.18.0 and 0.19.0.

```python
import folium

# Esri_WorldTopoMap
arcgis = 'https://server.arcgisonline.com/ArcGIS/rest/services'
esri = {'Esri_WorldTopoMap': 'World_Topo_Map'}
tile = 'MapServer/tile/{z}/{y}/{x}'
tile_esri = dict(zip(esri.keys(), ['/'.join([arcgis, v, tile]) for v in esri.values()]))

# GOOGLE online map
tile_google = {
    'Maps': 'https://mt1.google.com/vt/lyrs=r&x={x}&y={y}&z={z}',
    'Satellite': 'http://www.google.cn/maps/vt?lyrs=s@189&gl=cn&x={x}&y={y}&z={z}',
    'Roads': 'https://mt1.google.com/vt/lyrs=h&x={x}&y={y}&z={z}',
}

m = folium.Map(
    location=(-45.54, 169.74),
    tiles=None,
    zoom_start=9,
    zoom_control=False,
    attributionControl=False,
)

# Add 'Esri_WorldTopoMap' as 'Esri_WorldTopoMap'
folium.raster_layers.TileLayer(
    tiles=tile_esri['Esri_WorldTopoMap'],
    attr='Esri_WorldTopoMap',
    name='World Topographic Map (ESRI)',
[...]

PR #2044, opened 2024-12-11
TileLayer: max_native_zoom cannot be null

Conengmo:
Closes https://github.com/python-visualization/folium/issues/2043.

In https://github.com/python-visualization/folium/pull/2029 we changed the way we parse options. Before, in `parse_options`, None values would be discarded. In the `tojavascript` filter this doesn't happen. This leads to different outcomes, and issues like some tile layers not rendering.

I first looked at removing all None values like before, but that seems heavy handed and might lead to other issues. Instead, do a targeted fix that just makes sure `max_native_zoom` always has an integer value. If it's not available, we can set it to max_zoom, which has the same effect as not including it at all.

### DIFF
--- a/folium/raster_layers.py
+++ b/folium/raster_layers.py
@@ -122,7 +122,6 @@ class TileLayer(Layer):
             attr = attr if attr else tiles.html_attribution  # type: ignore
             min_zoom = min_zoom or tiles.get("min_zoom")
             max_zoom = max_zoom or tiles.get("max_zoom")
-            max_native_zoom = max_native_zoom or tiles.get("max_zoom")
             subdomains = tiles.get("subdomains", subdomains)
             if name is None:
                 name = tiles.name.replace(".", "").lower()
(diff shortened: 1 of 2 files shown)


Conengmo:
@hansthen could you maybe take a look, see if you agree with this approach?

hansthen:
I agree with your targeted approach for now. 

I also feel that filtering `None` may cause other issues, because `None` or `null` is simply valid json. It felt more natural to do this during `parse_options`, because that code was scoped more limited.

Just to think out loud about possible alternatives.

The root cause is that in `Folium` arguments are statically typed, whereas in `Leaflet` they are not. Static typing has its advantages, but it removes the possibility to not pass an argument if we use `None` as a sentinel value. 

Static typing cannot distinguish between:
`fun(arg=None)` and `fun()`. 

We could use a separate default argument as a sentinel value. E.g.
```
# define a sentinel
unset=object()

class ABC:
   def __init__(self, arg1=unset, arg2=unset):
```   
We could then create a new function `parse_options` that will discard `unset` values.  However, that is way more work and I am not sure it is worth it. Probably be the best would be to continue as is and simply ensure all `__init__` methods have sensible default values that agree with their Leaflet counterparts.

hansthen (approved):

Issue #2047, opened 2024-12-12
Popup sticky stays True in version>=0.19.0


Itaxhix:
Not much to say here, 0.18.0 and 0.17.0 working just fine (for me)

Same thing happens when I jump in between versions in [official user guide](https://python-visualization.github.io/folium/dev/user_guide/ui_elements/popups.html)

Opened HTML in browser [tested in both chrome and firefox on two computers]

Starting to think that this is intended(?)

Conengmo:
Thanks for the bug report @Itaxhix, very useful! We'll get a fix out asap.

PR #2048, opened 2024-12-12
fix popup sticky options

Conengmo:
Fix https://github.com/python-visualization/folium/issues/2047

This is a similar issue to https://github.com/python-visualization/folium/issues/2043, happening after the v0.19.0 release, caused by changes in https://github.com/python-visualization/folium/pull/2029/.

Fix it by using proper boolean values for these options on Popup, not None values.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -483,8 +483,8 @@ class Popup(Element):
         self.lazy = lazy
         self.options = dict(
             max_width=max_width,
-            autoClose=False if show or sticky else None,
-            closeOnClick=False if sticky else None,
+            autoClose=not (show or sticky),
+            closeOnClick=not sticky,
             **kwargs,
         )
 
--- a/tests/test_map.py
+++ b/tests/test_map.py
@@ -134,7 +134,7 @@ def test_popup_show():
     rendered = popup._template.render(this=popup, kwargs={})
     expected = """
     var {popup_name} = L.popup({{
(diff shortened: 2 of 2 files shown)


Conengmo:
@hansthen I'm not sure if we're going the right way by addressing these individually. Looking at https://github.com/python-visualization/folium/pull/2029/files there are lots of places where `parse_options` was removed, but the default arguments are None, so we are adding null options that were not there before. That means potentially lots more issues like this one.

I think we should go with a simple solution for now, because we have a release out and I'd like to avoid people getting bit by this. Maybe either:
- remove None arguments in `tojavascript`
- revert https://github.com/python-visualization/folium/pull/2029/

For that first solution, do you reckon that will lead to issues? Is there a use for explicit null arguments in Leaflet?

hansthen:
I would prefer to leave `None` values in the `tojavascript` because `null` values are part of the json spec. My feeling is that `tojavascript` is more generic than the original `parse_options`, even though it largely serves the same function. Also, I can't exclude the possibility that Leaflet options can be validly `null`.

I am open to reverting #2029 for now and do a proper release later. However, I did make a simple fix that again removes `None` values from the `options`. You could have a look at that to see if you think it is a viable solution. I still have to fix four tests though.

Conengmo:
Close in favor of https://github.com/python-visualization/folium/pull/2049

PR #2049, opened 2024-12-12
Remove empty values from the options.

hansthen:
@Conengmo can you have a look at this? I still need to update four tests because we now remove the `None` values.

### DIFF
--- a/folium/plugins/semicircle.py
+++ b/folium/plugins/semicircle.py
@@ -1,6 +1,7 @@
 from folium.elements import JSCSSMixin
 from folium.map import Marker
 from folium.template import Template
+from folium.utilities import remove_empty
 from folium.vector_layers import path_options
 
 
@@ -80,6 +81,7 @@ class SemiCircle(JSCSSMixin, Marker):
                 stop_angle=stop_angle,
             )
         )
+        self.options = remove_empty(self.options)
 
         if not (
             (direction is None and arc is None)
--- a/folium/plugins/semicircle.py
+++ b/folium/plugins/semicircle.py
@@ -81,7 +81,7 @@ class SemiCircle(JSCSSMixin, Marker):
                 stop_angle=stop_angle,
             )
         )
-        self.options = remove_empty(self.options)
+        self.options = remove_empty(**self.options)
 
         if not (
             (direction is None and arc is None)
--- a/folium/features.py
+++ b/folium/features.py
@@ -31,6 +31,7 @@ from folium.utilities import (
     javascript_identifier_path_to_array_notation,
     none_max,
     none_min,
+    remove_empty,
     validate_locations,
 )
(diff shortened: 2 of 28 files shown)


Conengmo:
Yes this makes a lot of sense. I also think this is the way to go for this. I assume you put the new function in all places that used to haven parse options?

I’ll do a full review tomorrow morning, just wanted to say I think this is a good approach.

Conengmo (changes requested):
Great work! Good approach and very thorough. I tried finding instances where we had `parse_options` before and are not adding `remove_empty` now, to see if there's a potential issue. Found only one possible mark:

- I think you missed adding `remove_empty` in `plugins/semicircle.py`. That used to have a `parse_options` call and it seems to have values that are None.

After we get this merged, I'll go ahead and make a v0.19.2 release.

hansthen:
> Great work! Good approach and very thorough. I tried finding instances where we had `parse_options` before #2029 and are not adding `remove_empty` now, to see if there's a potential issue. Found only one possible mark:
> 
> * I think you missed adding `remove_empty` in `plugins/semicircle.py`. That used to have a `parse_options` call and it seems to have values that are None.
> 
> After we get this merged, I'll go ahead and make a v0.19.2 release.

Thanks for the compliments. I felt a tad guilty for breaking the release :-(.

folium/plugins/semicircle.py
```diff
@@ -80,6 +81,7 @@ def __init__(
                 stop_angle=stop_angle,
             )
         )
+        self.options = remove_empty(self.options)
```
Conengmo, line 84:
```suggestion
        self.options = remove_empty(**self.options)
```

Conengmo (approved):
> I felt a tad guilty for breaking the release :-(.

No worries, those things happen :) I developed a bit of a paranoia about big changes for this reason. But I don't want development to stop because something could go wrong!

That said, a more complete test suite would help here. But doing integration tests in Selenium is much work and quite fragile. Maybe something to think about how we could set that up properly.

Conengmo:
@hansthen do you want to take a final look and merge? If you can't get to it today, I'll merge this one in a bit. I'll make a release afterwards.

hansthen:
> > I felt a tad guilty for breaking the release :-(.
> 
> No worries, those things happen :) I developed a bit of a paranoia about big changes for this reason. But I don't want development to stop because something could go wrong!
> 
> That said, a more complete test suite would help here. But doing integration tests in Selenium is much work and quite fragile. Maybe something to think about how we could set that up properly.

`streamlit-folium` uses `playwright`. I found it quite intuitive. It has a recording mode and a visual playback mode, which makes it easy to debug failing tests. It is not perfect, I found that tests can break randomly due to changes in underlying components, but rather easy to fix.

PR #2053, opened 2024-12-14
multiple markers with one icon

Conengmo:
This is an exploration of an idea.

Allow using one Icon object for multiple markers by:
- separating the icon Javascript object creation from setting it on the marker
  - by having a method that allows registering a marker to an icon
- making sure setting it on the markers happens after all markers are created

Test it out with:

```
m = Map((0, 0))
icon = Icon(icon='music')
Marker((0, 0), icon=icon).add_to(m)
Marker((0, 1), icon=icon).add_to(m)
```

We could extend this to all marker and icon like classes. This does not apply to popups and tooltips, since those are not exclusively used markers, which icons are.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -271,7 +271,6 @@ class Icon(MacroElement):
             var {{ this.get_name() }} = L.AwesomeMarkers.icon(
                 {{ this.options|tojavascript }}
             );
-            {{ this._parent.get_name() }}.setIcon({{ this.get_name() }});
         {% endmacro %}
         """
     )
@@ -321,6 +320,24 @@ class Icon(MacroElement):
             extra_classes=f"fa-rotate-{angle}",
             **kwargs,
         )
+
+    class SetIcon(MacroElement):
+        """Set the icon of a marker after both are created."""
+        _template = Template("""
+            {% macro script(this, kwargs) %}
+                {{ this.marker.get_name() }}.setIcon({{ this.icon.get_name() }});
+            {% endmacro %}
+        """)
+
+        def __init__(self, marker: 'Marker', icon: 'Icon'):
+            super().__init__()
+            self._name = "SetIcon"
+            self.marker = marker
+            self.icon = icon
+
+    def register_marker(self, marker: "Marker") -> None:
+        """Register a marker to use this icon, so one icon can have multiple markers."""
+        self.add_child(Icon.SetIcon(marker=marker, icon=self))
 
 
 class Marker(MacroElement):
@@ -383,7 +400,11 @@ class Marker(MacroElement):
         self.options = remove_empty(
             draggable=draggable or None, autoPan=draggable or None, **kwargs
         )
+        self.icon: Optional[Icon] = None
         if icon is not None:
+            # here we make sure the icon is rendered after the last marker that needs it
+            if icon._parent is not None:
+                del icon._parent._children[icon.get_name()]
             self.add_child(icon)
             self.icon = icon
         if popup is not None:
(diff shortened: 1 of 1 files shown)


folium/map.py
```diff
@@ -322,6 +321,24 @@ def __init__(
...
+            {% endmacro %}
+        """)
+
+        def __init__(self, marker: 'Marker', icon: 'Icon'):
+            super().__init__()
+            self._name = "SetIcon"
+            self.marker = marker
+            self.icon = icon
+
+    def register_marker(self, marker: "Marker") -> None:
```
hansthen, line 338: To me it would feel more natural to have a `setIcon` method on `Marker` (which does the same thing).

folium/map.py
```diff
@@ -383,7 +400,11 @@ def __init__(
         self.options = remove_empty(
             draggable=draggable or None, autoPan=draggable or None, **kwargs
         )
+        self.icon: Optional[Icon] = None
         if icon is not None:
+            # here we make sure the icon is rendered after the last marker that needs it
```
hansthen, line 405: It feels more natural to make SetIcon a child of Marker and add the Icon to the map instead of to the Marker.

folium/map.py
```diff
@@ -383,7 +400,11 @@ def __init__(
...
+            if icon._parent is not None:
+                del icon._parent._children[icon.get_name()]
             self.add_child(icon)
```
hansthen, line 408: In my proposal, the Icon could be added to the map (or to the parent of the Marker). We can keep this line to avoid breaking backwards compatibility. I am not sure if `add_child` is idempotent. In general I think for `Folium` at least it should be.

folium/map.py
```diff
@@ -383,7 +400,11 @@ def __init__(
...
             self.icon = icon
```
hansthen, line 409: If you follow my setIcon proposal, we could call it here.

hansthen:
> This is an exploration of an idea.
This is a **great** idea. I really like this proposal. It is very pragmatic and solves the immediate reported issue. In the review are a few suggestions, mainly to add the generated icon to the map, instead of the Marker. I think it will make the resulting code simpler.

User code would be like this:

```
m = Map((0, 0))
icon = Icon(icon='music').add_to(m)
Marker((0, 0), icon=icon).add_to(m)
Marker((0, 1), icon=icon).add_to(m)
```

Conengmo:
Thanks for your review! I'll take a look at it next!

Conengmo:
I've been messing around with your suggestions, but I haven't figured it out so far. When I was working on this two weeks ago, I actually began with trying to solve this in the `Marker` class, but couldn't figure it out. Maybe you'd like to play around with your idea a bit, see if you can get it to work?

We have make sure the `setIcon` JS call is always called **after** both the `Icon` and the `Marker` have been initialized in JS. This should be true for multiple cases:

- Creating an Icon, creating a Marker with the icon argument
- Creating a Marker, creating an Icon, adding the icon to the marker with `icon.add_to(marker)`.
- Creating a Marker, creating an Icon, adding the icon to the marker with `marker.add_child(icon)`.

Currently, Branca/Folium don't check if an element was already added to another element. That means we'll get duplicates if multiple markers add the same icon object to their parent/map/figure.

Conengmo:
Alright let's simplify a bit and only support icon reuse when the `Marker(icon=icon)` argument is used.

Conengmo:
It could actually work the same as `Layer`. Where in the `render` method, we add an `Element` to `Figure.script` that combines the marker and icon. If we do that after all children of Marker are rendered, we are guaranteed both icon and marker JS objects exist. Additional component is we have to make sure the icon is rendered only once, we can because `Figure.script` contains key names that we can check in.

Conengmo:
Close in favor of https://github.com/python-visualization/folium/pull/2068

PR #2056, opened 2024-12-16
Add DivIcon and CustomIcon as accepted type of Marker icon

pixelsapphire:
A `DivIcon` works when passed as the `icon` argument of `Marker`, but the argument is annotated as `Optional[Icon]` instead of `Optional[Union[Icon, "DivIcon"]]`.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -373,7 +373,7 @@ class Marker(MacroElement):
         location: Optional[Sequence[float]] = None,
         popup: Union["Popup", str, None] = None,
         tooltip: Union["Tooltip", str, None] = None,
-        icon: Optional[Icon] = None,
+        icon: Optional[Union[Icon, "DivIcon"]] = None,
         draggable: bool = False,
         **kwargs: TypeJsonValue,
     ):
--- a/folium/map.py
+++ b/folium/map.py
@@ -5,7 +5,7 @@ Classes for drawing maps.
 
 import warnings
 from collections import OrderedDict
-from typing import List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, List, Optional, Sequence, Union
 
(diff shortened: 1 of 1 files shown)


pixelsapphire:
> It's missing an import of that class

My bad, should have checked that first.
> I assume this argument also accepts CustomIcon objects. Maybe you can add that as well?

Done and dusted.

hansthen:
@Conengmo Do you know what is the issue with the failing mypy checks? It seems mypy is bothered by some method signature redefinitions between Branca and Folium.

pixelsapphire:
@hansthen This is something I've been wondering about too. Why does the `render` method in Folium classes return `None`, when this method in Branca classes (as in the `Element` class itself) returns `str`?

Conengmo:
I'm trying to reproduce this issue locally...

Conengmo:
Thanks @pixelsapphire! This PS is much appreciated, I hadn't noticed this!

hansthen:
> @hansthen This is something I've been wondering about too. Why does the `render` method in Folium classes return `None`, when this method in Branca classes (and in the `Element` class itself) returns `str`?

I have been banging my head on this for a few days now. The signature redefinition seems to start already in Branca. See `brana.element.Figure` for instance. The children are void rendered (without actually returning anything). Then the `_template` of Figure is rendered. This will render just the `header`, `html` and `script` attributes and return the result.

```
    def render(self, **kwargs):
        """Renders the HTML representation of the element."""
        for name, child in self._children.items():
            child.render(**kwargs)
        return self._template.render(this=self, kwargs=kwargs)
 ``` 
In `branca.element.MacroElement`, the children are also void rendered, but here the `render` method itself also has a void signature.

PR #2057, opened 2024-12-17
unique option for folium.plugin.draw

NaoCoding:
Add `unique` option when developers want their Folium web can only draw one object at a time.
This option will automatically remove previous drawn object when `draw:created`

The reason I write this option is because I was using the `draw` for user to select specific area.
However, the current `draw` will not delete last drawn object.
In some cases, this option may help and improve the user experience for the visitors of the website.
It may also help developers since they don't need to modify the source code (or template) by themselves according to their needs.

Below video capture is the example while activating this option.

https://github.com/user-attachments/assets/a04bd2c8-9a78-495c-b6a7-548bb89d2f8e

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -31,6 +31,10 @@ class Draw(JSCSSMixin, MacroElement):
     on : dict, optional
         Event handlers to attach to the created layer. Pass a mapping from the
         names of the events to their `JsCode` handlers.
+    unique : bool , default False
+        Set if the draw object should only exist one at a time.
+        Automatically delete the last drawn object, ensure the map will only exist one draw object at a time.
+        This option can sometimes increase user experience in specific conditions since they won't need to keep deleting last drawn object.
 
     Examples
     --------
@@ -51,6 +55,7 @@ class Draw(JSCSSMixin, MacroElement):
     ...     """
     ...         )
     ...     },
+            unique=False
     ... ).add_to(m)
 
     For more info please check
@@ -92,6 +97,8 @@ class Draw(JSCSSMixin, MacroElement):
                 });
                 {%- endif %}
 
+                
+
                 {%- for event, handler in this.on.items()   %}
                 layer.on(
                     "{{event}}",
@@ -101,6 +108,11 @@ class Draw(JSCSSMixin, MacroElement):
                 drawnItems_{{ this.get_name() }}.addLayer(layer);
             });
             {{ this._parent.get_name() }}.on('draw:created', function(e) {
+                {%- if this.unique %}
+                for(const k in drawnItems_{{ this.get_name() }}._layers){
+                    drawnItems_{{ this.get_name() }}.removeLayer(drawnItems_{{ this.get_name() }}._layers[k])
+                }
+                {%- endif %}
                 drawnItems_{{ this.get_name() }}.addLayer(e.layer);
             });
 
@@ -144,6 +156,7 @@ class Draw(JSCSSMixin, MacroElement):
         draw_options=None,
         edit_options=None,
         on=None,
+        unique=False,
     ):
         super().__init__()
         self._name = "DrawControl"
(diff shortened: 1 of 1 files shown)


hansthen (commented):
I am a bit hesitant to accept this pull request. 

The changes made inside the plugin could also be implemented in client code, by adding an event handler. I can try to come up with an example on how to do this.

NaoCoding:
I totally understand your hesitate, and I completely agree with your subsequent decision.

The main reason I send this pull request was I consider that if an option can fulfill the whole function easily, 
it may help or make the develop progress faster and easier for people who is using Folium for their first time in the future.

hansthen:
> I totally understand your hesitate, and I completely agree with your subsequent decision.
> 
I have not made up my mind yet. It all depends on how easy or difficult this is in client code. If I can implement this in client code easily and cleanly, we can add it to the documentation, which also will help new users.

martinfleis:
@hansthen Event handler requires writing JavaScript no? IMHO simce the point of folium is to generate leaflet maps using Python, anything that removes reliance on typing JS is in scope.

hansthen:
@martinfleis I can certainly see the allure of not having to write any Javascript :-) However, there is a potentially unlimited number of use cases that could be included in Folium for that reason. Since Folium is maintained by volunteers we also try to keep the codebase small. So far my (not very well thought out) rule of thumb has been that we keep Folium a thin wrapper around Leaflet. That implies that we do not add any functionality that is not already in Leaflet, especially if the functionality can also be achieved using existing means.

With this in mind what would you recommend to me? Should I change my rule of thumb? Or just make an exception in this case? I am really of two minds regarding this, since in this case the JsCode needs to refer to other Elements, which makes it a rather ugly solution.

hansthen:
@NaoCoding Can you have a look at this example? I think it does the same as your Pull Request, but using only client code. As said, I am not really happy with it, because of the string interpolation. 

```
      import streamlit as st
      m = folium.Map(
          location=[39.949610, -75.150282],
          zoom_start=5,
      )
      items = folium.FeatureGroup()
      items.add_to(m)
  
      draw_plugin = Draw(
          export=False,
          feature_group=items,
          show_geometry_on_click=False
      )  
      draw_plugin.add_to(m)
  
      # We need a dict here, because we cannot used named
      # parameters for event names with colons in them.
      handlers = {
          "draw:created":
              JsCode("""
                  (e) => {
                      var features = %s;
                      for(const k in features._layers){
                          var layer = features._layers[k];
                          if(layer !== e.layer) {
                              drawnItems.removeLayer(layer);
                          }
                      }
                  }
              """ % items.get_name()),
      }
      m.on(**handlers)
```

NaoCoding:
> @NaoCoding Can you have a look at this example? I think it does the same as your Pull Request, but using only client code. As said, I am not really happy with it, because of the string interpolation.
> 
> ```
> import streamlit as st
> from folium.plugins import Draw    
> m = folium.Map(
>     location=[39.949610, -75.150282],
>     zoom_start=5,
> )
> items = folium.FeatureGroup()
> items.add_to(m)
>   
> draw_plugin = Draw(
>     export=False,
>     feature_group=items,
>     show_geometry_on_click=False
> )  
> draw_plugin.add_to(m)
>   
> # We need a dict here, because we cannot used named
> # parameters for event names with colons in them.
> handlers = {
>     "draw:created":
>         JsCode("""
>             (e) => {
>                 var features = %s;
>                 for(const k in features._layers){
>                     var layer = features._layers[k];
>                     if(layer !== e.layer) {
>                         drawnItems.removeLayer(layer);
>                     }
>                 }
>             }
>         """ % items.get_name()),
> }
> m.on(**handlers)
> ```

Yes, I think the code do the same thing.

martinfleis:
> So far my (not very well thought out) rule of thumb has been that we keep Folium a thin wrapper around Leaflet.

yeah , that is fair.

hansthen:
I am closing this Pull Request. The requested functionality is not part of Leaflet and can be implemented using the existing mechanism for event handlers.

PR #2060, opened 2024-12-29
Fix type hints after adding Branca type checking

Conengmo:
In the recent Branca release we enabled type checking. This leads to issues when type checking Folium, since it relies on Branca. This PR addresses these issues.

Each commit in this PR contains a single fix, so you can walk through the commits to see what changes.

## Render function
Note about the `render` function: I removed the return type, which effectively disables type checking. This is unfortunately necessary. In Branca we specify this function should return a string, but that's not what's done in `MacroElement` and so basically everything in Folium. We should update the type in Branca, do a release, then add the types back in Folium. This is a lot of trouble so maybe not worth the effort.

## Bounds
The return type of bounds functions should allow for None values to be consistent. Even if we're sure a function returns floats not None, we still need to type it as optional. That's because the values are returned in a list, which is mutable.

Another thing is that some functions accept a `bounds` argument, which can be a list or tuple, but must have non-None values. Those types are distinct from bounds as return values, which should be list of list of optional float. Split the two in distinct types.

## VegaLite
In practice, a `VegaLite` only works if it's added to a `Figure`, `Div` or `Popup`. We only have examples of that in our documentation, but it's not explicitly listed. Update the code to raise a meaningful error when this element is added to the wrong parent. This also then satisfies Mypy.

### DIFF
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -12,7 +12,7 @@ class JSCSSMixin(Element):
     default_js: List[Tuple[str, str]] = []
     default_css: List[Tuple[str, str]] = []
 
-    def render(self, **kwargs) -> None:
+    def render(self, **kwargs):
         figure = self.get_root()
         assert isinstance(
             figure, Figure
--- a/folium/features.py
+++ b/folium/features.py
@@ -165,7 +165,7 @@ class Vega(JSCSSMixin, Element):
         self.top = _parse_size(top)
         self.position = position
 
-    def render(self, **kwargs) -> None:
+    def render(self, **kwargs):
         """Renders the HTML representation of the element."""
         super().render(**kwargs)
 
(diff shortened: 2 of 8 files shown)


folium/elements.py
hansthen, line 1: I'd prefer to rename these methods, since they really do something different than `branca.Element.render`. See the discussion I started on the signature redefinition.

hansthen: I do not understand all the typing issues fully, but in general it looks good. I don't mind going ahead with this as is to fix the immediate issues. We can discuss my proposal to rename `render` from MacroElement and its children at a later moment. Whatever you think is best for now.

hansthen (approved):
I have some suggested changes but I am also fine to do this as is for now.

Conengmo:
Thanks for the review!

You're right about `render`: the type issues surface the larger issue. I'd prefer to get this fix out so the type checks don't fail for our user. Then we can look at how to address this more structurally next.

PR #2066, opened 2025-01-01
Cleanup make MacroElements from Elements

hansthen:
This change involves several classes that require being added to a Figure. Since these classes follow the `render` semantics of a MacroElement it makes more sense to make them also inherit from MacroElement.

Excluded from this change is actually making use of the template mechanics of the MacroElement.

### DIFF
--- a/folium/elements.py
+++ b/folium/elements.py
@@ -1,6 +1,12 @@
 from typing import List, Tuple
 
-from branca.element import CssLink, Figure, JavascriptLink, MacroElement
+from branca.element import (
+    CssLink,
+    Element,  # NoQA: F401  needed as a reexport
+    Figure,
+    JavascriptLink,
+    MacroElement,
+)
 
 from folium.template import Template
 from folium.utilities import JsCode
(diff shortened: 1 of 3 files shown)


Conengmo (changes requested):
Nice change! It's interesting how these classes are not exactly MacroElements either, since they don't use macros. But at least we get the return type of the `render` method correct.

I added two comments about the todo comments, after addressing those this should be good to merge.

hansthen:
Before you merge this, do you have some example usage of the `Vega` and `VegaLite` classes? I wanted to run some tests for myself, but I could not find anything in the documentation (or in the doc-string).

Conengmo:
We have examples for both classes in the Popup documentation: https://python-visualization.github.io/folium/latest/user_guide/ui_elements/popups.html#Vega-chart-in-popup

Conengmo (approved):
Code looks good! If you manually check these classes still work as expected, go ahead and merge it 👍 I'm less available next few days.

PR #2064, opened 2025-01-01
Update LICENSE.txt, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

PR #2067, opened 2025-01-01
Update copyright in LICENSE.txt

Conengmo:
See the discussion here: https://github.com/python-visualization/folium/pull/2064

Update the copyright statement to refer to all contributors to the project. Add a hyphen to indicate ongoing development. We don't want to have to update this file every year. Add a link to the list of contributors.

I'm open to other suggestions! Don't care that much to be honest since it's MIT and will stay that way, just don't want this to be a yearly ongoing issue.


martinfleis (approved):
I think this is ideal.

PR #2068, opened 2025-01-01
Multiple markers with one icon

Conengmo:
Based on the comments on #2053 I made a new version that sets the icon from the markers perspective. This is also similar to how `Layer` works.

- Check if a marker has an icon added as a child, if so, add another element that sets the icon on the marker.

Tested it with this case:

```
    m = Map((44, -93), tiles="cartodb dark matter", zoom_start=8)

    icon = Icon(icon='music')
    markers = [
        Marker((52, 5), icon=icon, popup="0"),
        Marker((52, 5.1), icon=icon, popup="1"),
        Marker((52.1, 5)),
        Marker((52.1, 5.1)),
    ]
    Icon(icon="cloud").add_to(markers[-1])
    fg = FeatureGroup(name='group').add_to(m)
    for marker in markers:
        marker.add_to(fg)

    FitOverlays(padding=10).add_to(m)
```

If this approach looks good, I'll generate some test cases.

Not sure how much of a breaking change this is. From a direct user perspective, I don't think there's an issue. Not sure about packages that use Folium internals though.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -373,13 +373,16 @@ class Marker(MacroElement):
 
     class SetIcon(MacroElement):
         """Set the icon of a marker after both are created."""
-        _template = Template("""
+
+        _template = Template(
+            """
             {% macro script(this, kwargs) %}
                 {{ this.marker.get_name() }}.setIcon({{ this.icon.get_name() }});
             {% endmacro %}
-        """)
+        """
+        )
 
-        def __init__(self, marker: 'Marker', icon: 'Icon'):
+        def __init__(self, marker: "Marker", icon: "Icon"):
             super().__init__()
             self._name = "SetIcon"
             self.marker = marker
--- a/folium/map.py
+++ b/folium/map.py
@@ -372,6 +372,20 @@ class Marker(MacroElement):
         """
     )
 
+    class SetIcon(MacroElement):
+        """Set the icon of a marker after both are created."""
+        _template = Template("""
+            {% macro script(this, kwargs) %}
+                {{ this.marker.get_name() }}.setIcon({{ this.icon.get_name() }});
+            {% endmacro %}
+        """)
+
+        def __init__(self, marker: 'Marker', icon: 'Icon'):
+            super().__init__()
+            self._name = "SetIcon"
+            self.marker = marker
+            self.icon = icon
+
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


hansthen:
Looks good to me.

folium/map.py
```diff
@@ -402,9 +401,7 @@ def __init__(
             draggable=draggable or None, autoPan=draggable or None, **kwargs
         )
         if icon is not None:
-            # this makes sure it is added only once
-            self._parent.add_child(icon, name=icon.get_name(), index=0)
-            self.icon = icon
+            self.add_child(icon)
```
hansthen, line 407: Why did the above code not work?

folium/map.py
```diff
@@ -421,12 +418,13 @@ def _get_self_bounds(self) -> TypeBoundsReturn:
         return cast(TypeBoundsReturn, [self.location, self.location])
 
     def render(self):
-        if self.icon:
-            self.add_child(self.SetIcon(marker=self, icon=self.icon))
         if self.location is None:
             raise ValueError(
                 f"{self._name} location must be assigned when added directly to map."
             )
+        for child in list(self._children.values()):
```
hansthen, line 430: This looks strange to me. A marker can have only one `Icon`, correct?  Why do we need to look through all the children? Would it not be easier to use `self.icon`?

hansthen (approved):

hansthen:
In general I have no objections, but I do have two comments that you may want to look at.

folium/map.py
Conengmo: It worked I think, but it's not necessary. `Figure.script` is an OrderedDict, so we are already guaranteed that multiple renders of the same Icon object will result in only one entry in `Figure.script`.

folium/map.py
Conengmo: > Would it not be easier to use self.icon?

That would not work for the case where the icon is added using `marker.add_child(icon)` or `icon.add_to(marker)`.

Conengmo:
Thanks for the review! I replied to your comments, let me know if you have further questions or suggestions!

martinfleis:
> Not sure how much of a breaking change this is. From a direct user perspective, I don't think there's an issue. Not sure about packages that use Folium internals though.

#2084 is caused by this as per commit history. So this is indeed a breaking change, at least for geopandas.

Issue #1942, opened 2024-04-30
Add example pattern for pulling JsCode snippets from external javascript files


thomasegriffith:
**Is your feature request related to a problem? Please describe.**
I have found myself embedding a lot of javascript code via `JsCode` when using the `Realtime` plugin.  This gets messy for a few reasons:
- unit testing the javascript code is not possible
- adding leaflet plugins can result in a huge amount of javascript code embedded inside python
- and the code is just not as modular as it could be

I'm wondering if there's a recommended way to solve this.

**Describe the solution you'd like**
A solution would allow one to reference specific functions from external javascript files within a `JsCode` object.

It's not clear to me what would be better: whether to do this from within `folium` (using the API) or to use some external library to populate the string of javascript code that's passed to `JsCode` based on some javascript file and function.  In either case, having an example in the `examples` folder would be great. 

**Describe alternatives you've considered**
If nothing is changed about the `folium` API, this could just be done external to `folium`.  As in, another library interprets a javascript file, and given a function name, the function's definition is returned as a string.  Is this preferable to building the functionality into `folium`?   If so, does anybody know of an existing library that can already do this?

**Additional context**
n/a

**Implementation**
n/a

hansthen:
Very interesting idea. I have noticed the same drawbacks of having javascript as strings inside python. (And I would add there is no editor support or linting of javascript from inside python). 

I don't know of a library that does exactly this, but there are javascript parsers available in python. E.g. https://github.com/PiotrDabkowski/pyjsparser. And there is pythonmonkey. This serves a totally different purpose, but also does javascript parsing. 

I think with either we could hack up a PoC that does what you want. Do you have an example of how you would want to use this library to be?

thomasegriffith:
Agreed: lack of editor and linting support are also painful.

> Do you have an example of how you would want to use this library to be?

Maybe something like:

pointToLayer.js
```
function pointToLayer(f, latlng) {
    ...
}
```

plugin.js
```
function somePlugin() {
    ...
}
```

folium.py
```
point_to_layer_function = ExternalJsFunction(file="/path/to/foo.js", function="pointToLayer")
some_plugin_function = ExternalJsFunction(file="/path/to/plugin.js", function="somePlugin")

# A list of functions is passed to `point_to_layer`
realtime_plugin = folium.plugins.Realtime(..., point_to_layer=[point_to_layer_function, some_plugin_function])
```

hansthen:
[...]

PR #2081, opened 2025-01-16
Use the JsCode object from the js_loader package

hansthen:
This PR is based on a discussion in #1942. 

In that issue, the user wanted to load `JsCode` objects from a javascript file, instead of embedding javascript code as strings in a python file. The advantages are that users can use linters, syntax highlighting and autocomplete on the JavaScript code.

The `js_loader` package allows this. It provides an import hook that will parse Common JS source files and exposes exported functions as `JsCode` objects. It allows users to do this:

```
from js_loader import install_js_loader
install_js_loader()

from path_to_js_module import js_module
  
js_module.my_event_handler # can now be used as an event handler in a Folium map.
```

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -36,8 +36,8 @@ jobs:
           --file requirements.txt
           --file requirements-dev.txt
 
-    - name: Install folium from source
-      run: python -m pip install js_loader
+    - name: Install js_loader
+      run: python -m pip install js_loader pythonmonkey
 
     - name: Install folium from source
       run: python -m pip install -e . --no-deps --force-reinstall
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -37,6 +37,9 @@ jobs:
           --file requirements-dev.txt
 
     - name: Install folium from source
+      run: python -m pip install js_loader
+
+    - name: Install folium from source
       run: python -m pip install -e . --no-deps --force-reinstall
 
     - name: Build documentation
--- a/.github/workflows/test_code.yml
+++ b/.github/workflows/test_code.yml
@@ -33,5 +33,8 @@ jobs:
     - name: Install folium from source
       run: python -m pip install -e . --no-deps --force-reinstall
 
+    - name: Install js_loader
+      run: python -m pip install js_loader
+
     - name: Code tests
       run: python -m pytest -vv --ignore=tests/selenium
--- a/.github/workflows/test_latest_branca.yml
+++ b/.github/workflows/test_latest_branca.yml
@@ -26,6 +26,10 @@ jobs:
       shell: bash -l {0}
       run: python -m pip install -e . --no-deps --force-reinstall
 
+    - name: Install js_loader
+      shell: bash -l {0}
+      run: python -m pip install js_loader
+
     - name: Tests with latest branca
       shell: bash -l {0}
       run: |
--- a/.github/workflows/test_mypy.yml
+++ b/.github/workflows/test_mypy.yml
@@ -27,6 +27,11 @@ jobs:
       run: |
         python -m pip install -e . --no-deps --force-reinstall
 
+    - name: Install js_loader
+      shell: bash -l {0}
+      run: |
+        python -m pip install js_loader
+
     - name: Mypy test
       shell: bash -l {0}
       run: |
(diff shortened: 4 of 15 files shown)


Conengmo (commented):
First question, why not add it to the requirements.txt, but add it 'manually' to the Github Actions workflows and setup.py?

As it is, I don't see the benefit of this PR. We exchange 11 lines of code for a new mandatory dependency. I know you made that library, so it's not that I don't trust you. But seen from Folium, we'd give up control of this part to just you and whoever is in the 'streamlit-community' org, which we don't know.

The difficulty with Branca being a separate library has shown that there's a downside to splitting interwoven functionality off. I don't want to repeat that with `JsCode`.

So having an external library that helps with loading Javascript files? Sounds good! Moving a class that is used everywhere throughout Folium to a separate library? Maybe not a good idea.

hansthen:
> First question, why not add it to the requirements.txt, but add it 'manually' to the Github Actions workflows and setup.py?

The `js_loader` library is not hosted on conda. So adding it to requirements.txt will raise an error when running the tests, because it tries to install js_loader from there.
 
> As it is, I don't see the benefit of this PR. We exchange 11 lines of code for a new mandatory dependency. I know you made that library, so it's not that I don't trust you. But seen from Folium, we'd give up control of this part to just you and whoever is in the 'streamlit-community' org, which we don't know.

Fair enough, but I think it is not so bad in practice. See my comments below.

> The difficulty with Branca being a separate library has shown that there's a downside to splitting interwoven functionality off. I don't want to repeat that with `JsCode`.

`Branca` is much more tightly interwoven with `Folium` than `JsCode` and `Folium`. `JsCode` is not structurally related to `Folium` at all. In fact `Folium` existed just fine until December 2024 without it. `JsCode` is much more comparable to using `xyzservices` or `geopandas`.  If I had written `js_loader` first and only then used the `JsCode` from it in `Folium` would you feel the same?

> So having an external library that helps with loading Javascript files? Sounds good! Moving a class that is used everywhere throughout Folium to a separate library? Maybe not a good idea.

I see your concern about losing control over a key class. (It is also flattering to see how quickly `JsCode` became core `Folium` even though I copied it from the Streamlit world.) However, I also think loading event handlers from javascript files is useful for Folium. Do you have suggestions how we can achieve the second and avoid the drawbacks of the first?

One suggestion from me is that we can make `js_loader` part of the `Folium` world, so you don't lose control. Perhaps you have other suggestions?

Conengmo:
> The js_loader library is not hosted on conda

Ah I understand. I guess we should get it on Conda if we want to adopt it.

> JsCode is much more comparable to using xyzservices or geopandas

That's an interesting comparison. I feel like the contact surface with xyzservices is pretty limited in comparison, and also not expected to change. Geopandas is not even an required dependency. I see JsCode being used in much more places, and also being something we might want to adapt going forward. 

> It is also flattering to see how quickly JsCode became core Folium

Definitely true, it's been a great idea!

> I also think loading event handlers from javascript files is useful for Folium

Agreed. This PR just moves JsCode to another repo, so I'm mainly responding to that. I'd be very interested to instead learn more about how it would look like to support the extended behavior you mention here. Maybe we can discuss some code snippets that achieve that?

If we know what behavior we want to include, we can then decide where we ideally want to place those. Some options:
- include them in the Folium repo
- make a new repo under `python-visualization`
- a repo in an external org
  - since that would be a required dependency, I would have some requirements for that in terms of stability and availability.

hansthen:
I added some documentation to the advanced guide to show someone could load JsCode from a javascript file.

Moving `JsCode` was mainly done to keep the type checker happy. In case our users load `JsCode` from `js_loader` I don't want them to see a mypy error about it. We could just ask our users to ignore type checking in this case. "If Duck Typing is good enough for Guido, it should be good enough for you!"

Alternatively, we could also work around that by using a mypy protocol wherever we now expect a `folium.utilities.JsCode` object as a parameter. I have no experience with protocols, but they look similar to a Java interface (but in Python). That way our users can use js_loader.JsCode since they implement the same protocol as `JsCode`.

hansthen:
@Conengmo I implemented typing using a `Protocol`. Apparently we only need the `js_code` attribute. This way we do not rely on `js_loader` for `JsCode`. I still need it for the tests though. Let me know if you like this better.

Conengmo:
Sorry Hans I'm a bit overwhelmed at the moment, I do want to take a moment to properly understand this, but I don't have the space for it right now. But soon!

hansthen:
> Sorry Hans I'm a bit overwhelmed at the moment, I do want to take a moment to properly understand this, but I don't have the space for it right now. But soon!

No problem sir. I am not particularly attached to this change either. So its perfectly fine with me if you just reject it as not worth the effort.

PR #2083, opened 2025-01-18
Issue 2077 playwright

hansthen:
Use playwright and an example streamlit app for testing Folium maps.


hansthen:
That makes sense. I guess we could try to do something similar with packages such as `geopandas`.

Conengmo:
> I guess we could try to do something similar with packages such as geopandas.

Yes, agreed.

Conengmo:
I'll close this PR then in favor of #2112 and #2113

Issue #2084, opened 2025-01-18
Passing folium.Icon to folium.Marker seems to be broken on main


martinfleis:
GeoPandas CI is failing with Folium main due to the icon passed to the marker no longer being used.

```py
import folium

fc = {'type': 'FeatureCollection',
 'features': [{'id': '0',
   'type': 'Feature',
   'properties': {'name': 'Vatican City'},
   'geometry': {'type': 'Point', 'coordinates': (12.4533865, 41.9032822)},
   'bbox': (12.4533865, 41.9032822, 12.4533865, 41.9032822)}],
 'bbox': (12.4533865, 41.9032822, 12.4533865, 41.9032822)}

m = folium.Map()
marker = folium.Marker(icon=folium.Icon(icon="star"))
folium.GeoJson(
        fc,
        marker=marker,
).add_to(m)
m
```

This creates a marker with a star symbol in folium 0.19.4 but uses default on main.

Conengmo:
Sorry about that Martin! I made a fix that should hopefully solve this issue.

LukasWallisch:
Shouldn't this be fixed in 0.19.5 ?

For me it still is there in this Version 🙃

```
> pip show folium   
Name: folium
Version: 0.19.5
> pip show geopandas      
Name: geopandas
Version: 1.0.1
```

Conengmo:
Hi @LukasWallisch, I tried the example Martin posted in the OP with Folium 0.19.5 and Geopandas 1.0.1 and it works. If it doesn't work for you, could you maybe share a complete reproducible example that leads to your issue?

martinfleis:
Yeah,can confirm that the issue GeoPandas was facing has been fixed.

PR #2086, opened 2025-01-24
Fix icon when passing Marker to GeoJson

Conengmo:
Fixes https://github.com/python-visualization/folium/issues/2084, which is an issue introduced by https://github.com/python-visualization/folium/pull/2068.

We shouldn't have removed `self.icon` in `Marker`. It's not directly visible where it is used, so I added a comment to prevent this from happening again.

### DIFF
--- a/folium/map.py
+++ b/folium/map.py
@@ -404,7 +404,7 @@ class Marker(MacroElement):
             draggable=draggable or None, autoPan=draggable or None, **kwargs
         )
         # this attribute is not used by Marker, but by GeoJson
-        self.icon: Union[Icon, "CustomIcon", "DivIcon", None] = None
+        self.icon = None
         if icon is not None:
             self.add_child(icon)
             self.icon = icon
--- a/folium/map.py
+++ b/folium/map.py
@@ -424,15 +424,12 @@ class Marker(MacroElement):
         return cast(TypeBoundsReturn, [self.location, self.location])
 
     def render(self):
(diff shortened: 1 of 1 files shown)


hansthen (commented):
This still breaks if people use the `add_child` method to set the icon.

folium/map.py
```diff
@@ -403,8 +403,11 @@ def __init__(
         self.options = remove_empty(
             draggable=draggable or None, autoPan=draggable or None, **kwargs
         )
+        # this attribute is not used by Marker, but by GeoJson
+        self.icon: Optional[Icon] = None
```
hansthen: This still fails if people use the `add_child` method to add an icon. I am okay with that, but I think we should be aware of it.

hansthen: Also the `render` method (roughly lines 430-436) was originally rewritten to allow people to use `add_child`. If we go back to icon as a member variable, we should make it consistent throughout the class.

hansthen (changes requested):

martinfleis:
Is this waiting for some changes or are we fine with it as is? I am unable to properly see the consequences here so will let you be the judges but I'd love this to be merged to fix geopandas CI.

hansthen:
> Is this waiting for some changes or are we fine with it as is? I am unable to properly see the consequences here so will let you be the judges but I'd love this to be merged to fix geopandas CI.

Ah, sorry it escaped my attention. Will look at it presently.

PR #2088, opened 2025-01-26
Cleanup MacroElements

hansthen:
During an earlier change, I noted that many `MacroElement` subclasses do not fully use the templating facilities of their parent class.

This PR makes the code more internally consistent by using the template route where possible. I skipped the `DualMap`, `Vega` and `Vegalite` classes because I do not fully understand what is happening there.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -7,7 +7,7 @@ import time
 import webbrowser
 from typing import Any, List, Optional, Sequence, Union
 
-from branca.element import Element, Figure
+from branca.element import Figure
 
 from folium.elements import JSCSSMixin
 from folium.map import Evented, FitBounds, Layer
@@ -60,23 +60,6 @@ _default_css = [
         "https://cdn.jsdelivr.net/gh/python-visualization/folium/folium/templates/leaflet.awesome.rotate.min.css",
     ),  # noqa
 ]
-
-
-class GlobalSwitches(Element):
-    _template = Template(
-        """
-        <script>
-            L_NO_TOUCH = {{ this.no_touch |tojson}};
-            L_DISABLE_3D = {{ this.disable_3d|tojson }};
-        </script>
-    """
-    )
-
-    def __init__(self, no_touch=False, disable_3d=False):
-        super().__init__()
-        self._name = "GlobalSwitches"
-        self.no_touch = no_touch
-        self.disable_3d = disable_3d
 
 
 class Map(JSCSSMixin, Evented):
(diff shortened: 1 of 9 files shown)


folium/folium.py
```diff
@@ -62,23 +62,6 @@
 ]
 
 
-class GlobalSwitches(Element):
```
Conengmo, line 65: this is a potential breaking change for users, so if we want to do this I think we should do it in a separate PR with a motivation why.

folium/map.py
```diff
@@ -454,25 +454,27 @@ class Popup(MacroElement):
...
-            {% endif %}
-        {% endfor %}
-
-        {{ this._parent.get_name() }}.bindPopup({{ this.get_name() }})
-        {% if this.show %}.openPopup(){% endif %};
-
-        {% for name, element in this.script._children.items() %}
-            {{element.render()}}
-        {% endfor %}
+        {% macro script(this, kwargs) %}
```
Conengmo, line 457: I'm not sure but this seems like a change that might break things. I know popups are a bit different from the rest. This is also a change I'd like to consider separately. And that would need proper testing.

folium/plugins/search.py
```diff
@@ -123,17 +122,9 @@ def __init__(
         self.placeholder = placeholder
         self.collapsed = collapsed
         self.options = remove_empty(**kwargs)
+        self.test_params()
```
Conengmo, line 125: not sure if this is the same as testing these at render time. To be safe, I'd keep them at render time, since things may change between initialization and rendering.

Conengmo (changes requested):
Nice that you identified some areas where templates are not used in the default way. I added some comments to places that may require some more attention. If you could limit this PR to just the straight forward cases, it should be easy to merge. Then we can maybe look at the more complicated cases separately or leave them as is.

folium/folium.py
hansthen: Ok restored.

hansthen:
> Nice that you identified some areas where templates are not used in the default way. I added some comments to places that may require some more attention. If you could limit this PR to just the straight forward cases, it should be easy to merge. Then we can maybe look at the more complicated cases separately or leave them as is.

I made the requested changes.

folium/plugins/draw.py
```diff
@@ -60,6 +60,29 @@ class Draw(JSCSSMixin, MacroElement):
 
     _template = Template(
         """
+        {% macro html(this, kwargs) %}
+            {% if this.export %}
+            <style>
```
Conengmo, line 65: this style element was originally added to the header. Here it's in the body. I guess it doesn't make a difference really, does it?

Conengmo (approved):
Very nice cleanup!

PR #2096, opened 2025-02-20
Doc String -- Dictionary key of Gradient Argument -- update heat_map.py -- folium.plugins

krishna0125:
In the docstring, the gradient dictionary keys are mentioned as float, but I believe the backend code expects them to be strings. So, I changed the docstring to reflect this, from float values (e.g., .8) to string values (e.g., ".8").

### DIFF
--- a/folium/plugins/heat_map.py
+++ b/folium/plugins/heat_map.py
@@ -36,7 +36,7 @@ class HeatMap(JSCSSMixin, Layer):
         Amount of blur
     gradient : dict, default None
         Color gradient config. Defaults to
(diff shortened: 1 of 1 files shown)


hansthen:
There is an error in the automated tests. At first glance it does not appear to be related to your change. Will have a look at it over the weekend.

krishna0125:
@hansthen , were you able to look into the error during the weekend. Please let me know if you need help on this.

hansthen:
I had a look at your PR, but there seems to be an underlying issue. We did a refactor that broke passing dicts with non-string keys to Leaflet.

PR #2097, opened 2025-02-27
Fix problems with the documentation build

hansthen:
This solves an issue with the build

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -32,7 +32,7 @@ pytest
 scipy
 selenium<4.27.0
 setuptools_scm
-sphinx
+sphinx==7.3.7
 types-requests
 vega_datasets
 vincent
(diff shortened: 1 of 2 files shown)


hansthen:
@Conengmo Are you up for code reviews? You complained about being very busy a few weeks ago. I fixed a problem with the build due to a new sphinx version. It is a fairly small change that does not affect the main library functionality. If I don't hear from you, I will merge this myself soon.

requirements-dev.txt
```diff
@@ -32,7 +32,7 @@ pytest
 scipy
 selenium<4.27.0
 setuptools_scm
-sphinx
+sphinx==7.3.7
```
ocefpaf, line 35: What are the errors with the newer versions?

hansthen: `exception: module 'sphinx.util' has no attribute 'console'`

ocefpaf: I'll check that next week. Let's merge this one for now. Thanks!

PR #2103, opened 2025-03-06
unpin sphinx

ocefpaf:
Testing latest sphinx.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -32,7 +32,7 @@ pytest
 scipy
 selenium<4.27.0
 setuptools_scm
-sphinx==7.3.7
+sphinx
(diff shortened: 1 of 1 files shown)


requirements-dev.txt
```diff
@@ -32,7 +32,7 @@ pytest
 scipy
 selenium<4.27.0
 setuptools_scm
-sphinx==7.3.7
+sphinx
```
ocefpaf, line 35: @hansthen you added this pin but it seems that things are working without it. Any reason to keep it that I may be missing?

hansthen: Yes, the problem seems to be only with sphinx 8.2.1.

PR #2106, opened 2025-03-11
Docs: `utilities.image_to_url()` `'image'` and related docstrings

elliot-100:
Fixes #2905.

- Amends docstring for `utilities.image_to_url()` `'image'` parameter (not part of publicly documented API)
- Adapts this amend for these members of publicly documented API, depending on context:
  - `features.CustomIcon` `'icon_image'` parameter/attributes
  - `raster_layers.ImageOverlay` `'image'` parameter/attributes
- Uses more consistent indenting

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1862,13 +1862,13 @@ class CustomIcon(Icon):
 
     Parameters
     ----------
-    icon_image :  string, file or array-like object
-        The data you want to use as an icon.
-        * If string, it will be written directly in the output file.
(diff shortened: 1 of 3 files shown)


Conengmo (approved):
Thanks @elliot-100, great that these docs are now better in line with the actual functionality. I was building docs locally anyway so I made sure the lists turn out alright in the API reference in the docs: turns out you need an empty line before the list. Also made sure the number of spaces between the `*` and the text is one.

PR #2108, opened 2025-03-11
Docs: explain which layer kinds are base layer / overlay by default

elliot-100:
Explain which layer kinds are base layer / overlay by default.


Conengmo:
Thanks @elliot-100! Really nice you took the effort to improve the docs.

Issue #1713, opened 2023-01-26
Folium Search plugin doesn't work for FeatureGroup


Krukarius:
**Describe the bug**
The Folium Search plugin doesn't work for FeatureGroup. It simply returns nothing saying that "Feature is not found".

My code looks like this:

search_tool = Search(
    layer = or_cluster,
    search_label='post code',
    position='topright',
    search_zoom=20,
    geom_type='Point',
    popup_on_found=True
    )
map.add_child(search_tool)

Is it possible to sort it out? I use layer bound in Markerclusters.

Krukarius:
Related issue:

https://github.com/python-visualization/folium/issues/1711

Conengmo:
Could you include a fully working example? So that I can just run your code to reproduce your issue, without having to add anything. And please make it as minimal as possible.

Krukarius:
import pandas as pd
import folium

map = folium.Map(location=[51.86762,0.8624],
                 tiles='OpenStreetMap',
                 zoom_start=8,
                 min_zoom=7,
                 max_zoom=19,
                 max_bounds=True,
                 control_scale=True,
                 prefer_canvas=True
                 )
                 
df = pd.read_csv("or_geo.csv")

fo=FeatureGroup(name="Or",overlay = True)
or_cluster = MarkerCluster(name="General").add_to(map)

for i,row in df.iterrows():
    lat =df.at[i, 'lat']
    lng = df.at[i, 'lng']
    sp = df.at[i, 'sp']
    link = df.at[i, 'url']
    stat = df.at[i,'status']     

popup = folium.Popup(or_html, max_width=1000)
  

    or_marker = folium.Marker(location=[lat,lng], tooltip='<strong>Job details</strong>', popup=popup, icon = folium.Icon(
        color='blue', icon='glyphicon-calendar'))

    or_cluster.add_child(or_marker)
    
    This is the code. To be honest I have several layers such as this and would like to have this plugin functional for all of them on the same basis.
Thank you in advance,

Krukarius:
leaflet-search.min.js:17 
        
       Uncaught TypeError: Cannot read properties of undefined (reading 'properties')
       
       This is what JavaScript console says

Conengmo:
Here's a minimal example to search in a FeatureGroup:

```
coords = [45.5236, -122.6750]
m = Map(location=coords, tiles="carto db positron")

fg = FeatureGroup(name="fg1").add_to(m)
[...]

PR #2115, opened 2025-03-16
docs: add entry on Search plugin with FeatureGroup

Conengmo:
I noticed we don't have an example in our docs on how to use the Search plugin with anything else than a GeoJson object. So I added an example of FeatureGroup. This is based on https://github.com/python-visualization/folium/issues/1713. 

### Screenshot

afbeelding


hansthen (approved):
Looks good to me.

Conengmo:
Thanks!

PR #2116, opened 2025-03-16
CI: run Geopandas tests

Conengmo:
Add a workflow to run the Geopandas tests on the dev version of Folium.

Verified this works by pushing a commit to Folium that breaks the Geopandas tests, then restoring it of course.

### DIFF
--- a/.github/workflows/test_geopandas.yml
+++ b/.github/workflows/test_geopandas.yml
@@ -1,4 +1,4 @@
-name: Run Geopandas tests
+name: Geopandas tests
 
 on:
   pull_request:
--- a/.github/workflows/test_geopandas.yml
+++ b/.github/workflows/test_geopandas.yml
@@ -27,6 +27,7 @@ jobs:
     - name: Checkout Geopandas
       uses: actions/checkout@v4
       with:
+        repository: geopandas/geopandas
         path: geopandas
 
     - name: Install Geopandas dev dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/test_geopandas.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Install folium from source
+      shell: bash -l {0}
+      run: |
+        python -m pip install -e . --no-deps --force-reinstall
+
+    - name: Run Geopandas tests
+      run: |
+        cd geopandas
+        pytest
```
martinfleis, line 45:
```suggestion
        pytest geopandas/tests/test_explore.py
```

This is the only relevant file where we interact with folium in our CI. No need to run the other.

Conengmo: Thanks for the suggestion Martin!

.github/workflows/test_geopandas.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Checkout Geopandas
+      uses: actions/checkout@v4
+      with:
+        repository: geopandas/geopandas
+        path: geopandas
+
+    - name: Install Geopandas dev dependencies
+      run: |
+        pip install -r geopandas/requirements-dev.txt
+        pip install geodatasets
```
Conengmo, line 38: @martinfleis I was a bit surprised that I had to install `geodatasets` separately, is that expected? Or did I mis something? I don't mind, just checking.

martinfleis: Need to update requirements-dev. Sorry for that, we're not really using it.

Conengmo: No worries

hansthen (approved):
Looks good to me.

.github/workflows/test_geopandas.yml
martinfleis, line 38:
```suggestion
```

Should not be needed anymore.

Conengmo: Thanks for the quick update!

Issue #1992, opened 2024-09-07
Add Leaflet-Ruler plugin


MahmoudArafaaat:
i used all possible functions to add leaflet ruler plugin but it gives me errors like :
js: Uncaught ReferenceError: map is not defined

# Add Leaflet-Ruler plugin
        m.get_root().header.add_child(folium.CssLink("https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"))
        m.get_root().header.add_child(folium.CssLink("https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.css"))
        m.get_root().html.add_child(folium.JavascriptLink("https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"))
        m.get_root().html.add_child(folium.JavascriptLink("https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.js"))
        my_js = '''
        L.control.ruler().addTo(map);
        '''
        m.get_root().script.add_child(folium.Element(my_js))

i got all links from here: https://github.com/gokertanrisever/leaflet-ruler

hansthen:
Do you have a code example?

MahmoudArafaaat:
this is html of the map with the plugin :
https://jsfiddle.net/kvsgt9hu/

MahmoudArafaaat:
this one i reformed of the link

MahmoudArafaaat:
<!DOCTYPE html>
<html>
<head>
    <title>Leaflet Map with Ruler</title>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" />
    <link rel="stylesheet" type="text/css" href="https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.css">
    <style>
        #map { height: 600px; }
    </style>
</head>
<body>
    <div id="map"></div>
    <script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"></script>
    <script
    src="https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.js"
></script>
    <script>
        // Create a map centered around Cairo, Egypt
        var map = L.map('map').setView([30.0444, 31.2357], 8);

        // Add Esri Satellite tile layer
[...]

PR #2122, opened 2025-04-06
Allow users to add new Control classes without implementing a plugin

hansthen:
@Conengmo 

This is a new, smaller implementation of the `Control` class I implemented earlier. For this PR I focused only on the user's ability to load a new control in user space, without having to write a new plugin. It is not difficult to write a plugin, but it does require knowledge of Jinja. Also, there is quite some boilerplate code involved. 

With this `Control` clas I was able to implement the `minimap` and `ruler` plugins, using just a few lines of code.

```
import folium
import streamlit as st
from folium import Control, TileLayer

from streamlit_folium import st_folium

st.set_page_config(
    page_title="minimap",
    layout="wide"
)

m = folium.Map(
    location=[39.949610, -75.150282],
    zoom_start=5,
    zoom_control=False
)
tiles = TileLayer(
    tiles="OpenStreetMap",
    show=False,
    control=False,
)
tiles.add_to(m)

minimap = Control("MiniMap", tiles)
minimap.add_js_link(
    "minimap_js",
    "https://cdnjs.cloudflare.com/ajax/libs/leaflet-minimap/3.6.1/Control.MiniMap.min.js"
)
minimap.add_css_link(
    "minimap_css",
    "https://cdnjs.cloudflare.com/ajax/libs/leaflet-minimap/3.6.1/Control.MiniMap.css"
)
minimap.add_to(m)

ruler = Control("Ruler", tiles)
ruler.add_js_link(
    "ruler_js",
    "https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.js",
)
ruler.add_css_link(
    "ruler_css",
    "https://cdn.rawgit.com/gokertanrisever/leaflet-ruler/master/src/leaflet-ruler.css",
)
ruler.add_to(m)

st_folium(m, width=1600, height=600)
```


hansthen:
This also closes #1992

hansthen:
@ocefpaf can you have a look?

folium/features.py
```diff
@@ -1992,3 +2002,54 @@ def __init__(
...
+class Control(JSCSSMixin, MacroElement):
+    """
+    Add a Leaflet Control object to the map
+
+    Parameters
+    ----------
+    control: str
+        The javascript class name of the control to be rendered.
+    position: str
+        One of "bottomright", "bottomleft", "topright", "topleft"
```
ocefpaf, line 2016: One problem that we have in folium templates are the silent erros when passing keywords to it. For example, if someone types `bottonright` instead of `bottomright`, it will take user a while to figure out what is wrong.

Maybe we should check for this 4 valid kw? I know this doesn't solve it for all templates, but in this case it may be small enough that it is worth it.

hansthen: Do you mean as a static typing check or as an assert?

ocefpaf: We should avoid `assert` use  [outside of tests](https://docs.astral.sh/ruff/rules/assert/).

I'm old school but here is what I would do:

1. make position a kw argument, not an arg;
2. Sanitize it with a `position.lower()` to validate names copied from leaflet JS docs;
3. Check if the sanitized positions is in a valid list with those 4 names and raise a ValueError if not.
4. 
What do you think?

hansthen: Makes sense.

hansthen: I see we have been inconsistent in banning `assert` :-)

folium/features.py
```diff
@@ -1992,3 +2004,61 @@ def __init__(
...
+        position: Optional[TypePosition] = None,
+        **kwargs,
+    ):
+        super().__init__()
+        if control:
+            self._name = control
+
+        if position is not None:
+            position = position.lower()  # type: ignore
+            if position not in (args := get_args(TypePosition)):
```
ocefpaf, line 2059: While I was a Fortran programmer in my early days, type checking is not my forte. (Heck, I moved to Python to get away from this ;-p)

My crude review tells me that this is equivalent to what I described in my review comment, right? If so, merge away!

hansthen: I wrote my first program for money in RPG :-) which is a fairly obscure language almost as old as Fortran. Whitespace is significant to a fault.

```
     C* PARAMETER LIST
     C           *ENTRY    PLIST
     C                     PARM           DATEIN  80
     C* PROGRAM START
     C           DATEIN    DIV  100       DATEIN
     C                     MVR            DAY     20
     C           DATEIN    DIV  100       DATEIN
     C                     MVR            MONTH   20
     C                     Z-ADDDATEIN    YEAR    40
     C* SHOW RESULTS
     C           YEAR      DSPLY
     C           MONTH     DSPLY
     C           DAY       DSPLY
     C                     SETON                         LR
     C* EXECUTION EXAMPLE
     C* CALL SEPDATE PARM(X'019960531F')
```
I added one more test.

PR #2123, opened 2025-04-06
Add parameters to JsCode

### LINKED ISSUE #2121: Add extra parameters to JsCode objects
I'd like to be able to refer to other folium /leaflet objects inside a JsCode object. So far I have been using string interpolation inside the body of the JsCode object.

I'd like to add parameters to the JsCode `__init__` method, together with a basic string templating solution. For the string templating I'd prefer to use the good-old `%(name)s` format.

I considered using the newer formatting templates such as used in python `format`, or Jinja templates. However, due to brace heavy nature of javascript, I think a completely new variable marker will lead to more readable code.

**Implementation**
folium is maintained by volunteers. Can you help make a PR if we want to implement this feature?

hansthen:
closes #2121

### DIFF
--- a/folium/template.py
+++ b/folium/template.py
@@ -1,47 +1,6 @@
-import json
-from typing import Union
-
 import jinja2
-from branca.element import Element
 
-from folium.utilities import JsCode, TypeJsonValue, camelize
-
-
-def tojavascript(obj: Union[str, JsCode, dict, list, Element]) -> str:
-    if isinstance(obj, JsCode):
-        return obj.js_code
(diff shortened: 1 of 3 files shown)


hansthen:
I am making this a draft for now. I saw some movement on python to implement a new version of template strings, which uses a different format to specify parameters.

ocefpaf:
> I am making this a draft for now. I saw some movement on python to implement a new version of template strings, which uses a different format to specify parameters.

Hehe. With t-strings, if we had enough time, we should re-write `folium` without `jinja` 😄

PR #2129, opened 2025-04-08
Issue 2125 broken selenium tests

### LINKED ISSUE #2125: The selenium tests in the CI are broken
**Describe the bug**
The selenium tests in the CI are currently broken.

hansthen:
Version fix for jupytext to fix broken selenium tests. Closes #2125 .

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,7 +16,7 @@ gpxpy
 ipykernel
 jenkspy
 jupyter
-jupytext
+jupytext==1.16.*
 matplotlib
 mypy
 nbconvert
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -16,7 +16,7 @@ gpxpy
 ipykernel
 jenkspy
 jupyter
-jupytext==1.16.*
+jupytext==1.16.*   # Due to selenium test failures in 1.17.0.
 matplotlib
 mypy
 nbconvert


hansthen:
I am annoyed by these breaking CI's. 

Each time this happens it takes me quite some time to figure out exactly what package causes the issue. Is it an idea to add the output of `pip freeze` to the repository whenever we make a release? We can then create update our workflows so that on failure they compare the stored versions to the actual versions. This will help us identify breaking issues much faster.

ocefpaf:
> Is it an idea to add the output of `pip freeze` to the repository whenever we make a release? We can then create update our workflows so that on failure they compare the stored versions to the actual versions. This will help us identify breaking issues much faster.

We can have a "locked" CI and an unlocked one. That may  help? However, I like to test against unlocked CIs to be sure we are hitting these errors and not the users.

hansthen:
> We can have a "locked" CI and an unlocked one. That may help? However, I like to test against unlocked CIs to be sure we are hitting these errors and not the users.

By a locked CI you mean something like npm's lock files? I agree with you that I prefer to run the CI without version fixes.

That was not what I had in mind, though.

My idea is to (after a successful release) store the output of `pip freeze` somewhere for documentation purposes. Also, when a CI fails we can diff the current versions against the versions that produced the last successful CI run. 

Another idea is to periodically schedule a CI run. That way so we notice it earlier when the CI is broken and not on the next PR.

ocefpaf:
I'm OK with any of those ideas. Mine is kind of costly, it would always run 1 matrix item with a lockfile, in theory always passing with regards to dependency changes, b/c there none, plus what we have now. Then, if the locked one passes but the unlocked doesn't, we know to look for dependency changes.

hansthen:
@ocefpaf I understand. I will create a new branch to experiment with those ideas. Easiest is probably to start with scheduled workflows. In the meantime, can you have a look at the current PR? It is just a version fix until I can figure out what changed in  the `jupytext` package.

ocefpaf (approved):

PR #2132, opened 2025-04-13
Make DivIcon and CustomIcon work with the new Marker approach

hansthen:
This is a continuation of the PR by @conengmo.

To make the PR more robust against uses from other flows (like `add_child` and `add_to`), we check the ways that we can add children and catch setting the icon there.

The tests for this are included in the `playwright_tests` branch, for which there is another PR open.


hansthen:
@ocefpaf can you have a look?

ocefpaf:
Nice one!

PR #2134, opened 2025-04-18
Issue 1989: Convert null values to empty strings in popup and tooltip

### LINKED ISSUE #1989: Geojsonpopup change when cell is empty
**Describe the bug**
Very small bug, but there has been a small change in how empty values are handled and updating to the latest folium has affected how popups are displayed.
When creating a geojsonpopup using the fields keywords and a geodataframe, if a cell was empty the the content used to be empty. This allowed me to use simple css to hide row when empty:
```
    .leaflet-popup-pane tr:has(td:empty){
    display: none;
}
```
This behaviour was useful, as it was a simple way to add large amounts of data if present, but also simplify the output when no data available.
Since updating, the empty cells now have the string 'null' inside, breaking the styling. See example image:
image
Related to #1858 ?

**To Reproduce**
update to latest version of folium
```
m = Map()
gdf= gpd.read_file(geopackagePath, layer='groupedData')
data = gdf.drop('geometry', axis=1)
popup = folium.GeoJsonPopup(fields=data.columns.tolist(), labels=True, class_name="mystyle")
mapDataFrame= gdf.copy()
mapDataFrame['index'] = mapCopy.index
geojson = folium.GeoJson(data=mapDataFrame, popup=popup)
m.add_child(geojson)
[ Include a data sample or link to your data if necessary to run the code ]

```

**Expected behavior**
Empty cell should be empty in the popup

 - Browser chrome, firefox
 - html files
 - Python version sys.version_info(major=3, minor=11, micro=9, releaselevel='final', serial=0)
 - folium version: 0.17.0
 - branca version:0.7.2

**Possible solutions**
I haven't found a simple solution yet. I had a look at the source, but not sure where the actual change occurred.

hansthen:
Close #1989

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1108,7 +1108,15 @@ class GeoJsonDetail(MacroElement):
     function(layer){
     let div = L.DomUtil.create('div');
     {% if this.fields %}
-    let handleObject = feature=>typeof(feature)=='object' ? JSON.stringify(feature) : feature;
+    let handleObject = feature => {
+        if (feature === null) {
+            return '';
+        } else if (typeof(feature)=='object') {
+            return JSON.stringify(feature);
+        } else {
+            return feature;
+        }
+    }
     let fields = {{ this.fields | tojson | safe }};
     let aliases = {{ this.aliases | tojson | safe }};
     let table = '<table>' +


hansthen:
@ocefpaf can you have a look?

ocefpaf:
Can we add a test to avoid regression? Maybe ask in https://github.com/python-visualization/folium/issues/1989 for some code/file we can use?

hansthen:
> Can we add a test to avoid regression? Maybe ask in #1989 for some code/file we can use?

I wrote a small program to visually test this. But since this is behavior in the generated code, it is difficult to test with our current tests frameworks.  

I have a few ideas on how to make visual snapshots and compare against those during tests. Is it okay if I work on snapshot / regression testing in a new branch?

ocefpaf:
> I have a few ideas on how to make visual snapshots and compare against those during tests.

We can use https://github.com/matplotlib/pytest-mpl. I had some success with it in the past. However, we can try a text based text that checks if the HTML has `null` in those fields.

>Is it okay if I work on snapshot / regression testing in a new branch?

Sure. IMO it is OK to merge as-is and we can open an issue for further testing here.

hansthen:
> However, we can try a text based text that checks if the HTML has `null` in those fields.

It's not so easy. The `null` values are still there. We just filter them in javascript when we generate the popup.
```
    let handleObject = feature => {
        if (feature === null) {
            return '';
        } else if (typeof(feature)=='object') {
            return JSON.stringify(feature);
        } else {
            return feature;
        }
    }
```
I can of course create to test that specific string, but that would feel a bit pointless.

hansthen:
I created an issue for the regression tests. #2140

PR #2135, opened 2025-04-18
Issue 1877: Fix `_to_png` when using `fit_bounds`

### LINKED ISSUE #1877: Fit bounds not fitting to provided points
**Describe the bug**
I have a plot which I'd like to export as `png` so I'm using the `_to_png` feature. Ideally I'd like a zoom level of around 4.5 - my plot should cover from Madrid to St.Petersburg (roughly) - but I cannot provide this. When I provide bounds the image I get out doesn't seem to be obeying them.

I appreciate this is likely a function of dynamic window resizing, but exporting as `png` should be able to adhere to the provided bounds despite the `png` support being via selenium. 

**To Reproduce**

```python
import folium

m = folium.Map()

sw = [38, -3]
ne = [60, 31]
m.fit_bounds([sw, ne]) 

m.save("map.html")

import io
from PIL import Image

img_data = m._to_png(1)
img = Image.open(io.BytesIO(img_data))
img.save('map_plot.png')
```

**Expected behavior**
A `png` file contained within the provided bounds

 - Browser [e.g. chrome, firefox] Chromium based
 - Jupyter Notebook or html files? HTML and jupyer
 - Python version (check it with `import sys; print(sys.version_info)`) 3.12.1
 - folium version (check it with `import folium; print(folium.__version__)`) 0.15.1
 - branca version (check it with `import branca; print(branca.__version__)`) 0.7.1

hansthen:
Close #1877 

Beware, this change may break existing behavior. Previously all the images were generated against a full screen version of the map, even if the code contained a `fit_bounds` statement. We need to decide if this is a feature or a bug.

I did add a new parameter to `_to_png()` that allows the caller to set the size of the resulting image.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -358,7 +358,9 @@ class Map(JSCSSMixin, Evented):
             out = self._parent._repr_html_(**kwargs)
         return out
 
-    def _to_png(self, delay: int = 3, driver: Any = None) -> bytes:
+    def _to_png(
+        self, delay: int = 3, driver: Any = None, size: Optional[Sequence[int]] = None
+    ) -> bytes:
         """Export the HTML to byte representation of a PNG image.
 
         Uses selenium to render the HTML and record a PNG. You may need to
@@ -372,6 +374,7 @@ class Map(JSCSSMixin, Evented):
         >>> m._to_png(time=10)  # Wait 10 seconds between render and snapshot.
 
         """
+
         if self._png_image is None:
             if driver is None:
                 from selenium import webdriver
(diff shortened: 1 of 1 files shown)


hansthen:
@ocefpaf can you have a look?

ocefpaf:
> Previously all the images were generated against a full screen version of the map, even if the code contained a `fit_bounds` statement. We need to decide if this is a feature or a bug.

Indeed a breaking change, but in a way to_png was always an experimental feature, so I'm fine with the updated behavior that, IMO, is better.

ocefpaf (approved):

Issue #1520, opened 2021-10-20
Tooltip and Popup don't work in GeoJson with MultiPoint geometry


martinfleis:
**Describe the bug**
If you have MultiPoint geometry in your GeoJSON, Tooltip and Popup do not work. Normal Points are fine as well as MultiPolygons.

**To Reproduce**

```py
import json
import folium

geojson = '{"type": "FeatureCollection", "features": [{"id": "0", "type": "Feature", "properties": {"foo": 0}, "geometry": {"type": "MultiPoint", "coordinates": [[0.0, 0.0]]}}, {"id": "1", "type": "Feature", "properties": {"foo": 1}, "geometry": {"type": "MultiPoint", "coordinates": [[1.0, 1.0]]}}, {"id": "2", "type": "Feature", "properties": {"foo": 2}, "geometry": {"type": "MultiPoint", "coordinates": [[2.0, 2.0]]}}, {"id": "3", "type": "Feature", "properties": {"foo": 3}, "geometry": {"type": "MultiPoint", "coordinates": [[3.0, 3.0]]}}, {"id": "4", "type": "Feature", "properties": {"foo": 4}, "geometry": {"type": "MultiPoint", "coordinates": [[4.0, 4.0]]}}]}'

geojson = json.loads(geojson)

m = folium.Map()
folium.GeoJson(
    geojson,
    tooltip=folium.GeoJsonTooltip(["foo"]),
    popup=folium.GeoJsonTooltip(["foo"]),
    marker=folium.CircleMarker(radius=20)
).add_to(m)
m
```

**Expected behavior**
Both tooltip and popup should normally work as with any other geometry type.

**Environment (please complete the following information):**
 - Browser [e.g. chrome, firefox]: Safari
 - Jupyter Notebook or html files? Jupyter
 - Python version (check it with `import sys; print(sys.version_info)`): 3.9.6
 - folium version (check it with `import folium; print(folium.__version__)`): master
 - branca version (check it with `import branca; print(branca.__version__)`): master

**Additional context**
xref https://github.com/geopandas/geopandas/issues/2190

**Possible solutions**
You can explode MultiPoints to Points but that is not a viable solution as it breaks the geometry into pieces.

I can have a look at the fix unless it requires a lot of JavaScript :).

hansthen:
[...]

PR #2136, opened 2025-04-18
Issue 1770: Add on_each_feature handler to GeoJson

### LINKED ISSUE #1770: Support GeoJsonToolTips/GeoJsonPopup when using GeoJson together with the marker cluster plugin
At the moment, it is possible to use `GeoJson` and add a tooltip based on the `GeoJson` property. However, as soon as the marker cluster plugin is used, this doesn't work anymore. Means, the tooltips are just not shown.

I spent a lot of time searching in the internet for a solution but I couldn't find anything. Therefore, I debugged the issues by myself. As a hint: I have experience in Python but I'm not at all a JavaScript expert. So, please excuse possible stupid questions.

The problem with current appraoch in _folium_ is the following: A `GeoJsonTooltip` object is add as child to the `GeoJson` object and during code generation this leads to JavaScript code where the `bindTooltip` method of the `GeoJson` object is used. According to the _Leaflet_ doc this seems to be right but what actually need to be done is, calling the `bindTooltip` method of the layer for each feature. I tested this on my fork and it works. With the current implementation, you can even see a Javascript error in the debug console of the browser. 

To make it work, I changed the template code of the`GeoJsonTooltip`object so that the `bindTooltip` method of its parent (`GeoJson`object) is not directly called but generates a JavaScript function which must be called in the `onEachFeature` callback in the generated `GeoJson` JavaScript code. However, this concept does not feel right. It might be a general solution for the `GeoJsonTooltip` class but will probably not work for the general `Tooltip` class or at least breaks existing code.

I thought a bit about a general solution. Maybe an extra argument like `on_each_feature` in the Python `add_child` method could indicate the child object to change the generated code so that the parent object can bind the tooltip on each feature instead. But I'm not so familiar with the whole project. Maybe there is a much better solution.

hansthen:
Close #1770.

This will allow users to customize the Tooltips and Popups for GeoJson data. It should also help with issue #1520.

Testing was done manually, using a streamlit program.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -28,6 +28,7 @@ from folium.folium import Map
 from folium.map import FeatureGroup, Icon, Layer, Marker, Popup, Tooltip
 from folium.template import Template
 from folium.utilities import (
+    JsCode,
     TypeBoundsReturn,
     TypeContainer,
     TypeJsonValue,
@@ -516,6 +517,10 @@ class GeoJson(Layer):
         embedding is only supported if you provide a file link or URL.
     zoom_on_click: bool, default False
         Set to True to enable zooming in on a geometry when clicking on it.
+    on_each_feature: JsCode, optional
+        Javascript code to be called on each feature.
+        See https://leafletjs.com/examples/geojson/
+        `onEachFeature` for more information.
     **kwargs
         Keyword arguments are passed to the geoJson object as extra options.
 
@@ -585,6 +590,10 @@ class GeoJson(Layer):
         {%- endif %}
 
         function {{this.get_name()}}_onEachFeature(feature, layer) {
+            {%- if this.on_each_feature %}
+            ({{this.on_each_feature}})(feature, layer);
+            {%- endif %}
+
             layer.on({
                 {%- if this.highlight %}
                 mouseout: function(e) {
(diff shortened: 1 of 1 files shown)


hansthen:
@ocefpaf can you have a look?

ocefpaf:
> Testing was done manually, using a streamlit program.

Is there any testing we can do here? To check if the rendering as expected?

hansthen:
> > Testing was done manually, using a streamlit program.
> 
> Is there any testing we can do here? To check if the rendering as expected?

I have a proposal for regression testing in the playwright branch. It makes use of Playwrights visual snapshots to compare the current version against stored images under version control. 

However, when working on the #1877 issue, I got an idea that may be even simpler. We can use `_to_png()` with fixed size to make visual snapshots. 

I have a boatload of regression tests that I'd like to check in. I think it would make folium development much more solid if we had better snapshot testing.

ocefpaf:
Got it. Looks like a small code change with a larger impact. While I'm OK merging this as-is not. I would like an issue to track the development of the tests. What do you think?

hansthen:
I created an issue for the regression tests: #2140.

Issue #2140, opened 2025-04-22
Create regression tests for the recent bugfixes


hansthen:
We recently fixed some behavior related to popups and tooltips that only show up after interacting with the map. 
We currently do not have a framework for this type of regression tests where user interaction is required.

hansthen:
Should also create tests for #2138.

PR #2144, opened 2025-05-11
New snapshot tests

hansthen:
A framework for performing easy regression tests.

My usual workflow for testing issues is as follows

1. create a small program to reproduce an issue
2. perform the fixes
3. use the reproduction scenario to verify that the issue is gone

I also created a github test workflow for these snapshots. The snapshot tests create screenshots which are then compared against earlier screenshots which are stored in revision control. On failure both the new screenshot and a visual difference is uploaded for easy comparison.

This was originally written using playwright and streamlit, but I found it easier to make the screenshots directly.


hansthen:
Closes #2140

ocefpaf (approved):
LGTM!

PR #2153, opened 2025-06-02
Fix or mitigate https://github.com/python-visualization/folium/issues…

sgrefen:
…/2152. CustomIcon doesn't exist (anymore?) in leaflet


hansthen:
I updated the commit slightly. I wanted to make it as a suggestion, but I forgot how to do that. Let me know if you are okay with my change.

hansthen:
Indeed.

hansthen:
@sgrefen Thanks! Your feature is live now.

PR #2158, opened 2025-06-09
Coverage testing

hansthen:
Add coverage measurements to the tests and upload coverage report.

### DIFF
--- a/.github/workflows/test_geopandas.yml
+++ b/.github/workflows/test_geopandas.yml
@@ -47,3 +47,11 @@ jobs:
       run: |
         cd geopandas
         coverage run -p -m pytest -r a geopandas/tests/test_explore.py
+
+    - name: Upload coverage
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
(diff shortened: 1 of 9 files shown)


hansthen:
The streamlit-folium-tests are excluded from the coverage. I could not get playwright to work within the coverage module. Also, I did not yet test the final upload of the coverage report, since that only works on the main branch.

.github/workflows/test_geopandas.yml
```diff
@@ -46,4 +46,12 @@ jobs:
     - name: Run Geopandas tests
       run: |
         cd geopandas
-        pytest -r a geopandas/tests/test_explore.py
+        coverage run -p -m pytest -r a geopandas/tests/test_explore.py
+
+    - name: Upload coverage
+      if: always()
```
ocefpaf, line 52: This this needed?

ocefpaf: Did some research. That is it. I learned something new 😄

PR #2160, opened 2025-06-11
Issue 2014 - Multilingual Mapping Use

hkfrese:
Issue:
People wanted to make the same map twice - once with English numbers (like 1,234.56) and once with French (or really any other country outside of the US styling) numbers (like 1 234,56). But Folium was stuck using whatever language the computer was set to.
Proposed Solution:
I created a helper tool that can change how numbers look in map legends without changing the computer's language settings.
How It Works:

Make the map normally
Add code - it waits for the map to finish loading
Find all the numbers - in the legend/scale on the side
Change the format - swap periods for commas, spaces for commas, etc.
Save different versions - now you can make an English version AND a "foreign" version

Additional Info:

You don't need to restart your computer or change system settings, and while the default was French (because that's my second language anyways), you can make as many language versions as you want. Other people can easily add new languages by just adding the number rules. It should work with any map data you already have

### DIFF
--- a/languagetest.py
+++ b/languagetest.py
@@ -1,0 +1,374 @@
+"""
+Folium Multi-Language Choropleth Solution
+
+This module provides multiple approaches to render folium choropleths 
+with different number locales for legends, enabling creation of maps
+in multiple languages without changing system locale.
+
+Requirements:
+- folium
+- pandas
+- geopandas (optional, for geojson data)
+- selenium (for image generation)
+- pillow (for image processing)
+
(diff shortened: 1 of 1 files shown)


hansthen:
Thanks for this PR! I had a quick look. 

There are some issues in the pre-commit hook, though. Probably that means you missed a step in setting up your local environment. Please have a look here: https://github.com/python-visualization/folium/blob/main/.github/CONTRIBUTING.md.

hansthen:
Also, the geoman documentation file seems to be empty.

PR #2169, opened 2025-07-08
enforce py39 min

ocefpaf:
Should help with #2167

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -34,7 +34,7 @@ jobs:
         environment-name: TEST
         init-shell: bash
         create-args: >-
-          python=3.12
+          python=3
           --file requirements.txt
           --file requirements-dev.txt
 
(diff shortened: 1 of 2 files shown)


ocefpaf:
pre-commit.ci autofix

hansthen:
> @hansthen I'd love your help with the mypy failure here.

The issue is upstream in Branca. The offending code returns `Element._parent` as `Element` but in Branca that is an `object`.

I was able to get rid of the error using this:
```
diff --git a/branca/element.py b/branca/element.py
index 1912aca..ad21bbf 100644
--- a/branca/element.py
+++ b/branca/element.py
@@ -55,7 +55,7 @@ class Element:
         self._name = "Element"
         self._id = hexlify(urandom(16)).decode()
         self._children = OrderedDict()
-        self._parent = None
+        self._parent: Element = None
         self._template_str = template
         self._template_name = template_name
``` 
However, Branca does not use much typing so far. Let me check if there is a better way.

hansthen:
I could not get typing to work without either changing `branca` or adding `# type: ignore` line to the offending function.

ocefpaf:
Thanks! Please merge I'd you are happy with this PR.

PR #2172, opened 2025-08-09
Use uv for streamlit_folium installation

hansthen:


ocefpaf (approved):
LGTM. What was the motivation? I'm not against `uv`, but whatever we do here should work with plain `pip` too. If we had issues with `pip` we should try to solve them. If it is just speed, merge away!

PR #2193, opened 2025-10-16
Add text elements docs

craigsw86:
This is me contributing to the open source project.


hansthen:
@craigsw86 Thanks for the PR. I did not have time to look at it yet. Also I notice there are two PR's with similar purposes.

PR #2195, opened 2025-11-02
Pyproj build failures

hansthen:


.github/workflows/test_streamlit_folium.yml
```diff
@@ -43,6 +43,9 @@ jobs:
     - name: Install streamlit_folium dev dependencies
       run: |
         cd streamlit_folium
+        echo hello world
```
ocefpaf: Do we need this? Looks like a debug line, right? Same for the one after the installation.

I wonder if we should either restrict the tests here to not need pyproj/proj, we are focusing on `streamlit_folium` and I believe we can do that without them. Or we can use conda for the tests. I'll leave to you to decide on the best approach.

PR #2213, opened 2026-03-08
build(pyproject): declare `requires-python` via PEP 621 `[project]` metadata

chid:
Add a minimal `[project]` table to `pyproject.toml` and set:

  - `requires-python = ">=3.9"`
  - `name = "folium"`
  - `dynamic = [...]` for fields still sourced from `setup.py`

  Why this is needed:

  `python_requires` in `setup.py` is not enough for tools that read interpreter constraints from `pyproject.toml` metadata. Per PEP 621, `requires-python` belongs in `[project]`, which allows `uv` to resolve/lock against the correct Python range instead of defaulting incorrectly (see https://github.com/astral-sh/uv/issues/11424).

### DIFF
--- a/folium/plugins/beautify_icon.py
+++ b/folium/plugins/beautify_icon.py
@@ -93,7 +93,7 @@ class BeautifyIcon(JSCSSMixin, MacroElement):
         inner_icon_style="",
         spin=False,
         number=None,
-        **kwargs
+        **kwargs,
     ):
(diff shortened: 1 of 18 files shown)


hansthen:
The pre-commit hooks are failing on the autoformatter. (Which is weird because I see no change in the corresponding files). This change is required for uv? It seems okay for now.

PR #2220, opened 2026-04-30
Fix sleeping before taking a screenshot in Map._to_png

neezr:
In the experimental `Map._to_png`, currently a selenium `WebDriver` opens the local HTML file and `time.sleep()` is used to wait for the folium `Map` to load before taking the screenshot. In this PR, I replaced `time.sleep` with `WebDriverWait`, which actually awaits the `Map` loading in JavaScript.

The `delay` variable now specifies the _maximum_ of time to wait, the `WebDriverSleep` object stops waiting once the page actually loaded.

Also fixed a small typo in the docstring of `_to_png`.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -392,11 +392,16 @@ class Map(JSCSSMixin, Evented):
                     *size,
                 )
                 driver.set_window_size(*window_size)
+            from selenium.webdriver.support.ui import WebDriverWait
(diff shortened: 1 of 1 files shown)


hansthen:
Very cool! I am running the tests as we speak, but it looks like it cannot go wrong :-)

PR #2223, opened 2026-05-15
fix test with proj 9.8.1

autra:
Apparently newest proj doesn't accept `epsg: 4326` (with a space) any more. This is not yet visible everywhere because current pyproj ships proj 9.5.1 on pypi, but on nixpkgs, the proj version is already 9.8.1.

Thanks!


hansthen:
Running the automated tests now. This looks like a nice cleanup.

PR #2228, opened 2026-05-29
Add LayerGroup support (#1034)

### LINKED ISSUE #1034: LayerGroup support
#### Short question
Will Folium ever allow a generation of LayerGroup, or a decision had been made to keep only FeatureGroup?

#### Details
According to LeafletJS documentation, [FeatureGroup](https://leafletjs.com/reference-1.3.4.html#featuregroup) extends [LayerGroup](https://leafletjs.com/reference-1.3.4.html#layergroup). They have different _JS_ constructors: 
```L.featureGroup(<Layer[]> layers)``` with only a list of layers, 
```L.layerGroup(<Layer[]> layers?, <Object> options?)``` with some **options** in addition to layers.

When creating a FeatureGroup in Folium, there is a parameter `name` in its _Python_ constructor (as well as some visualization-related params):
```__init__(self, name=None, overlay=True, control=True, show=True)```
However, this `name` is only used to display a name of this featuregroup in a LayerControl:
```
var feature_group_bdeeee... = L.featureGroup(name="events").addTo(map_...);
var layer_control_... = {
    base_layers : { "openstreetmap" : ...},
    overlays : { "events" : feature_group_bdeeee... }
}
```

What I need to have in a generated code for my use-case is a constructor with those **options** parameter, proposed by [LayerGroup](https://leafletjs.com/reference-1.3.4.html#layergroup-l-layergroup):
```var layer_group_1... = L.layerGroup(null, {'step': '1'})
var layer_group_2... = L.layerGroup(null, {'step': '2'})
```

This would allow to access those attributes  (**step** in my example) of my featuregroup (layergroup) using some custom JS code to put listeners on some objects.  

Example of use case:  when hovering `step=5` name in a custom `div` element, 5th circle from `events` group and 5th marker of another (say, `places`) group change their CSS classes to highlight the connection between 5th "event" and "place", a bit like `zip` in Python.

To make it possible, i need `L.layerGroup(null, {'step': '1'})` generated code instead of just empty `L.featureGroup()`.  I could write corresponding code and make a PR, if ever this functionality is welcomed by the community :-) 

`folium version 0.7.0.`

Sorry if anything is unclear, this is the 1st issue i open. Don't hesitate to ask for more details.
Thanks!

TheChyeahhh:
Closes #1034

Adds a new `LayerGroup` class to folium that mirrors `FeatureGroup` but uses `L.layerGroup()` in the generated JavaScript instead of `L.featureGroup()`.

In LeafletJS, `L.layerGroup(layers?, options?)` accepts additional custom options that can be used in custom JavaScript code, while `L.featureGroup(layers)` does not support options.

This is a lightweight addition — the class shares the same Layer parent and behaves identically to FeatureGroup for rendering and layer control purposes.

Changes:
- `folium/map.py`: Added `LayerGroup` class with `L.layerGroup()` template
- `folium/__init__.py`: Export `LayerGroup`
- `tests/test_folium.py`: Added test for LayerGroup rendering and custom options


hansthen:
Interesting PR. I had to run black on the code because it did not comply with the style guide. Can you give a concrete example when this would be useful? It would be nice if we had some documentation for this class.

hansthen:
Looks good. I merged the code.

PR #2229, opened 2026-05-29
fix: use viewport units for percentage map dimensions (fixes #2186)

### LINKED ISSUE #2186: Add better control for height when browser window is not maximized
I can set the map to something like:
folium.Map(location=[x, y], height='1000px')

and it looks proper when a browser is full screen.  However if the browser window is not full screen then the map goes to a much reduced height.

If I use folium.Map(location=[x, y], height='100%')

then the height seems to potentially become 50px.

I would like to be able to set both height and width at a percentage and the display is proper in a maximized or not maximized browser window.  Straight forward such as this:

folium.Map(location=[x, y], height='100%')
folium.Map(location=[x, y], width='100%')

Tried a few CSS controls but could not seem to get something permanent.

**Additional context**

Kakarot35:
Fixes #2186

## Problem
`folium.Map(height='100%')` collapses to nearly zero height when the map
is embedded inside a Django or Flask template. This happens because
`height: 100%` in CSS requires every ancestor element to have an explicit
height — which page templates typically don't have.

Similarly, `folium.Map(height='1000px')` can shrink in a non-maximized
browser window because there is no minimum floor set.

## Changes

- **Viewport units for percentages**: when `height` or `width` is given as
  a percentage, emit `vh`/`vw` instead of `%`. Viewport units are always
  relative to the browser window size, not the ancestor chain, so they work
  correctly in both standalone and embedded (Django/Flask) contexts.

- **Min constraints for pixel sizes**: when `height` or `width` is given in
  pixels, also emit `min-height`/`min-width` so the map cannot shrink below
  its intended size in a smaller window.

- **Remove dead CSS rule**: the template was emitting
  `#map { position:absolute; ... }` which never matched anything — folium
  uses hashed IDs like `map_d7a447c...`, not `map`. Removed.

## Before / After

```css
/* Before — collapses in embedded templates */
#map_abc123 { height: 100%; }

/* After — always fills the viewport correctly */
#map_abc123 { height: 100vh; }
```

## Tests
13 new tests added covering percentage → viewport units, pixel → min
constraints, dead rule removal, and full backwards compatibility.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -5,8 +5,7 @@ Make beautiful, interactive maps with Python and Leaflet.js
 
 import time
 import webbrowser
-from collections.abc import Sequence
-from typing import Any, Optional, Union
+from typing import Any, List, Optional, Sequence, Union
 
 from branca.element import Element, Figure
 
@@ -64,12 +63,14 @@ _default_css = [
 
 
 class GlobalSwitches(Element):
-    _template = Template("""
+    _template = Template(
+        """
         <script>
             L_NO_TOUCH = {{ this.no_touch |tojson}};
             L_DISABLE_3D = {{ this.disable_3d|tojson }};
         </script>
-    """)
+    """
+    )
 
     def __init__(self, no_touch=False, disable_3d=False):
         super().__init__()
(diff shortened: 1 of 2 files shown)


Kakarot35:
@hansthen The failing "Publish Docs" check was caused by a missing export in folium/plugins/__init__.py — now fixed and merged via #2230. This PR should be good to review now. Would appreciate your feedback!

folium/folium.py
```diff
@@ -392,16 +396,11 @@ def _to_png(
                     *size,
                 )
                 driver.set_window_size(*window_size)
-            from selenium.webdriver.support.ui import WebDriverWait
```
hansthen, line 395: Not sure why these changes are necessary? Did you branch from an older version of the code? This was changed not too long ago. Using `WebDriverWait` ensures we wait only until the appropriate conditions are met, instead of waiting a fixed amount of time.

hansthen:
Apart from the change that removed WebDriverWait, this looks like a great addition. Could you rebase the change on latest master, so we don't have the problem documentation not building?

Kakarot35:
@hansthen Thanks for the feedback! I see that you are correct, since it was based off of an earlier version, my branch accidentally stripped out WebDriverWait. However, I have rebased back to the current main and put folium/folium.py back into upstream with only the viewport unit updates. The WebDriverWait functionality remains completely intact.

Kakarot35:
@hansthen I will open another PR for the formatting error in GroupedLayerControl documentation. I think the problem with network timeout to pae-paha.pacioos.hawaii.edu is due to external server connection - please advise whether I should look into that as well.

hansthen:
> @hansthen I will open another PR for the formatting error in GroupedLayerControl documentation. I think the problem with network timeout to pae-paha.pacioos.hawaii.edu is due to external server connection - please advise whether I should look into that as well.

The network timeout of pae-paha.pacioos.hawaii.edu is not related to your change. I think the university has blocked the github servers, as manually accessing the data works fine. 

I you want you can have a look at the timeout issue. (Perhaps we can store the downloaded file somewhere else and use that.) The problem is intermittent though. I can merge the change if that is the only issue open.

PR #2235, opened 2026-06-06
add coordinate_input plugin

banderlog:
Add input form to the map for entering coordinates and placing markers at those locations.
Supports marker removal by double-clicking.

image


hansthen:
@banderlog thank you for the pull request. I am a bit hesitant to accept it into core Folium. I see Folium as a thin wrapper around Leaflet and its plugins. This also feels like something that would not very useful for most people. Also we have the Draw and Geoman plugins. Perhaps you can distribute this plugin by yourself? We have some other Folium plugins that are not part of Folium itself.

banderlog:
@hansthen Hi, no problem

https://github.com/banderlog/folium-coord_input-plugin

PR #2242, opened 2026-07-03
overlay the time slider so it stops hiding the attribution

### LINKED ISSUE #1301: TimeSliderChoropleth hides the attribution
#### Problem description
See the last cell at https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/TimeSliderChoropleth.ipynb. The attribution is supposed to be at the lower-right corner but it's hidden. (It is in HTML but the attribute layer doesn't show.)

#### Expected Output
Other plugins don't hide the attribution: https://nbviewer.jupyter.org/github/python-visualization/folium/blob/master/examples/Plugins.ipynb

0.10.1

CodingSelim:
Fixes #1301.

The `TimeSliderChoropleth` slider was inserted into the body as a block sibling of the map. Since the map fills 100% of the height, adding the slider pushed the whole page taller than the viewport and the Leaflet/tile attribution got clipped off the bottom. You could see the slider or the attribution, never both.

Positioned the slider as an absolute overlay at the top of the map (small translucent background so it stays readable). The map keeps its full height and the attribution is visible again. Matches the maintainer's earlier suggestion of putting the slider on top of the map rather than changing how `Map` sizes itself.

Verified in a browser: before, the page overflowed by the slider's height and the attribution sat below the fold; after, no overflow and the attribution shows bottom-right with the slider still working at top-center. Added a small test that doesn't need geopandas.

### DIFF
--- a/folium/plugins/time_slider_choropleth.py
+++ b/folium/plugins/time_slider_choropleth.py
@@ -50,7 +50,17 @@ class TimeSliderChoropleth(JSCSSMixin, Layer):
             }
 
             let slider_body = d3.select("body").insert("div", "div.folium-map")
-                .attr("id", "slider_{{ this.get_name() }}");
+                .attr("id", "slider_{{ this.get_name() }}")
+                // overlay the slider on the map so it doesn't push the map
+                // (and its attribution) out of view
+                .style("position", "absolute")
+                .style("top", "10px")
+                .style("left", "50%")
+                .style("transform", "translateX(-50%)")
+                .style("z-index", "999")
+                .style("background", "rgba(255, 255, 255, 0.8)")
+                .style("border-radius", "4px")
+                .style("padding", "2px 8px");
             $("#slider_{{ this.get_name() }}").hide();
             // insert time slider label
             slider_body.append("output")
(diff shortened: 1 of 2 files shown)


CodingSelim:
heads up, the failing docs check is unrelated to this PR. it dies executing user_guide/geojson/geojson.md with an ImportError (pyogrio/fiona not available in the docs env), and this branch only touches time_slider_choropleth. same docs build is currently red on main too, so it looks like a pre-existing env issue not something here. code and geopandas tests pass.

hansthen:
> heads up, the failing docs check is unrelated to this PR. it dies executing user_guide/geojson/geojson.md with an ImportError (pyogrio/fiona not available in the docs env), and this branch only touches time_slider_choropleth. same docs build is currently red on main too, so it looks like a pre-existing env issue not something here. code and geopandas tests pass.

We are currently working on the pygogrio/fiona errors and I can confirm it is a pre-existing issue. Any idea what could cause the snapshot error? It also does not appear to be related to your change. I will rerun the test to see if it is just flakyness.

CodingSelim:
thanks for confirming. looks like the snapshot test came back green on the rerun, the only red one left is that same geojson.md docs build failing on the pyogrio/fiona import, so nothing outstanding from this branch as far as i can tell. let me know if you want me to change anything.

hansthen:
> thanks for confirming. looks like the snapshot test came back green on the rerun, the only red one left is that same geojson.md docs build failing on the pyogrio/fiona import, so nothing outstanding from this branch as far as i can tell. let me know if you want me to change anything.

After a rerun that test also failed. We experience more than usual flakyness from GH lately.

PR #2244, opened 2026-07-06
Fix HeatMap crash on integer numpy weight arrays

gaoflow:
### Problem

`HeatMap`'s docstring documents *"a numpy.array of shape (n,2) or (n,3)"* as
valid input, but an all-integer array crashes at render time:

```python
import numpy as np, folium

m = folium.Map()
HeatMap(np.array([[3, 4, 1], [5, 6, 2]])).add_to(m)   # int64 dtype
m.get_root().render()
# TypeError: Object of type int64 is not JSON serializable
```

The float-weight version renders fine. This is common in real usage, e.g.
`HeatMap(df[["lat", "lon", "count"]].values)` where the weight column has an
integer dtype (pandas `.values` yields an `int64` array).

### Cause

In `HeatMap.__init__`, `validate_location` coerces the lat/lon columns to Python
`float`, but the weight column (`line[2:]`) is passed through unchanged:

```python
self.data = [
    [*validate_location(line[:2]), *line[2:]] for line in data
]
```

`np.float64` **is** a subclass of `float`, so `json.dumps` (via the template's
`|tojson`) tolerates float weights; `np.int64` is **not** a subclass of `int`,
so integer weights raise. The existing `test_heatmap_data` masks this because
its array contains a `0.5`, forcing the whole array to `float64`.

### Fix

Coerce the weight column to `float`, mirroring the lat/lon normalization. The
`NaN` guard immediately below still works (`float(np.nan)` is caught by
`np.isnan`).

### Tests

- `test_heatmap_integer_numpy_weights`: an `int64` (n,3) array — weights
  normalized to Python `float`, render contains `L.heatLayer`, and integer vs
  float arrays produce identical `.data`.
- `test_heatmap_integer_numpy_no_weight`: an `int64` (n,2) array renders.

`pytest tests/plugins/` passes (the one unrelated `test_time_slider_choropleth`
failure is a pre-existing missing optional dependency, identical on a clean
checkout). `ruff check` clean.

### DIFF
--- a/folium/plugins/heat_map.py
+++ b/folium/plugins/heat_map.py
@@ -79,7 +79,8 @@ class HeatMap(JSCSSMixin, Layer):
         self._name = "HeatMap"
         data = if_pandas_df_convert_to_numpy(data)
         self.data = [
-            [*validate_location(line[:2]), *line[2:]] for line in data  # noqa: E999
+            [*validate_location(line[:2]), *[float(w) for w in line[2:]]]
+            for line in data  # noqa: E999
         ]
         if np.any(np.isnan(self.data)):
             raise ValueError("data may not contain NaNs.")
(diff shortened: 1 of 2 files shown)


hansthen:
Very well documented change. Looks good, I will run the automated tests.

PR #2246, opened 2026-07-13
Add zizmor checks and fix all lints

ocefpaf:
@hansthen I'll prepare the PR to fetch the supporting data from a release afte this one. I had this sitting on my laptop for a long time.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,5 +1,8 @@
 name: Publish Docs
 
+# no permissions by default
+permissions: {}
+
 on:
   schedule:
     - cron: "0 13 * * *"
@@ -24,12 +27,12 @@ jobs:
         shell: bash -l {0}
 
     steps:
-    - uses: actions/checkout@v7
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
       with:
         fetch-depth: 0
 
     - name: Setup Micromamba env
-      uses: mamba-org/setup-micromamba@v3
+      uses: mamba-org/setup-micromamba@d7c9bd84e824b79d2af72a2d4196c7f4300d3476  # v3.0.0
       with:
         environment-name: TEST
         init-shell: bash
@@ -45,6 +48,7 @@ jobs:
       run: |
         set -e
         pushd docs
+
         make clean html linkcheck
         popd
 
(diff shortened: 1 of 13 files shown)


ocefpaf:
We need to either find another source for the WMS time plugin or we need to remove it from the docs. The source we are using there is failing and prevent us from building the docs.

```
    ConnectTimeout: HTTPSConnectionPool(host='pae-paha.pacioos.hawaii.edu', port=443): Max retries exceeded with url: /thredds/wms/dhw_5km?service=WMS&request=GetCapabilities&version=1.1.1 (Caused by ConnectTimeoutError(<HTTPSConnection(host='pae-paha.pacioos.hawaii.edu', port=443) at 0x7ff7942367b0>, 'Connection to pae-paha.pacioos.hawaii.edu timed out. (connect timeout=100)'))
```

@hansthen are you OK with removing this part of the docs?

hansthen:
> @hansthen are you OK with removing this part of the docs?
Yes. Do you want me to remove it in a separate PR or will you make it part of this one?

.github/workflows/deploy-docs.yml
```diff
@@ -24,12 +27,12 @@ jobs:
         shell: bash -l {0}
 
     steps:
-    - uses: actions/checkout@v7
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
```
hansthen, line 29: Why is this required? I take it the long string is a version hash? What will this mean for future maintenance?

ocefpaf:
> Yes. Do you want me to remove it in a separate PR or will you make it part of this one?

I can do it after this one. I'm still looking for alternatives and/or keeping some parts of the example alive without reaching the URL. More on that soon.

Let me know if you are OK with this PR and I'll tackle the docs+downloadable assets right after.

hansthen (approved):

hansthen:
> > Yes. Do you want me to remove it in a separate PR or will you make it part of this one?
> 
> I can do it after this one. I'm still looking for alternatives and/or keeping some parts of the example alive without reaching the URL. More on that soon.
> 
> Let me know if you are OK with this PR and I'll tackle the docs+downloadable assets right after.

I'm okay with it. I just had one question (why the long version hashes are required).

.github/workflows/deploy-docs.yml
ocefpaf: Dependabot should update those automatically, and the comment with the human readable version.

This is necessary b/c only a hash is immutable, the releases are not. It is one of the security checks that zizmor does.

ocefpaf: BTW  GitHub did introduce immutable releases recently, but those are optional and it is not worth checking each project to ensure they are using it or not. The recommendation is to still use the hashes.

ocefpaf:
I'll address the docs failure next.

PR #2250, opened 2026-07-24
Fix TopoJson non-collection geometries

### LINKED ISSUE #1816: TopoJSON fails with Polygon/MultiPolygon objects
**Describe the bug**

I got a topojson file and want to draw it with folium.
This failed with:

```
  File ".../lib/python3.11/site-packages/folium/features.py", line 981, in style_data
    geometries = recursive_get(self.data, self.object_path.split("."))[
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'geometries'
```

**To Reproduce**

```python
topo_json = {
  "type": "Topology",
  "objects": {
    "A": {
      "type": "Polygon",
      "id": "A",
      "bbox": [
        9.47021484375, 46.34692761055676, 17.259521484375, 49.059469847170526
      ],
      "arcs": [[0]]
    }
  },
  "arcs": [
    [
      [28, 4614],
      [437, 219],
      [607, -875],
      [99, 738],
      [818, -109],
      [56, -437],
      [338, 54],
      [1185, 1255],
      [874, -626],
      [-310, 1657],
      [423, 782],
      [325, 189],
      [169, 832],
      [141, 107],
      [240, -161],
      [112, 882],
      [424, -507],
      [268, -214],
      [832, 1599],
      [394, -80],
      [1072, -718],
      [282, -187],
      [212, 347],
      [606, -480],
      [71, -829],
      [-155, -511],
      [451, -1430],
      [-183, -1224],
      [-508, -81],
      [-127, 272],
      [-99, -163],
      [240, -246],
      [56, -300],
      [-239, -713],
      [-28, -1294],
      [-311, -110],
      [-366, -609],
      [14, -611],
      [-268, 222],
      [-480, -389],
      [-733, -112],
      [-451, -753],
      [-2892, 1226],
      [-324, 1247],
      [-395, -249],
      [-663, -55],
      [-253, -665],
      [-748, 249],
      [-70, 443],
      [-367, -470],
      [-663, 747],
      [-141, 881],
      [141, 440],
      [-113, 820]
    ]
  ],
  "transform": {
    "scale": [0.0007790085649189919, 0.0002712813517965561],
    "translate": [9.47021484375, 46.34692761055676]
  },
  "bbox": [
    9.47021484375, 46.34692761055676, 17.259521484375, 49.059469847170526
  ],
  "properties": {
  }
}

def fixup_topo_json_for_folium(topo_json: dict):
    # Folium only supports TopoJSON with GeometryCollection, so hack it together
    for key in topo_json["objects"].keys():
        orig = topo_json["objects"][key]
        new = {"type": "GeometryCollection", "geometries": [orig]}
        topo_json["objects"][key] = new

import folium
map = folium.Map()
# fixup_topo_json_for_folium(topo_json)
folium.TopoJson(topo_json, "objects.A").add_to(map)
map.save("bug.html")
```

**Expected behavior**

It draws the topojson

 - Browser [e.g. chrome, firefox]: Firefox
 - Jupyter Notebook or html files? html
 - Python version (check it with `import sys; print(sys.version_info)`) 3.11
 - folium version (check it with `import folium; print(folium.__version__)`) 0.14.0
 - branca version (check it with `import branca; print(branca.__version__)`) 0.6.0

**Additional context**

It looks like it only supports `GeometryCollection` and not any of the other types, I've included my current workaround in the reproducer called fixup_topo_json_for_folium() which makes it work by wrapping all objects in a dummy GeometryCollection.

KhadeerBasha1232:
Fixes #1816

## Problem
When a TopoJSON's `objects` entry had a `type` of `Polygon` or
`MultiPolygon` (instead of `GeometryCollection`), folium raised:

    KeyError: 'geometries'

This happened because the code assumed every object had a `geometries`
key and didn't handle "bare" geometry objects that store their arcs
directly.

## Fix
Updated the TopoJson handling in `folium/features.py` so it also
supports non-collection geometry types (`Polygon`, `MultiPolygon`,
etc.), not just `GeometryCollection`.

## Testing
Reproduced the exact example from #1816 — it rendered without the
KeyError after this change. Ran the existing test suite locally with
`python -m pytest tests --ignore=tests/selenium` and all tests pass.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -1053,9 +1053,12 @@ class TopoJson(JSCSSMixin, Layer):
             else:
                 return data
 
-        geometries = recursive_get(self.data, self.object_path.split("."))[
-            "geometries"
-        ]  # noqa
+        geometry = recursive_get(self.data, self.object_path.split("."))
+        geometries = (
+            geometry["geometries"]
+            if geometry["type"] == "GeometryCollection"
+            else [geometry]
+        )
         for feature in geometries:
             feature.setdefault("properties", {}).setdefault("style", {}).update(
                 self.style_function(feature)
--- a/folium/features.py
+++ b/folium/features.py
@@ -1047,6 +1047,14 @@ class TopoJson(JSCSSMixin, Layer):
     def style_data(self) -> None:
         """Applies self.style_function to each feature of self.data."""
 
+        for feature in self._get_geometries():
+            feature.setdefault("properties", {}).setdefault("style", {}).update(
+                self.style_function(feature)
+            )  # noqa
+
+    def _get_geometries(self) -> list:
+        """Return the selected TopoJSON object as a list of geometries."""
+
         def recursive_get(data, keys):
             if len(keys):
                 return recursive_get(data.get(keys[0]), keys[1:])
@@ -1054,15 +1062,7 @@ class TopoJson(JSCSSMixin, Layer):
                 return data
 
         geometry = recursive_get(self.data, self.object_path.split("."))
-        geometries = (
-            geometry["geometries"]
-            if geometry["type"] == "GeometryCollection"
-            else [geometry]
-        )
-        for feature in geometries:
-            feature.setdefault("properties", {}).setdefault("style", {}).update(
-                self.style_function(feature)
-            )  # noqa
+        return geometry["geometries"] if "geometries" in geometry else [geometry]
 
     def render(self, **kwargs):
         """Renders the HTML representation of the element."""
@@ -1203,12 +1203,7 @@ class GeoJsonDetail(MacroElement):
             )
             self.warn_for_geometry_collections()
         elif isinstance(self._parent, TopoJson):
-            obj_name = self._parent.object_path.split(".")[-1]
-            keys = tuple(
-                self._parent.data["objects"][obj_name]["geometries"][0][
-                    "properties"
-                ].keys()
-            )
+            keys = tuple(self._parent._get_geometries()[0]["properties"].keys())
         else:
             raise TypeError(
                 f"You cannot add a {self._name} to anything other than a "
--- a/tests/test_features.py
+++ b/tests/test_features.py
@@ -17,6 +17,27 @@ from folium.elements import EventHandler
 from folium.utilities import JsCode
 
 
+@pytest.mark.parametrize("geometry_type", ["Polygon", "MultiPolygon"])
+def test_topojson_non_collection_geometry(geometry_type):
+    """TopoJson supports objects that are not GeometryCollections."""
+    topology = {
+        "type": "Topology",
+        "objects": {
+            "shape": {
+                "type": geometry_type,
+                "arcs": [[0]] if geometry_type == "Polygon" else [[[0]]],
+            }
+        },
+        "arcs": [[[0, 0], [1, 0], [0, 1], [-1, -1]]],
+        "transform": {"scale": [1, 1], "translate": [0, 0]},
+    }
+
+    map_ = folium.Map()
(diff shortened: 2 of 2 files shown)


KhadeerBasha1232:
This fixes #1816 — TopoJson raised a KeyError('geometries') when the
topology object's type was Polygon or MultiPolygon instead of
GeometryCollection, since the code assumed every object had a
`geometries` key.

Fix checks `type` first and falls back to treating the object itself
as a single-item geometry list when it's not a GeometryCollection.

Added a parametrized test covering both Polygon and MultiPolygon cases.
All existing tests + pre-commit checks pass locally. Happy to make any
changes if needed!

adarshsm (commented):
Thanks for picking this up, @KhadeerBasha1232 — the normalize-to-a-list approach is the right shape. A few things I hit while testing it against #1816:

1. **Type-key robustness:** `geometry["type"] == "GeometryCollection"` will itself raise `KeyError: 'type'` if an object omits `type`. Keying on the member you actually need — `"geometries" in geometry` — sidesteps that and reads as "does this hold sub-geometries, or is it one itself?"
2. **Test could assert styling, not just rendering:** since `style_data`'s job is to attach `properties.style`, passing a `style_function` and asserting the style lands would catch a future regression that silently drops styling.
3. **There's a second crash site this doesn't reach.** `style_data` runs before children render, so once it's fixed, a single-geometry TopoJSON with a fields-based `GeoJsonPopup`/`GeoJsonTooltip` hits the same assumption again at `features.py` ~L1205 (`["geometries"][0]["properties"]`). Minimal repro:
   ```python
   import folium
   from folium.features import GeoJsonPopup
   topo = {"type": "Topology",
           "objects": {"A": {"type": "Polygon", "arcs": [[0]], "properties": {"name": "A"}}},
           "arcs": [[[0, 0], [100, 0], [100, 100], [0, 100], [0, 0]]],
           "transform": {"scale": [1e-4, 1e-4], "translate": [9.47, 46.34]}}
   m = folium.Map()
   t = folium.TopoJson(topo, "objects.A")
   t.add_child(GeoJsonPopup(fields=["name"]))
   t.add_to(m)
   m.render()  # KeyError: 'geometries'
   ```
   I have a fix + regression tests for both sites ready — happy to push it as a follow-up PR once this lands, or you're welcome to fold the second-site fix in here. Whatever you and the maintainers prefer.

KhadeerBasha1232:
Thanks for the thorough review, @adarshsm  — really appreciate you digging into the edge cases and providing the repro script.

Pushed a follow-up commit (7157e9b) that addresses all three points:

Type-key robustness — switched the check to "geometries" in geometry instead of geometry["type"] == "GeometryCollection", so it no longer breaks on objects missing a type key. Added a regression test (test_topojson_non_collection_geometry_without_type) covering that exact case.
Second crash site — extracted a shared _get_geometries() helper and reused it in GeoJsonDetail, fixing the KeyError in GeoJsonPopup/GeoJsonTooltip for non-collection TopoJSON. Verified with your repro script, and added a parametrized test (test_topojson_non_collection_geometry_detail) covering both GeoJsonPopup and GeoJsonTooltip.
Styling assertion — updated the existing test to pass a real style_function and assert the style actually lands in properties.style, rather than just checking the map renders.

All local tests pass (pytest tests/test_features.py) along with pre-commit. Let me know if you'd like anything else adjusted — happy to iterate further.

adarshsm:
That's a clean set of changes, @KhadeerBasha1232 — the shared `_get_geometries()` helper reads well, and the parametrized Popup/Tooltip tests cover the second site nicely. Looks complete to me, and pre-commit.ci is green. Thanks for the quick turnaround!

Issue #1806, opened 2023-09-19
Add position control for the Export button of Draw plugin


zxdawn:
The [export button](https://github.com/python-visualization/folium/blob/57e8eaeb33e9a5a70f69130192bece66a6b6d070/folium/plugins/draw.py#L132-L151) is hard-coded and can be overlayed with other layers (e.g. LayerControl on the topright). It's better to add the position parameter.

PR #2251, opened 2026-07-25
Give each Draw export button a unique element id

Htet-Kaung-San:
The `Draw` export button is hard-coded as `id='export'`, and its click handler looks it up with `document.getElementById('export')`. Adding two `Draw` controls with `export=True` emits two elements sharing one id — invalid HTML, and the feature breaks:

```python
import folium

m = folium.Map()
Draw(export=True, filename="a.geojson").add_to(m)
Draw(export=True, filename="b.geojson").add_to(m)
m.save("out.html")
```

On `main` that renders two `<a id='export'>` elements. Every `getElementById('export')` call resolves to the first one, so the second `onclick` assignment overwrites the first. The result is that the **first** button exports the **second** control's `FeatureGroup` under the second control's filename, and the second button does nothing at all.

This derives the id from `get_name()` instead, which is what the other plugins injecting their own elements already do — `ScrollZoomToggler` and `FloatImage` both use `id="{{this.get_name()}}"`.

I also dropped the duplicated `top: 5px` from the button's CSS. `top: 90px` is declared later in the same block and already won, so it was dead code — the rendered position is unchanged.

### On the id change

The id is not documented as public API, but anyone currently styling the button with a `#export { ... }` rule would need to switch to targeting the element another way. Happy to keep a stable class alongside the unique id if you'd rather not break that.

I found this while reading the plugin for #1806; this PR doesn't address that request.

### Testing

`tests/plugins/test_draw.py` is new — the Draw plugin had no test coverage. It covers the unique id, the `export=False` case, and two controls each staying wired to their own `FeatureGroup` and filename. The two id-related tests fail on `main` and pass here.

Full suite, excluding `tests/selenium` and `tests/snapshots` (missing `pixelmatch` locally): 258 passed, 6 failed. The same 6 fail identically on an unmodified checkout — `test_timedynamic_geo_json`, `test_dead_map_id_rule_absent`, `test_icon_invalid_marker_colors`, and three `test_repr.py` png tests — so none are from this change.

`ruff` and `codespell` are clean. `black` wants to rewrap the `Template(...)` call in `draw.py`, but it asks for exactly that same rewrap on an unmodified checkout; I get black 25.11.0 locally since 26.5.1 doesn't install on Python 3.9, so I left the formatting alone rather than add unrelated churn.

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -58,9 +58,8 @@ class Draw(JSCSSMixin, MacroElement):
         {% macro html(this, kwargs) %}
             {% if this.export %}
             <style>
-                #export {
+                #export_{{ this.get_name() }} {
                     position: absolute;
-                    top: 5px;
                     right: 10px;
                     z-index: 999;
                     background: white;
@@ -74,7 +73,7 @@ class Draw(JSCSSMixin, MacroElement):
                     top: 90px;
                 }
             </style>
-            <a href='#' id='export'>Export</a>
+            <a href='#' id='export_{{ this.get_name() }}'>Export</a>
             {% endif %}
         {% endmacro %}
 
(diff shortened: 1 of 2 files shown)


adarshsm (commented):
Nice catch and a clean fix. I went through it against `main` and it holds up:

- **Root cause confirmed:** two `Draw(export=True)` controls emit two `<a id='export'>` elements, and `getElementById('export')` only ever resolves the first — so the second button is dead and the first exports the wrong `FeatureGroup`. Deriving the id from `get_name()` matches how `ScrollZoomToggler`/`FloatImage` already namespace their elements.
- **The `top: 5px` removal is safe:** it and `top: 90px` sit in the same `#export {}` rule, so `top: 90px` already won — rendered position is unchanged.
- **Tests:** ran `tests/plugins/test_draw.py` locally, all 3 pass including the two-controls regression, and confirmed the emitted ids are unique.

One small thing worth a mention (not a blocker): this renames the element id from `export` to `export_<name>`, so anyone with custom CSS targeting `#export` would need to update it — fine given it was duplicate/invalid HTML anyway, just flagging in case it's worth a release note.

LGTM.

PR #2257, opened 2026-07-31
Vendor a glyphicons-only stylesheet instead of the full Bootstrap 3 base

### LINKED ISSUE #1820: Default CSS for glyphicons produces conflict with bootstrap5 components (with potential solution)
**Describe the bug**
Manually adding a bootstrap5 modal to the html body results in wrong position of the modal (see screenshots at the end).

**To Reproduce**

```py
import folium

my_map = folium.Map()

# Adding modal from https://www.w3schools.com/bootstrap5/bootstrap_modal.php
my_map.get_root().html.add_child(folium.Element("""
<!-- Button to Open the Modal -->
<button type="button" class="btn btn-primary" data-bs-toggle="modal" data-bs-target="#myModal">
  Open modal
</button>

<!-- The Modal -->
<div class="modal" id="myModal">
  <div class="modal-dialog">
    <div class="modal-content">

      <!-- Modal Header -->
      <div class="modal-header">
        <h4 class="modal-title">Modal Heading</h4>
        <button type="button" class="btn-close" data-bs-dismiss="modal"></button>
      </div>

      <!-- Modal body -->
      <div class="modal-body">
        Modal body..
      </div>

      <!-- Modal footer -->
      <div class="modal-footer">
        <button type="button" class="btn btn-danger" data-bs-dismiss="modal">Close</button>
      </div>

    </div>
  </div>
</div>
"""))

my_map.save(outfile="map.html")
```

**Expected behavior**
The modal should appear in the same way as in the bootstrap5 modal example on [w3schools](https://www.w3schools.com/bootstrap5/bootstrap_modal.php).

**Environment:**
 - OS: macOS 14.0
 - Browser: Firefox 118.0.1
 - Python version: 3.10.7
 - folium version: 0.14.0
 - branca version: 0.6.0
 - Plain .py, no Jupyter Notebook

**Possible solutions**
This issue is resolved when uncommenting the stylesheet on netdna for bootstrap 3.0.0, however, I cannot tell what the implications to other functions of folium that would have (e.g. glyphicons).

```html
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/leaflet@1.9.3/dist/leaflet.css"/>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.2.2/dist/css/bootstrap.min.css"/>
<!--    <link rel="stylesheet" href="https://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css"/>-->
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@6.2.0/css/all.min.css"/>
    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/Leaflet.awesome-markers/2.0.2/leaflet.awesome-markers.css"/>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/python-visualization/folium/folium/templates/leaflet.awesome.rotate.min.css"/>
```

**Screenshots**
Wrong: with default CSS `<link rel="stylesheet" href="https://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css"/>`
folium_bs5_modal_wrong

Correct: without default CSS `<link rel="stylesheet" href="https://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css"/>`
folium_bs5_modal_correct

adarshsm:
Closes #1820. Thanks @hansthen for the steer.

folium loads netdna's `bootstrap-glyphicons.css` for the Awesome Markers glyphicon font. Despite the name, that file is a **full Bootstrap 3 base stylesheet** — alongside the `@font-face`/`.glyphicon` rules it ships a normalize reset and global `body` rules (font, colour, background, `margin`). Loaded *after* Bootstrap 5, those `body` rules leak onto the host page and, for example, mis-position a user-added Bootstrap 5 modal (the original report).

This vendors `folium/templates/glyphicons.css` containing **only** the `@font-face` and `.glyphicon`/`.glyphicon-*` rules (fonts served from jsDelivr), referenced the same way as the existing `leaflet.awesome.rotate.css`. `Icon(prefix="glyphicon")` markers keep working exactly as before; the page-wide `body`/reset pollution is gone.

Not a customer-facing change — glyphicon markers are unaffected; the only difference is that folium stops overriding the host page's `body` styles. Added a regression test asserting the vendored file is used and the netdna URL is gone.


adarshsm:
Pushed a follow-up that should fix the failing snapshot job.

**What was wrong:** the vendored stylesheet was linked from `cdn.jsdelivr.net/gh/python-visualization/folium/folium/templates/glyphicons.css`, which jsDelivr serves from the default branch — so it 404s until this PR is merged (`curl` → `HTTP 404`). In CI the `.glyphicon` rules never load, and for a released package the URL would also point at a mutable `main`. It's the same `/gh` pattern `awesome_rotate_css` uses, but that file already exists on `main`; a brand-new vendored file can't resolve that way pre-merge.

**Fix (`59a744c`):** inline the glyphicons-only rules into every `Map` as a `<style>` block, read once from the packaged `folium/templates/glyphicons.css`. No external stylesheet request, no chicken-and-egg, and it keeps the GH-1820 goal (no Bootstrap-3 body reset leaking onto the host page). The glyph font still loads from the stable `bootstrap@3.3.7` npm files (HTTP 200). Locally: 23 `test_folium.py` tests pass, black/ruff clean, and glyphicon markers render (icon classes map to the inlined `.glyphicon-*` rules).

**One thing that still needs doing:** the 7 snapshot baselines legitimately change, because dropping the leaked Bootstrap body-reset alters rendering even for non-glyphicon maps (e.g. `issue_1885`). Those need regenerating from the CI Ubuntu/Chrome environment. The workflows on my latest push are sitting in `action_required` — if a maintainer approves the run, I'll grab the `screenshots` artifact, sanity-check the renders, and commit them as the updated baselines.

adarshsm:
CI is fixed in 9d24e8d, and the cause is worth a look before merging.

**The snapshots were failing for a real reason, not a rendering fluke.** All seven failed, and the pixel diffs are confined to text: Leaflet's attribution in the corner and, on `issue_1989`, branca's colorbar caption and tick labels. The map tiles, polygons, markers and controls are pixel-identical. The glyphicons themselves render fine.

The old netdna stylesheet was not glyphicons-only — it carried a full Bootstrap 3 reset, and among it:

```css
html{font-size:62.5%;}
body{font-family:"Helvetica Neue",Helvetica,Arial,sans-serif;font-size:14px;...}
*,*:before,*:after{box-sizing:border-box;}
```

`html{font-size:62.5%}` makes `1rem` = 10px instead of 16px. Leaflet 1.9 sizes the attribution control in `rem`, and branca's colorbar text inherits, so both were rendering about a third smaller than either library intends. Dropping the reset restores them to their own sizes — that is the entire visual delta.

So this is the leak from GH-1820 showing up on the map's own page too, not only on the host page embedding it. The old screenshots recorded the shrunken text; the new ones record what Leaflet and branca ask for.

I took the replacement screenshots from the CI artifact of the run against 59a744c0 (`ubuntu-latest`), so they come from the same environment that compares them rather than from my machine.

One thing to flag explicitly since it is your call, not mine: this makes the attribution and colorbar text slightly larger in **every** map, not just the one in the issue. I think that is correct — folium should not be resizing its own root font — but it is a visible change for existing users, so say the word if you would rather scope the stylesheet differently.

adarshsm:
Rebased onto current `main` (picks up the GeoJson child-layer fixes in #2263 and a few other merged PRs) — the branch had drifted since @hansthen's merge on Aug 24. This should also refresh the stale Selenium run from before the rebase. No changes to the CSS fix itself.

PR #2271, opened 2026-08-22
Add export_position option to the Draw plugin

### LINKED ISSUE #1806: Add position control for the Export button of Draw plugin

Htet-Kaung-San:
Closes #1806.

The Draw plugin's export button is pinned to the top-right corner with hard-coded CSS (`right: 10px; top: 90px;`), so it can end up overlapping another control in that corner — the issue mentions a `LayerControl` on the top right.

This adds an `export_position` parameter taking the four Leaflet corner names (`topright`, `topleft`, `bottomright`, `bottomleft`):

```python
Draw(export=True, export_position="topleft").add_to(m)
```

- Default is `"topright"`, which reproduces the previous placement **exactly** — `position: absolute; top: 90px; right: 10px;` — so existing maps render unchanged.
- An unknown value raises `ValueError` listing the valid corners, rather than silently producing a button with no offsets.

The corner → CSS mapping lives in a small module-level dict, and the template interpolates the resulting edge offsets in place of the old hard-coded lines.

### Testing

Extended `tests/plugins/test_draw.py`:
- `test_draw_export_position_default` — the default still emits `top: 90px` + `right: 10px`.
- `test_draw_export_position_corners` — each corner emits its expected edge pair.
- `test_draw_export_position_invalid` — a bad value raises `ValueError`.

`tests/plugins/test_draw.py` is 6 passed. `ruff check` and `codespell` are clean on both files (ruff v0.16, matching the pinned hook). I left the pre-existing `draw.py` template formatting untouched.

### DIFF
--- a/folium/plugins/draw.py
+++ b/folium/plugins/draw.py
@@ -3,6 +3,25 @@ from branca.element import MacroElement
 from folium.elements import JSCSSMixin
 from folium.template import Template
 
+# Each corner maps to the pair of CSS edge offsets that pins the export button
+# there. "topright" reproduces the historical hard-coded placement exactly.
+_EXPORT_POSITION_CSS = {
+    "topright": "top: 90px;\n                    right: 10px;",
+    "topleft": "top: 90px;\n                    left: 10px;",
+    "bottomright": "bottom: 20px;\n                    right: 10px;",
+    "bottomleft": "bottom: 20px;\n                    left: 10px;",
+}
+
+
+def _export_position_to_css(export_position):
+    try:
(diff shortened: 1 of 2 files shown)


hansthen:
LGTM

PR #2272, opened 2026-08-23
Forward unrecognised Leaflet options in path_options

Htet-Kaung-San:
`path_options()` (shared by `Circle`, `CircleMarker`, `PolyLine`, `Polygon`, `Rectangle`) pops the options it names explicitly and builds the result from those, but never forwards the remaining `kwargs`. So any Leaflet [Path](https://leafletjs.com/reference.html#path) option it doesn't list — `interactive`, `pane`, `renderer`, `attribution`, … — is silently dropped:

```python
from folium.vector_layers import path_options
"interactive" in path_options(radius=10, interactive=False)   # -> False on main
```

A common case is `interactive=False` to make a shape non-clickable — on `main` it just vanishes.

This is inconsistent with the rest of folium: `FeatureGroup`, `LayerGroup`, `TileLayer`, `VideoOverlay` and `GeoJson` all forward `**kwargs` to Leaflet through `remove_empty(**kwargs)`. And `path_options` already special-cases `tags`, `className` and `gradient`, which shows the intent was to support Leaflet options — the general pass-through was just never added.

The fix forwards the leftover options after the named ones:

```python
    default.update(extra_options)
    default.update(kwargs)   # <- pass remaining Leaflet options through
    return default
```

The remaining keys are already camelised at the top of the function, so `interactive`/`pane`/`renderer` and snake_case extras (`my_option` → `myOption`) all arrive correctly, and named options are unaffected because they were already popped.

### Testing

Two new tests in `tests/test_vector_layers.py`: one asserting `path_options` forwards `interactive`/`pane`/a custom option, one asserting `CircleMarker(..., interactive=False)` renders `"interactive": false`. Both fail on `main` and pass here.

`tests/test_vector_layers.py` is 9 passed; the existing 7 are unchanged. `ruff check` and `codespell` are clean. Across `test_features.py`, `test_map.py` and `tests/plugins/`, the only failures are `test_icon_invalid_marker_colors` and `test_timedynamic_geo_json`, which fail identically on an unmodified checkout.

### DIFF
--- a/folium/vector_layers.py
+++ b/folium/vector_layers.py
@@ -123,6 +123,11 @@ def path_options(
         "bubblingMouseEvents": kwargs.pop("bubblingMouseEvents", True),
     }
     default.update(extra_options)
+    # Pass any remaining options straight through to Leaflet, matching how
+    # the other folium option builders forward **kwargs. Without this, Path
+    # options such as ``interactive``, ``pane`` or ``renderer`` were silently
+    # dropped from vector overlays.
+    default.update(kwargs)
     return default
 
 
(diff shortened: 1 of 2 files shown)


hansthen:
Good catch.

PR #2274, opened 2026-08-27
Upgrade Leaflet from 1.9.3 to 1.9.4

nina-mir:
As requested in https://github.com/python-visualization/folium/pull/2263#issuecomment-5409714983.

Bumps the pinned Leaflet 1.9.3 → 1.9.4, and restores the `verify_js_logs()` call in `tests/selenium/test_multipoints_tooltip_selenium.py` that #2263 had to omit. 

1.9.4 guards the unchecked `layer.getElement()` in `_addFocusListenersOnLayer` (`Tooltip.js:396`): https://github.com/Leaflet/Leaflet/compare/v1.9.3...v1.9.4

The failing test is `test_notebook[coordinate_ordering.md]`, but the error doesn't originate there since that page has no tooltips, and in 1.9.3 only `Tooltip.js` calls `getElement()`, from the focus-listener path that runs only when a tooltip is bound. It's the MultiPoint test from #2263 that throws. Because that test skipped `verify_js_logs()`, the uncaught error stayed in the session-scoped driver's console log and surfaced in whichever later test read it first.

Worth noting for future selenium tests in this repo: skipping `verify_js_logs()` doesn't contain an error locally, it defers it to another test.

Reproducible on `main`: `coordinate_ordering` passes alone and fails when run after the MultiPoint test.

```
pytest tests/selenium/test_multipoints_tooltip_selenium.py \
       tests/selenium/test_selenium.py -k "multipoint or coordinate_ordering"
```

Locally on 1.9.4: `tests/selenium` 76 passed, 1 xfailed — including the heatmap
pixel-diff test, so no rendering changes.

### DIFF
--- a/folium/folium.py
+++ b/folium/folium.py
@@ -25,7 +25,7 @@ from folium.utilities import (
 )
 
 _default_js = [
-    ("leaflet", "https://cdn.jsdelivr.net/npm/leaflet@1.9.3/dist/leaflet.js"),
+    ("leaflet", "https://cdn.jsdelivr.net/npm/leaflet@1.9.4/dist/leaflet.js"),
     ("jquery", "https://code.jquery.com/jquery-3.7.1.min.js"),
     (
         "bootstrap",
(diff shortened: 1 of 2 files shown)


hansthen:
@nina-mir Thanks for picking this up.

nina-mir:
Thank you so much!

I opened #2282 with the docs follow-up since both GeoJSON pages still describe the
MultiPoint limitation as current.

PR #2276, opened 2026-09-01
Fix three stale documented defaults (TopoJson.overlay, TimeSliderChoropleth.overlay, GeoJsonPopup.localize)

dylanpulver:
Three classes override a default and kept the parameter description they were copied from, so the documentation states the old value:

| | documented | actually defaults to |
|---|---|---|
| `TopoJson.overlay` | `False` | `True` |
| `TimeSliderChoropleth.overlay` | `False` | `True` |
| `GeoJsonPopup.localize` | `False` | `True` |

In each case the code is right and the docstring is stale, and the tie is broken by a sibling in the same file rather than by preference:

- `Layer` (`map.py`) both documents and defaults `overlay=False`. `GeoJson` overrides it to `True` **and** documents `True`. `TopoJson` and `TimeSliderChoropleth` override it to `True` and kept `Layer`'s line.
- `GeoJsonTooltip` documents `localize` as `False` and defaults to `False`. `GeoJsonPopup` overrides to `True` and kept the tooltip's line.

`overlay` decides whether a layer shows up in `LayerControl` as a toggleable overlay or as a base layer, so this is a wrong mental model rather than a cosmetic slip.

The new test walks every class in the package — 96 of them — and compares the documented boolean default with the signature default for `overlay`, `control`, `show` and `localize`. It is restricted to those four because a bare `True`/`False` in a docstring is unambiguous to compare, whereas the numeric and string defaults elsewhere are written in too many styles to check mechanically without false positives.

Verification: with the three docstrings reverted to their current text, exactly those three parametrised cases fail and the other 94 pass, so the check is neither vacuous nor over-broad. There is also a small self-check asserting the docstring parser really does detect a mismatch. `black` 26.5.1 and `ruff` 0.16.3 (the versions pinned in `.pre-commit-config.yaml`) report the three files unchanged and clean.

Suite run locally, Python 3.14, excluding `tests/selenium`: 369 passed, 2 failed — `test_icon_invalid_marker_colors` and `test_valid_png_size`, both of which fail identically on unmodified `main` here, so they look like environment rather than regressions. `tests/snapshots` could not be collected at all locally because `pixelmatch` is not installed; that path is untested by me.

### DIFF
--- a/folium/features.py
+++ b/folium/features.py
@@ -950,7 +950,7 @@ class TopoJson(JSCSSMixin, Layer):
         A function mapping a TopoJson geometry to a style dict.
     name : string, default None
         The name of the Layer, as it will appear in LayerControls
-    overlay : bool, default False
+    overlay : bool, default True
         Adds the layer as an optional overlay (True) or the base layer (False).
     control : bool, default True
         Whether the Layer will be included in LayerControls.
@@ -1338,7 +1338,7 @@ class GeoJsonPopup(GeoJsonDetail):
         instead of the keys of `fields`.
     labels: bool, default True.
         Set to False to disable displaying the field names or aliases.
-    localize: bool, default False.
+    localize: bool, default True.
         This will use JavaScript's .toLocaleString() to format 'clean' values
         as strings for the user's location; i.e. 1,000,000.00 comma separators,
         float truncation, etc.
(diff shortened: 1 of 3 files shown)


hansthen:
@dylanpulver Thanks for fixing!

